PR #14, opened 2021-10-15
add path id type in illustration figure 3

yfmascgy:


mirjak:
I wil merge this one as I consider it editorial

PR #16, opened 2021-10-15
SHOULD -> MUST for use of different CIDs on different paths

mirjak:
There is a SHOULD here, which I believe should be a MUST as we taking about the receiver behaviour here.

### DIFF
--- a/draft-lmbdhk-quic-multipath.md
+++ b/draft-lmbdhk-quic-multipath.md
@@ -270,8 +270,9 @@ If endpoints negotiate multipath support with value 1 to use multiple
 packet number spaces, they SHOULD use ACK_MP frames
 instead of ACK frames to signal acknowledgement of 1-RTT packets on path number 0, and also 0-RTT packets as specified in {{handling-of-0-rtt-packets}}.
 
-1-RTT packets sent to different paths SHOULD carry different connection identifiers, if a non-zero
-connection ID is used.
+1-RTT packets sent to different paths MUST carry different connection identifiers, if a non-zero
+connection ID is used. If a packet is received on a new 4-tuple but with an already used connection ID,
+the sender has migrated to that address and the receiver reacts as specfied in Section 9.3 in {{QUIC-TRANSPORT}}.
 
 ## Handling of 0-RTT Packets
 


draft-lmbdhk-quic-multipath.md
```diff
@@ -270,8 +270,9 @@ If endpoints negotiate multipath support with value 1 to use multiple
 packet number spaces, they SHOULD use ACK_MP frames
 instead of ACK frames to signal acknowledgement of 1-RTT packets on path number 0, and also 0-RTT packets as specified in {{handling-of-0-rtt-packets}}.
 
-1-RTT packets sent to different paths SHOULD carry different connection identifiers, if a non-zero
-connection ID is used.
+1-RTT packets sent to different paths MUST carry different connection identifiers, if a non-zero
+connection ID is used. If a packet is received on a new 4-tuple but with an already used connection ID,
+the sender has migrated to that address and the receiver reacts as specfied in Section 9.3 in {{QUIC-TRANSPORT}}.
```
huitema, line 276: Section 9.5 of RFC 9000 says: _"At any time, endpoints MAY change the Destination Connection ID they transmit with to a value that has not been used on another path."_ I don't know whether we want to repeat that, but it seems that we have to consider two cases:

1) Packet received with a different 4-tuple, same CID: treat as specified in section 9.3 of RFC 9000, i.e., handle NAT rebinding.
2) Packet received on existing 4-tuple, different CID: treat as specified in section 9.5 of RFC 9000, i.e., renewal of CID on an existing path.

mirjak: Yes, the rebinding case is considered here. The text says that the sender must change the CID when you move to a new path, however, the receiver should not expect to always see a new CID if the address changes (because of NAT rebinding). This MUST is only for the receiver.

qdeconinck: Maybe a way to rephrase this:
"An endpoint sending 1-RTT packets MUST use different Connection IDs on different paths, if non-zero-length Connection IDs are used. Still, the receiver may observe the same Connection ID used on different 4-tuples due to, e.g., NAT rebinding. In such case, the receiver reacts as specified in {{Section 9.3 of QUIC-TRANSPORT}}."

PR #19, opened 2021-10-15
Simplify PATH_STATUS frame with PATH_ABANDON frame

yfmascgy:
Simplify path_status frame with path_abandon frame following the discussion on issue#2. Error code and Reason Phrase are left TBD.

### DIFF
--- a/draft-lmbdhk-quic-multipath.md
+++ b/draft-lmbdhk-quic-multipath.md
@@ -603,7 +603,7 @@ SHOULD use type 0x01 for path identifier. If both endpoints are using 0-length C
 endpoints SHOULD only use type 0x02 for path identifier.
 
 Error Code:
-: A variable-length integer that indicates the reason for closing this connection.
+: A variable-length integer that indicates the reason for abandoning this path.
 
 Reason Phrase Length:
 : A variable-length integer specifying the length of the reason phrase in bytes.


draft-lmbdhk-quic-multipath.md
```diff
@@ -606,32 +596,17 @@ Path Identifier: An identifier of the path, which is formatted as shown in {{fig
...
-
-If the value of Path Status field is 2 ("available"), the receiver side can use
-the Path Priority field to express the priority weight of a path for the peer.
-
-Frames may be received out of order. A peer MUST ignore an incoming PATH_STATUS
-frame if it previously received another PATH_STATUS frame for the same Path
-Identifier with a sequence number equal to or higher than the sequence number
-of the incoming frame.
+Error code: TBD
+Reason Phrase: TBD
```
huitema, line 607: How about just adapting the text from QUIC Connection_Close frame:
```
Error Code:
: A variable-length integer that indicates the reason for closing this connection.

Reason Phrase Length:
: A variable-length integer specifying the length of the reason phrase in bytes.
  Because an ABANDON_FRAME frame cannot be split between packets, any limits
  on packet size will also limit the space available for a reason phrase.

Reason Phrase:
: Additional diagnostic information for the closure. This can be zero length if
  the sender chooses not to give details beyond the Error Code value.
  This SHOULD be a UTF-8 encoded string {{!RFC3629}}, though the frame
  does not carry information, such as language tags, that would aid comprehension
  by any entity other than the one that created the text.
```

The error code will often be "No Error". However, some issues can happen, such as closing a path because resource are exceeded, or because there are too many transmission errors, or anything else. Having error codes and an optional reason phrase will help diagnostic and debugging.

draft-lmbdhk-quic-multipath.md
```diff
@@ -689,7 +664,7 @@ under the "QUIC Protocol" heading.
 Value                                              | Frame Name          | Specification
 ---------------------------------------------------|---------------------|-----------------
 TBD-00 - TBD-01 (experiments use 0xbaba00-0xbaba01)| ACK_MP              | {{mp-ack-frame}}
-TBD-02 (experiments use 0xbaba03)                  | PATH_STATUS         | {{path-status-frame}}
+TBD-02 (experiments use 0xbaba03)                  | PATH_ABANDON         | {{path-abandon-frame}}
 {: #frame-types title="Addition to QUIC Frame Types Entries"}
```
huitema, line 669: How about not reusing the exact same code as the old PATH_STATUS frame? I would suggest something like "experiments use 0xbaba05".

yfmascgy: Yes, the PATH_ABANDON type is changed to 0xbaba05 now.

draft-lmbdhk-quic-multipath.md
```diff
@@ -606,32 +596,29 @@ Path Identifier: An identifier of the path, which is formatted as shown in {{fig
...
-Path Status sequence number: A variable-length integer specifying the sequence number assigned for this PATH_STATUS frame.
-There is a different path status sequence number space for each path.
-
-Available values of Path Status field are:
-
-- 0: Abandon
-- 1: Standby
-- 2: Available
+Error Code:
+: A variable-length integer that indicates the reason for closing this connection.
```
huitema, line 607: Sorry, should be "closing this path".

huitema (approved):
I made a typo, see comment on closing this path. Apart from that, looks good.

Yanmei-Liu (approved):
Look good to me.

PR #23, opened 2021-10-18
ACK_MP only valid using multiple packet number spaces

### LINKED ISSUE #8: ACK_MP and single packet number space
When negotiating the use of a single packet number space, it is not clear whether the use of ACK_MP frame is acceptable or not. Should we only stick to ACK frames and consider reception of ACK_MP frames to be MP_PROTOCOL_VIOLATION error? Or should we tolerate their reception if the packet number space identifier is 0?

qdeconinck:
Fixes #8.


mirjak (approved):

Yanmei-Liu (approved):
Look good to me.

huitema (approved):

Issue #10, opened 2021-10-15
States of paths


qdeconinck:
Given that paths are initiating/validating/active/closed, it looks like we will need at some point to specify some path state (finite state machine?) and what behavior is acceptable in each of them (e.g., we cannot send packets on a closed state,...).

PR #24, opened 2021-10-18
Introduce finite state machine for path's state

qdeconinck:
Initial attempt to solve #10.

### DIFF
--- a/draft-lmbdhk-quic-multipath.md
+++ b/draft-lmbdhk-quic-multipath.md
@@ -243,14 +243,13 @@ after a spurious estimate of path abandonment by the client.
 
 ~~~
        o
-       | Initiate path validation
-       | PATH_CHALLENGE received on new path
+       | PATH_CHALLENGE sent/received on new path
        v
  +------------+    Path validation abandoned
  | Validating |----------------------------------+
  +------------+                                  |
        |                                         |
-       | Path validation succeeded               |
+       | PATH_RESPONSE received                  |
        |                                         |
        v        Associated CID have been retired |
  +------------+        Path's idle timeout       |
@@ -272,7 +271,7 @@ after a spurious estimate of path abandonment by the client.
 ~~~
 {: #fig-path-states title="States of a path"}
 
-In non-terminal states, hosts MUST track the following information.
+In non-final states, hosts have to track the following information.
 
 - Associated 4-tuple: The tuple (source IP, source port, destination IP, destination port)
 used by the endhost to send packets over the path.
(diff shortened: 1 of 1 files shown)


draft-lmbdhk-quic-multipath.md
```diff
@@ -256,6 +256,71 @@ Senders MUST manage per-path congestion status, and MUST NOT send more data on a
...
 Multipath TCP uses the LIA congestion control scheme specified in {{RFC6356}}.  This scheme can immediately be adapted to Multipath QUIC. Other coupled congestion control schemes have been proposed for Multipath TCP such as {{OLIA}}.
 
+# Path States
+
+{{fig-path-states}} shows the states that an endpoint's path can have.
+
+~~~
+       o
+       | Initiate path validation
+       | PATH_CHALLENGE received on new path
```
mirjak, line 266: Isn't the first step to send (not received) a PATH_CHALLENGE and then you stay in "Validating" state until you receive PATH_RESPONSE (which would trigger the transition to the next state)?

draft-lmbdhk-quic-multipath.md
```diff
@@ -256,6 +256,71 @@ Senders MUST manage per-path congestion status, and MUST NOT send more data on a
...
+
+- Path Packet Number Space: The endpoint maintains a separate packet number for sending and receiving packets over this path.
+Packet number considerations described in {{QUIC-TRANSPORT}} apply within the given path.
+
+In the "Active" state, hosts MUST also track the following information.
+
+- Associated Source Connection ID: The Connection ID used to receive packets over the path.
+
+A path in the "Validating" state performs path validation as described in {{Section 8.2 of QUIC-TRANSPORT}}.
+An endhost SHOULD NOT send non-probing frames on a path in "Validating" state, as it has no guarantee that packets
```
mirjak, line 311: We should not use normative language in this section, however, not sure if normative language is needed in this case.

draft-lmbdhk-quic-multipath.md
```diff
@@ -256,6 +256,71 @@ Senders MUST manage per-path congestion status, and MUST NOT send more data on a
...
+will actually reach the peer.
+
+The endhost can use all the paths in the "Active" state, provided that their usage is enabled by the path's congestion
```
mirjak, line 314: "usage is enabled by" is a bit unclear. Maybe: ... provided the congestion control and flow control currently allow sending of new data on a path

mirjak (changes requested):

draft-lmbdhk-quic-multipath.md
mirjak: I think I would omit the part in this diagram where  PATH_CHALLENGE is received by not not send and just define that the validating is enter as soon as a PATH_CHALENGE is sent. Or if you want to show this in the diagram then we would need a separate state for this? Or maybe a separate diagram for the initiator and the "responder"?

Issue #1, opened 2021-10-15
More illustration figures on different types of path identifiers?


yfmascgy:
There is one part I am afraid people might feel a little bit hard to follow: the three types of path identifiers in the PATH_STATUS frame. Those three types of path identifiers were introduced in order to be compatible with zero-length CID use cases, so I think we might need two additional example figures (like figure 3 in the current draft), one for (non-zero length CID, zero-length CID) and the other one for (zero-length CID & zero-length CID). Do you think that is a good idea? I can work on those figures if it doesn't complicate the flow of the draft.

yfmascgy:
@huitema I am editing the path closing figures with zero-length CID and single PN. Is it similar to what you have in mind?

~~~
  Client                                                      Server

  (client tells server to abandon a path)
  1-RTT[X]: DCID=S2 PATH_ABANDON[path_id_type=1, path_id=2]->
                             (server tells client to abandon a path)
                <-1-RTT[Y]: PATH_ABANDON[path_id_type=0, path_id=2],
                                                 ACK_MP[Seq=0, PN=X]
  (client abandons the path that it is using)
  1-RTT[X+1]: DCID=S3 RETIRE_CONNECTION_ID[2],
                            ACK_MP[Seq=0, PN=Y]->
                                 <-1-RTT[Y+1]: ACK_MP[Seq=0, PN=X+1]
  (client waits for idle timeout to release resources)

~~~
{: #fig-example-path-close2 title="Example of closing a path when the server chooses to use zero-length CID"}

~~~
  Client                                                     Server

  (client tells server to abandon a path)
  1-RTT[X]: DCID=S2 PATH_ABANDON[path_id_type=2]->
                            (server tells client to abandon a path)
                          <-1-RTT[Y]: PATH_ABANDON[path_id_type=2],
                                                ACK_MP[Seq=0, PN=X]
  1-RTT[X+1]: ACK_MP[Seq=0, PN=Y]->
  (client waits for idle timeout to release resources)
               (server waits for idle timeout to release resources)

~~~
{: #fig-example-path-close3 title="Example of closing a path when the both client and server choose to use zero-length CIDs."}

huitema:
In the case of single number space, we use ACK, not ACK_MP.  But in fact you do not need to draw the ACKs there, there is no special processing of ACKs required. Also, there is no need to wait for an idle timeout, which typically is 10 to 30 seconds. That's too large. 

[...]

PR #28, opened 2021-10-21
add more illustration figures to path close, maybe need revisions

yfmascgy:
I added 2 illustration figures describing path close when null-length CIDs are used. This changes are based on the discussion of issue #1 . But I think it may need additional revision.

### DIFF
--- a/draft-lmbdhk-quic-multipath.md
+++ b/draft-lmbdhk-quic-multipath.md
@@ -143,7 +143,7 @@ during the connection handshake, as specified in {{QUIC-TRANSPORT}}. The new tra
 defined as follow:
 
 - name: enable_multipath (TBD - experiments use 0xbaba)
-- value: 0 (default) for disabled. Endpoints use 2-bits in the value field for negotiating one or more 
+- value: 0 (default) for disabled. Endpoints use 2-bits in the value field for negotiating one or more
 PN spaces, available option value for client and server are listed in {{param_value_definition}} :
 
 Client Option| Definition                                      | Allowed server responses
@@ -154,10 +154,10 @@ Client Option| Definition                                      | Allowed server 
 0x3	      | support both one PN space and multiple PN space | 0x0, 0x1 or 0x2
 {: #param_value_definition title="Available value for enable_multipath"}
 
-If the peer does not carry the enable_multipath transport parameter, which means the peer does not 
-support multipath, endpoint MUST fallback to {{QUIC-TRANSPORT}} with single path and MUST NOT use 
-any frame or mechanism defined in this document. If endpoint receives unexpected value for the transport parameter 
-"enable_multipath", it MUST treat this as a connection error of type MP_CONNECTION_ERROR 
+If the peer does not carry the enable_multipath transport parameter, which means the peer does not
+support multipath, endpoint MUST fallback to {{QUIC-TRANSPORT}} with single path and MUST NOT use
+any frame or mechanism defined in this document. If endpoint receives unexpected value for the transport parameter
+"enable_multipath", it MUST treat this as a connection error of type MP_CONNECTION_ERROR
 and close the connection.
 
 Note that the transport parameter "active_connection_id_limit" {{QUIC-TRANSPORT}} limits the number of usable
@@ -495,7 +495,8 @@ not cause linkability issue.
 
 ## Path Establishment
 
-{{fig-example-new-path}} illustrates an example of new path establishment using multiple packet number spaces.
+{{fig-example-new-path}} illustrates an example of new path establishment using
+multiple packet number spaces.
 
 ~~~
    Client                                                  Server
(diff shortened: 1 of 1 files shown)


draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,111 @@ not cause linkability issue.
...
+the first path, the client's 1-RTT packets use DCID S2, which has a sequence
+number of 2, so the server can still use path identifier type 0x00 and a
+path_id=2 in its PATH_ABANDON frame. However, because the server now uses
+zero-length CID, the client needs to use path identifier type 0x01 with a
+path_id=2 in its PATH_ABANDON frame, which is different from the above example.
+The server can readily close the first path when it receives the
+RETIRE_CONNECTION_ID frame from the client. However, since the client will not
+receive a RETIRE_CONNECTION_ID frame, after sending out RETIRE_CONNECTION_ID
+frame, the client waits for 3 RTO before closing the path. Note that single
+packet number spaces are used in both directions.
```
mirjak, line 578: Isn't that always the case that if a single packet number space is used in one directly is also needs to be used in the other direction?

Also the assumption here is because there is a non-zero length CID in on direction, this has to use one packet number space? Maybe better so say this at the beginning of this paragraph.

draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,111 @@ not cause linkability issue.
...
-{{fig-example-path-close}} illustrates an example of path closing. In this
-case, we are going to close the first path. For the first path, the server's
-1-RTT packets use DCID C1, which has a sequence number of 1; the client's 1-RTT
-packets use DCID S2, which has a sequence number of 2.  For the second path,
-the server's 1-RTT packets use DCID C2, which has a sequence number of 2; the
-client's 1-RTT packets use CID S3, which has a sequence number of 3. Note that
-two paths use different packet number space.
+{{fig-example-path-close1}} illustrates an example of path closing when both
+the client and the server use non-zero length CIDs. Path identifier type 0x00
+is chosen. In this case, we are going to close the first path. For the first
```
qdeconinck, line 540: I would say "... use non-zero length CIDs. For the first path, the server's 1-RTT packets use DCID C1... Note that two paths use different packet number spaces." Then, I would complement by "In this case, the client is going to close the first path. It identifies it by the sequence number of the DCID it uses over that path (path identifier type 0x00), hence using the path_id 1. In reply, the server confirms the path using the sequence number of the DCID it uses over that path (path identifier type 0x00), corresponding to path_id 2.

draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,111 @@ not cause linkability issue.
...
+                                                  ACK_MP[Seq=3, PN=U]
 
 ~~~
-{: #fig-example-path-close title="Example of closing a path (path id type=0x00)"}
+{: #fig-example-path-close1 title="Example of closing a path when both the
+server and client use non-zero length CIDs"}
+
+{{fig-example-path-close2}} illustrates an example of path closing when the
+client uses non-zero length CID while the server uses zero-length CID. Similar
+to the above example, in this case, we are going to close the first path. For
```
qdeconinck, line 568: I would precise that this is still the client that initiates the path closure.

draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,111 @@ not cause linkability issue.
...
+the first path, the client's 1-RTT packets use DCID S2, which has a sequence
+number of 2, so the server can still use path identifier type 0x00 and a
```
qdeconinck, line 570: Shouldn't the server use 0x01 as path_identifier_type instead? Maybe the figure should be updated as well.

draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,111 @@ not cause linkability issue.
...
+path_id=2 in its PATH_ABANDON frame. However, because the server now uses
+zero-length CID, the client needs to use path identifier type 0x01 with a
```
qdeconinck, line 572: As I understand the path identifier type, it should be 0x00 here

draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,111 @@ not cause linkability issue.
...
+                         (server waits for 3*RTO to close the path)
+
+~~~
+{: #fig-example-path-close3 title="Example of closing a path when both client
+and server use zero-length CIDs."}
+
+In scenarios such as client detects the network environment change (client's
+4G/Wi-Fi is turned off, Wi-Fi signal is fading to a threshold), or endpoints
+detect that the quality of RTT or loss rate is becoming worse, client or server
+can terminate a path immediately.
```
qdeconinck, line 625: I suppose it can terminate using a path immediately and maybe send a PATH_ABANDON later, but I assume the use case here is to say that PATH_ABANDON is a kind of graceful path closure, but abrupt closures are also possible.

draft-lmbdhk-quic-multipath.md
```diff
@@ -517,45 +518,120 @@ not cause linkability issue.
...
+                                                  ACK_MP[Seq=2, PN=X]
   (client abandons the path that it is using)
-  1-RTT[U]: DCID=S3 RETIRE_CONNECTION_ID[2], ACK_MP[Seq=1, PN=Y] ->
-                             (server abandons the path that it is using)
-       <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[1], ACK_MP[Seq=3, PN=U]
+  1-RTT[U]: DCID=S3 RETIRE_CONNECTION_ID[2],
+                          ACK_MP[Seq=1, PN=Y]->
+                          (server abandons the path that it is using)
+                         <-1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[1],
+                                                  ACK_MP[Seq=3, PN=U]
```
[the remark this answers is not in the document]
mirjak: This is what we say in the dark currently:

`The receiver of an PATH_ABANDON frame MAY also send an PATH_ABANDON frame to signal its own willingness to not send any packet on this path anymore.`

I guess we could remove the abandon from from the server in this example or clarify in the text that this is optional.

@yfmascgy can you update the PR accordingly?

PR #73, opened 2021-11-03
[~] adjust the discussion for pros/cons of pn spaces to Section 7

Yanmei-Liu:


draft-lmbdhk-quic-multipath.md
```diff
@@ -541,6 +536,15 @@ ACK frames are received in 1-RTT packets while the state of multipath
 negotiation is ambiguous, they MUST be interpreted as acknowledging
 packets sent on path 0.
 
+Endpoints negotiate the use of one packet number space for all paths or
+separate packet number spaces per path during the connection handshake
+{{nego}}. While separate packet number spaces allow for more efficient
+ACK encoding, especially when paths have highly different latencies,
+this approach requires the use of a connection ID. Therefore use of
+a single number space can be beneficial when endpoints use zero-length
+connection ID for less overhead, or in highly constrained networks
```
mirjak, line 545:
```suggestion
connection ID for less overhead, e.g., in highly constrained networks
```

mirjak (approved):

draft-lmbdhk-quic-multipath.md
```diff
@@ -541,6 +536,15 @@ ACK frames are received in 1-RTT packets while the state of multipath
...
 
+Endpoints negotiate the use of one packet number space for all paths or
+separate packet number spaces per path during the connection handshake
+{{nego}}. While separate packet number spaces allow for more efficient
+ACK encoding, especially when paths have highly different latencies,
+this approach requires the use of a connection ID. Therefore use of
+a single number space can be beneficial when endpoints use zero-length
+connection ID for less overhead, e.g., in highly constrained networks
+that do not benefit from exposing the connection ID in the header.
+
```
huitema, line 547: I would drop the example in the last sentence. It implies that overhead is only an issue in "highly constrained" networks, while in practice Chrome does it to minimize aggregated overhead regardless of network constraints. Dropping the text from ", e.g., " to the final dot avoids speculation on why people do not like overhead.

huitema: Note that there are really 2 arguments for the single space approach. Overhead is one, but "fewer code changes" is another. So I would say "this approach requires the use of a connection ID and also more changes from QUIC version 1."

Yanmei-Liu: I deleted the example after "e.g.". 

"more changes from QUIC version 1" depends on the specific implementation, as multiple PN spaces could keep the loss detect and ack arrangement part unchanged.
So maybe we can learn from more implementations and I prefer not to draw the conclusion too early.

qdeconinck (approved):

PR #76, opened 2021-11-05
Do not assume real time formatting of server TP

### LINKED ISSUE #75: Handshake negotiation: what does "Allowed server responses" mean
In Section 2 you define a transport parameter and set of values. These are presented in a table with a column header "Allowed server responses", what does that mean? One way to read this is that it implies servers would tailor their Transport Parameters to the value sent by the client but that's conflicting with our philosophy that TPs are unilateral declarations, not active negotiation.

So the question for me is, is this just a wording issue or a design one?

huitema:
The current text assumes that the server transport parameters are composed after seeing the client's proposal. As pointed out in issue #75, this is a deviation from standard QUIC practice. The proposed text fixes that and specifies what happens if both endpoints select option 3, "support both". In that case, option 2 "multiple number spaces" is selected.

Close #75

### DIFF
--- a/draft-lmbdhk-quic-multipath.md
+++ b/draft-lmbdhk-quic-multipath.md
@@ -196,20 +196,27 @@ defined as follow:
 - value: 0 (default) for disabled.
 
 Endpoints use 2-bits in the value field for negotiating one or more
-PN spaces, available option value for client and server are listed in
+PN spaces, available option values are listed in
 {{param_value_definition}} :
 
-Client Option| Definition                                  | Allowed server responses
----------|-------------------------------------------------|--------------------------
-0x0      | don't support multipath                         | 0x0
-0x1      | only support one PN space for multipath         | 0x0 or 0x1
-0x2      | only support multiple PN spaces for multipath   | 0x0 or 0x2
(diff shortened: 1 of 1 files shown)


draft-lmbdhk-quic-multipath.md
```diff
@@ -196,21 +196,28 @@ defined as follow:
...
+---------|---------------------------------------
+0x0      | don't support multipath
+0x1      | only support one PN space for multipath
+0x2      | only support multiple PN spaces for multipath
+0x3      | support both one PN space and multiple PN space
 {: #param_value_definition title="Available value for enable_multipath"}
 
-If the parameter is absent or set to 0, the endpoints MUST fallback to
+If for anyone of the endpoints the parameter is absent or set to 0,
+or if the two enpoints select incompatible values,
```
LPardue, line 211:
```suggestion
If for any one of the endpoints the parameter is absent or set to 0,
or if the two endpoints select incompatible values,
```

LPardue (approved):
LTGTM just 2 nits

mirjak (approved):

PR #80, opened 2021-11-16
Fix typo in section "path close"

iyangsj:


yfmascgy (approved):
That indeed should be "path".

qdeconinck (approved):

Issue #63, opened 2021-10-27
Add "Path" to the terminology section


boucadair:
Consider adding this entry to the terminology section:

`- Path: refers to the 4-tuple {source IP address, source port number, destination IP address, destination port number}. A path refers to "network path" used in {{QUIC-TRANSPORT}}.`

This also clarifies the link with "network path" in RFC9000.

mirjak:
PR merged

PR #83, opened 2022-01-11
Define path

mirjak:
fixes issue #63


qdeconinck (approved):

obonaventure (approved):
Note that this definition is different from RFC6182 that required the utilization of at least one different addresses for different subflows. In practice, there were deployments where this requirement was not respected and it was overly conservative.

mirjak:
Good point but I guess this definition is inline with what's already specified for QUIC given NATs...

yfmascgy (approved):

PR #97, opened 2022-03-03
New text on ECN handling with single PNS

### LINKED ISSUE #87: ECN support and single/multiple packet number spaces
With a packet number space per path, the ACK_MP frame provides the per path ECN feedback that enables a sender to detect congestion using ECN on a specific path. The per-path congestion control can adjust the congestion window based on the received feedback.

With a single packet number space shared among the different paths, the ECN fields of the ACK frame provide ECN information that is common to all paths. This makes it impossible for a sender to determine which path is affected by congestion when it receives an ACK frame with positive ECN counters. A per-path congestion control cannot easily react to such an ambiguous ECN feedback.

mirjak:
resolves #87

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -640,7 +640,7 @@ If a host receives an
 ACK frame that increases the CE counter and contains packets that have
 been sent on different paths, it MUST treat the CE marking as if it was
 received on either of the path and therefore consider it as a congestion
-indication for all paths with packets that have been acknowledges in that
+indication for all paths with packets that have been acknowledged in that
 ACK frame. An host that receives an ACK with an ECN counter increase that
 acknowledges packets from different paths MAY disable ECN marking and
 send all subsequent packets as Not-ECN capable.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -641,8 +641,8 @@ ACK frame that increases the CE counter and contains packets that have
 been sent on different paths, it MUST treat the CE marking as if it was
 received on either of the path and therefore consider it as a congestion
 indication for all paths with packets that have been acknowledged in that
-ACK frame. An host that receives an ACK with an ECN counter increase that
-acknowledges packets from different paths MAY disable ECN marking and
+ACK frame. An host that receives an ACK with an ECN counter increase
+acknowledging packets from different paths MAY disable ECN marking and
 send all subsequent packets as Not-ECN capable.
 
 ## Using Multiple Packet Number Spaces
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -645,6 +645,9 @@ ACK frame. An host that receives an ACK with an ECN counter increase
 acknowledging packets from different paths MAY disable ECN marking and
 send all subsequent packets as Not-ECN capable.
 
+Alternatively, a sender MAY decide to disable ECN entirely when a single
+packet number space is used by marking all sent packets as Not-ECN capable.
+
 ## Using Multiple Packet Number Spaces
 
 If the multipath option is enabled with a value of 2, each path has
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -633,7 +633,7 @@ frame acknowledges multiple packets, the ECN feedback cannot be accounted
 to a specific packet. Respectively, if an ACK frames acknowledges multiple
 packets from different paths, the ECN feedback cannot unambiguously be assigned to
 a path. Therefore if one of the ECN counters increases, especially the CE
-counter, hosts SHOULD only acknowlegde packets from the same path where
+counter, hosts SHOULD only acknowledge packets from the same path where
 the ECN mark was observed in the same ACK frame.
 
 If a host receives an
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -625,6 +625,26 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
+### Handling ECN Feedback
+
+ECN feedback in QUIC is provided based on counters in the ACK frame
+(see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK
+frame acknowledges multiple packets, the ECN feedback cannot be accounted
+to a specific packet. Respectively, if an ACK frames acknowledges multiple
+packets from different paths, the ECN feedback cannot unambiguously be assigned to
+a path. Therefore if one of the ECN counters increases, especially the CE
+counter, hosts SHOULD only acknowlegde packets from the same path where
+the ECN mark was observed in the same ACK frame.
+
+If a host receives an
+ACK frame that increases the CE counter and contains packets that have
+been sent on different paths, it MUST treat the CE marking as if it was
+received on either of the path and therefore consider it as a congestion
+indication for all paths with packets that have been acknowledges in that
+ACK frame. An host that receives an ACK with an ECN counter increase that
+acknowledges packets from different paths MAY disable ECN marking and
+send all subsequent packets as Not-ECN capable.
+
 ## Using Multiple Packet Number Spaces
 
 If the multipath option is enabled with a value of 2, each path has
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -625,7 +625,7 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
-### Handling ECN Feedback
+### Handling ECN Feedback {#ecn-handling}
 
 ECN feedback in QUIC is provided based on counters in the ACK frame
 (see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK


draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,26 @@ delay will most of the time relate to the path with the shortest latency.
...
+
+ECN feedback in QUIC is provided based on counters in the ACK frame
+(see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK
+frame acknowledges multiple packets, the ECN feedback cannot be accounted
+to a specific packet. Respectively, if an ACK frames acknowledges multiple
+packets from different paths, the ECN feedback cannot unambiguously be assigned to
+a path. Therefore if one of the ECN counters increases, especially the CE
+counter, hosts SHOULD only acknowlegde packets from the same path where
+the ECN mark was observed in the same ACK frame.
+
```
huitema, line 638: I would change to:

Respectively, if an ACK frames acknowledges multiple
packets from different paths, the ECN counters provide the sum of all signals received on those paths.

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,26 @@ delay will most of the time relate to the path with the shortest latency.
...
+If a host receives an
+ACK frame that increases the CE counter and contains packets that have
+been sent on different paths, it MUST treat the CE marking as if it was
+received on either of the path and therefore consider it as a congestion
+indication for all paths with packets that have been acknowledges in that
+ACK frame. An host that receives an ACK with an ECN counter increase that
+acknowledges packets from different paths MAY disable ECN marking and
+send all subsequent packets as Not-ECN capable.
+
```
huitema, line 647: Yes. There are many cases in which an aggregate counter is just fine -- this is indeed what we have in QUIC V1, in which ECN counters count the marks received on successive paths, or even the marks received in response to probe packets. We want to say that implementers should be cautious when using this kind of ECN feedback. But there are many scenarios in which this is just fine, e.g., when the bulk of traffic is sent on a single path and the other path is used for ACKs.

huitema (changes requested):

draft-ietf-quic-multipath.md
mirjak: not sure if that is better, because the problem is that you don't know anymore to which path the CE belong when you received the ACK. I think your text proposal makes that point less clear.

draft-ietf-quic-multipath.md
mirjak: I agree but not sure what this mean for this PR. Do you think the text is okay as it is, or do we need to add additional notes. If so, what?

draft-ietf-quic-multipath.md
huitema: I think the acknowledgement process should be completely straightforward:

* when ECN mark received, increment the counters for the number space of the packet
* when sending ACK ECN, send the ECN for the number space of the ACK

Yes, in the case of single number space, that does mean single counter for all paths, and that makes the signal harder to use in congestion control. Implementers should be warned about that. But notice that the ACK-ECN are sent by the receiver, and the congestion control decisions are made by the sender. The sender does not have the option of "not receiving ACK-ECN" -- there is no parameter negotiation for that. The sender has different options:

* just not use ECN, i.e., send all packets with ECN flags "00".
* use ECN on only some path, e.g., the one path on which most data is sent, mark packets on this path with ECT-0 or ECT-1, mark packets on the other paths as "00".
* use ECT-0 on one path, ECT-1 on the other, and use some logic to attribute ECN counts to either ECT-0 or ECT-1.
* use ECT-0 or ECT-1 on all paths, and use correlation with other factors (e.g., losses, delays) to allocate the ECN marks to different paths.

I would not preclude any of these options.

mirjak: I think it would be better to discuss this in the issue #87 and not this PR but let me reply directly to a couple of points from you:

First, as you note this problem only exists for the single PNS solution. Therefore please note that this PR proposes to add this text to the respective section for single PNS considerations. We can add more text about general ECN but that shouldn't be much different from normal QUIC operation and would be a separate PR for a different section of the document.  

> Yes, in the case of single number space, that does mean single counter for all paths, and that makes the signal harder to use in congestion control.

I don't think "harder to use" is right. If multiple packets from different paths are ack'ed in the same frame it's simply impossible to assign the signal to one path. As you never should just ignore a congestion signal, the most conservative but probably only correct reaction is unfortunately to reduce your cwnd on all paths where packet got ack'ed. 

> just not use ECN, i.e., send all packets with ECN flags "00".

Yes, that is an option and I can add this. This was actually my first thought (see my reply in issue #87). However, for me that would be a strong argument against the single PNS solution. But then I thought that if the receiver would send separate ACKs per path that could also address the problem and keeping ACKs separated per path (at least if the sender sends blocks of PNs on each path) might be beneficial anyway. I think that also something that Ian had in mind.

> use ECN on only some path, e.g., the one path on which most data is sent, mark packets on this path with ECT-0 or ECT-1, mark packets on the other paths as "00".

That means you could only see ECN on one path. I don't think this an appropriate solution and again would be reason for me to not go for the single PNS solution.

> use ECT-0 on one path, ECT-1 on the other, and use some logic to attribute ECN counts to either ECT-0 or ECT-1.

I think this is out of scope for this document as it makes a different use of ECT1 against the recommendation in RFC8311 (and in conflict with L4S). So if you want to propose this, you have to go to tsvwg.

> use ECT-0 or ECT-1 on all paths, and use correlation with other factors (e.g., losses, delays) to allocate the ECN marks to different paths.

I don't think this could work. You don't have per packet delay information in the ACK, so that exactly the same ambiguity. And ideally loss should not be correlated with ECN - hopefully you have ECN instead of loss but that also not guaranteed as you might have multiple bottlenecks or other reason for loss.

draft-ietf-quic-multipath.md
mirjak, line 647:
```suggestion

Alternatively, a sender MAY decide to disable ECN entirely when a single
packet number space is used but marking all sent packet as Not-ECN capable.

```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,26 @@ delay will most of the time relate to the path with the shortest latency.
...
+packets from different paths, the ECN feedback cannot unambiguously be assigned to
+a path. Therefore if one of the ECN counters increases, especially the CE
+counter, hosts SHOULD only acknowlegde packets from the same path where
+the ECN mark was observed in the same ACK frame.
+
+If a host receives an
+ACK frame that increases the CE counter and contains packets that have
+been sent on different paths, it MUST treat the CE marking as if it was
+received on either of the path and therefore consider it as a congestion
+indication for all paths with packets that have been acknowledges in that
```
qdeconinck, line 643:
```suggestion
indication for all paths with packets that have been acknowledged in that
```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,26 @@ delay will most of the time relate to the path with the shortest latency.
...
+ACK frame. An host that receives an ACK with an ECN counter increase that
+acknowledges packets from different paths MAY disable ECN marking and
```
qdeconinck, line 645:
```suggestion
ACK frame. An host that receives an ACK with an ECN counter increase
acknowledging packets from different paths MAY disable ECN marking and
```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,26 @@ delay will most of the time relate to the path with the shortest latency.
...
 
+### Handling ECN Feedback
+
+ECN feedback in QUIC is provided based on counters in the ACK frame
+(see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK
+frame acknowledges multiple packets, the ECN feedback cannot be accounted
+to a specific packet. Respectively, if an ACK frames acknowledges multiple
+packets from different paths, the ECN feedback cannot unambiguously be assigned to
+a path. Therefore if one of the ECN counters increases, especially the CE
+counter, hosts SHOULD only acknowlegde packets from the same path where
```
qdeconinck, line 636:
```suggestion
counter, hosts SHOULD only acknowledge packets from the same path where
```

qdeconinck (approved):
The proposed text looks fine to me (with some minor rephrasing and typo suggestions).

draft-ietf-quic-multipath.md
goelvidhi: There is no explanation about why a host MAY disable ECN when an ACK acknowledges packets from different paths. It would be good to add some reasoning for that.

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,29 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
+### Handling ECN Feedback
```
goelvidhi, line 628: This text seems to be addressing the generation of ECN counters at the data receiver. How about, "Generating ACK with ECN feedback"?

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,29 @@ delay will most of the time relate to the path with the shortest latency.
...
+the ECN mark was observed in the same ACK frame.
+
+If a host receives an
+ACK frame that increases the CE counter and contains packets that have
+been sent on different paths, it MUST treat the CE marking as if it was
+received on either of the path and therefore consider it as a congestion
+indication for all paths with packets that have been acknowledged in that
+ACK frame. An host that receives an ACK with an ECN counter increase
+acknowledging packets from different paths MAY disable ECN marking and
+send all subsequent packets as Not-ECN capable.
```
mirjak, line 646:
```suggestion
ACK frame. Reducing the sending rate of all paths, even if congestion presumably
occurred only one of the paths, will likely impact performance negatively. 
Alternatively to avoid this performance penalty, a host that receives an ACK 
with an ECN counter increase
acknowledging packets from different paths MAY disable ECN marking and
send all subsequent packets as Not-ECN capable.
```

draft-ietf-quic-multipath.md
mirjak, line 628:
```suggestion
### Handling ECN Feedback {#ecn-handling}
```

PR #98, opened 2022-03-03
cc must be per path

### LINKED ISSUE #89: Congestion Control, RTT measurements and PMTU discovery should be per-path
The sentence above is part of the introduction, but the should is misleading for congestion control at least (we have a MUST for per-path congestion control later).

I would suggest to split the above sentence in two parts:
-  Congestion Control must be performed per-path 
-  Path MTU should be performed per path

We should probably list somewhere in the document the state that must/should be maintained on a per-path basis and the state that is related to the connection. This could clarify some of the mechanisms.

Concerning PathMTU, I would suggest to stay with a should and not a must since an implementation could simply track the minimum MTU among the different paths without maintaining per-path MTU. Such an implementation would work and if the MTUs do not differ significantly, it would probably be simpler than an implementation than maintains per-path MTU and needs to adjust the packets based on the path where each packet is sent.

Concerning RTT, we discuss the problem of measuring the RTT later with the packet number spaces discussion. Furthermore, some congestion control schemes need an accurate per-path RTT while others do not strictly require rtt estimations.

mirjak:
resolves #89

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -101,8 +101,10 @@ migration.
   * Use the same packet header formats as QUIC version 1 to avoid the
 risk of packets being dropped by middleboxes (which may only support
 QUIC version 1)
-  * Congestion Control, RTT measurements and PMTU discovery should be
-per-path (following {{QUIC-TRANSPORT}})
+  * Congestion Control must be per-path (following {{QUIC-TRANSPORT}})
+which usually also requires per-path RTT measurements
+  *  PMTU discovery should be performed per-path, however, the minimum
+PMTU of all paths may be used on each paths
   * A path is determined by the 4-tuple of source and destination IP
 address as well as source and destination port. Therefore there can be
 at most one active paths/connection ID per 4-tuple.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -103,7 +103,7 @@ risk of packets being dropped by middleboxes (which may only support
 QUIC version 1)
   * Congestion Control must be per-path (following {{QUIC-TRANSPORT}})
 which usually also requires per-path RTT measurements
-  *  PMTU discovery should be performed per-path
+  * PMTU discovery should be performed per-path
   * A path is determined by the 4-tuple of source and destination IP
 address as well as source and destination port. Therefore there can be
 at most one active paths/connection ID per 4-tuple.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -103,8 +103,7 @@ risk of packets being dropped by middleboxes (which may only support
 QUIC version 1)
   * Congestion Control must be per-path (following {{QUIC-TRANSPORT}})
 which usually also requires per-path RTT measurements
-  *  PMTU discovery should be performed per-path, however, the minimum
-PMTU of all paths may be used on each paths
+  *  PMTU discovery should be performed per-path
   * A path is determined by the 4-tuple of source and destination IP
 address as well as source and destination port. Therefore there can be
 at most one active paths/connection ID per 4-tuple.


draft-ietf-quic-multipath.md
```diff
@@ -101,8 +101,10 @@ migration.
   * Use the same packet header formats as QUIC version 1 to avoid the
 risk of packets being dropped by middleboxes (which may only support
 QUIC version 1)
-  * Congestion Control, RTT measurements and PMTU discovery should be
-per-path (following {{QUIC-TRANSPORT}})
+  * Congestion Control must be per-path (following {{QUIC-TRANSPORT}})
+which usually also requires per-path RTT measurements
+  *  PMTU discovery should be performed per-path, however, the minimum
+PMTU of all paths may be used on each paths
```
huitema, line 107: I would remove "however, the minimum PMTU of all paths may be used on each paths". Implementations can always send packets shorter than the path MTU, so we are not saying much here. Also, there is a minimum path MTU for QUIC, and implementations that don't do PMTUD can always use that, so again, no need to be too specific.

huitema (commented):
Please remove unnecessary text.

draft-ietf-quic-multipath.md
mirjak, line 107:
```suggestion
  *  PMTU discovery should be performed per-path
```

yfmascgy: If I remember correctly, the purpose of the sentence, "minimum PMTU of all paths may used ..." is to say when forming a packet before inserting it in the send_queue, it's better to constrain the size of the packet so that it can be scheduled to any of the paths for the sake of flexibility in scheduling. For example, when we want to use re-injection in our packet scheduling, we need to make sure that a packet sent on path 1 can also be reinjected on path 2. Therefore, what we did was to use per-path PMTU discovery and the use the minimum PMTU of all paths for each individual path. But indeed, this choice may vary in a different implementation.

huitema: That's completely implementation specific. Picoquic, for example, schedules the path first and then formats the packet. There is no need to go to that level of detail in the specification.

mirjak: @yfmascgy yes the intention of adding this what to give a hit towards the implementation option you mentioned. However, as you both say this is implementation specific and doesn't need to be in the intro part. So I'm fine removing this and already create a change request for this where I only need to push "commit".

If we want to discussion these implementation options more, we could add text in the "implementation consideration" section. @yfmascgy feel free to prose a PR for that or we could open a new issue to address this later.

draft-ietf-quic-multipath.md
```diff
@@ -101,8 +101,9 @@ migration.
   * Use the same packet header formats as QUIC version 1 to avoid the
 risk of packets being dropped by middleboxes (which may only support
 QUIC version 1)
-  * Congestion Control, RTT measurements and PMTU discovery should be
-per-path (following {{QUIC-TRANSPORT}})
+  * Congestion Control must be per-path (following {{QUIC-TRANSPORT}})
+which usually also requires per-path RTT measurements
+  *  PMTU discovery should be performed per-path
```
qdeconinck, line 106:
```suggestion
  * PMTU discovery should be performed per-path
```

qdeconinck (approved):

obonaventure (approved):

draft-ietf-quic-multipath.md
yfmascgy: Yes. I agree removing this from the intro and we can discuss this in the implementation consideration.

yfmascgy (approved):

draft-ietf-quic-multipath.md
yfmascgy: @mirjak I have submitted PR #105 for the PMTU implementation considerations.

Yanmei-Liu (approved):

PR #99, opened 2022-03-03
transport parameters in RFC9000 are unchanged

### LINKED ISSUE #91: Existing transport parameters
RFC9000 defines a list of transport parameters.

0x01 | max_idle_timeout | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x02 | stateless_reset_token | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x03 | max_udp_payload_size | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x04 | initial_max_data | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x05 | initial_max_stream_data_bidi_local | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x06 | initial_max_stream_data_bidi_remote | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x07 | initial_max_stream_data_uni | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x08 | initial_max_streams_bidi | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x09 | initial_max_streams_uni | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x0a | ack_delay_exponent | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x0b | max_ack_delay | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x0c | disable_active_migration | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x0d | preferred_address | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x0e | active_connection_id_limit | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x0f | initial_source_connection_id | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG] |  
0x10 | retry_source_connection_id | permanent | [RFC9000, Section 18.2] | 2021-02-11 | IETF | [QUIC_WG]

These transport parameters apply to all the paths that compose a Multipath QUIC connection. We should probably confirm this in the mpquic draft to avoid any misinterpretation (e.g. for max_idle_timeout or ack delays). There are two transport parameters that could warrant some discussion:

- disable_active_migration : This transport parameter is incompatible with enable_multipath. An endpoint that receives both this transport parameter and a non zero enable_multipath parameter MUST treat this as a connection error of type
MP_CONNECTION_ERROR and close the connection.
- preferred_address: RFC9000, section 9.2.1 contains the following:

   Once the handshake is confirmed, the client SHOULD select one of the two addresses provided by the server and initiate path    
  validation (see Section 8.2). A client constructs packets using any previously unused active connection ID, taken from either the 
  preferred_address transport parameter or a NEW_CONNECTION_ID frame.

  As soon as path validation succeeds, the client SHOULD begin sending all future packets to the new server address using the 
  new connection ID and discontinue use of the old server address. If path validation fails, the client MUST continue sending all 
  future packets to the server's original IP address.

For multipath, do we want to be stronger than SHOULD select one ? Can a client create news paths towards the initial address or only to one of the preferred addresses ? Should the client use both addresses (IPv4 and IPv6) if non-empty ?

mirjak:
resolves #91

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -240,7 +240,7 @@ Inline with the definition in {{QUIC-TRANSPORT}} disable_active_migration
 also disables multipath support, except "after a client has acted on a
 preferred_address transport parameter" {{Section 18.2. of QUIC-TRANSPORT}}.
 Further, it is out of scope for
-this document if the old address can also be used for multipath
+this document to specify if the old address can also be used for multipath
 after a client has migrated to the address provided in the preferred_address
 transport parameter. However, it SHOULD NOT be assumed that it is
 possible to use both addresses simultaneously without further confirmation
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -234,6 +234,18 @@ limits the number of concurrent paths. For the QUIC multipath extension
 this limit even applies when no connection ID is exposed in the QUIC
 header.
 
+This extension does not change the definition of any transport parameter
+defined in {{Section 18.2. of QUIC-TRANSPORT}}. 
+Inline with the definition in {{QUIC-TRANSPORT}} disable_active_migration
+also disables multipath support, except "after a client has acted on a
+preferred_address transport parameter" {{Section 18.2. of QUIC-TRANSPORT}}.
+Further, it is out of scope for
+this document if the old address can also be used for multipath
+after a client has migrated to the address provided in the preferred_address
+transport parameter. However, it SHOULD NOT be assumed that it is
+possible to use both addresses simultaneously without further confirmation
+from the other host.
+
 # Path Setup and Removal {#setup}
 
 After completing the handshake, endpoints have agreed to enable


huitema (approved):

obonaventure (changes requested):
There are two parts in this PR.

First, the following sentences:

This extension does not change the definition of any transport parameter
defined in {{Section 18.2. of QUIC-TRANSPORT}}. 
Inline with the definition in {{QUIC-TRANSPORT}} disable_active_migration
also disables multipath support, except "after a client has acted on a
preferred_address transport parameter" {{Section 18.2. of QUIC-TRANSPORT}}.

These sentences are at the right place. I would replace "any" by "all the" but that's a detail.

For the second part, I think that we should solve this problem differently. The current text is the following:

Further, it is out of scope for
this document if the old address can also be used for multipath
after a client has migrated to the address provided in the preferred_address
transport parameter. However, it SHOULD NOT be assumed that it is
possible to use both addresses simultaneously without further confirmation
from the other host.

I would suggest to discuss somewhere in the document how the client learns the server address and how preferred address is used. Since we do not have a way for the server to convey the list of "multipath-capable" addresses in this document, I would suggest a simple approach:
- When preferred address is not used, then the client can only establish multipath flows to the server address
- When preferred address is used, then if a client migrates to one of the preferred addresses according to 9.6 of RFC9000, then it must consider this address as the endpoint of all multipath flows

I guess that we'll need to define later a transport parameter that provides the list of "multipath-capable" addresses of a given server. In the mean time, we should opt for a simple solution.

mirjak:
@obonaventure you wrote:

> When preferred address is not used, then the client can only establish multipath flows to the server address
> When preferred address is used, then if a client migrates to one of the preferred addresses according to 9.6 of RFC9000, then it must consider this address as the endpoint of all multipath flows

That sounds like you want to restrict the multipath extension to only one server address (but potentially different paths from different clients address)...?

I think this restriction is unnecessary. Yes we don't specify a way to learn server addresses but that doesn't means that the client doesn't know (e.g. based on higher layer information) multiple addresses of the server. What should we restrict that?

mirjak:
However, can we maybe merge this PR as it is for now and you open a separate issue for your additional point @obonaventure ?

draft-ietf-quic-multipath.md
```diff
@@ -234,6 +234,18 @@ limits the number of concurrent paths. For the QUIC multipath extension
 this limit even applies when no connection ID is exposed in the QUIC
 header.
 
+This extension does not change the definition of any transport parameter
+defined in {{Section 18.2. of QUIC-TRANSPORT}}. 
+Inline with the definition in {{QUIC-TRANSPORT}} disable_active_migration
+also disables multipath support, except "after a client has acted on a
+preferred_address transport parameter" {{Section 18.2. of QUIC-TRANSPORT}}.
+Further, it is out of scope for
+this document if the old address can also be used for multipath
```
qdeconinck, line 243:
```suggestion
this document to specify if the old address can also be used for multipath
```

qdeconinck (approved):
The proposed PR sounds good to me, with the slight addition I mentioned.

Yanmei-Liu (approved):

obonaventure:
Mirja,

> > When preferred address is not used, then the client can only establish multipath flows to the server address
> > When preferred address is used, then if a client migrates to one of the preferred addresses according to 9.6 of RFC9000, then it must consider this address as the endpoint of all multipath flows
> 
> That sounds like you want to restrict the multipath extension to only one server address (but potentially different paths from different clients address)...?

That was my understanding of the sentences in the PR. Maybe I got it wrong.

> 
> I think this restriction is unnecessary. Yes we don't specify a way to learn server addresses but that doesn't means that the client doesn't know (e.g. based on higher layer information) multiple addresses of the server. What should we restrict that?

I think that we should discuss the role of preferred address in section 3 where we explain how paths can be opened.

obonaventure:
> However, can we maybe merge this PR as it is for now and you open a separate issue for your additional point @obonaventure ?

Fine, let's do it this way.

obonaventure (approved):

yfmascgy (approved):

Issue #96, opened 2022-03-03
Choosing between a single packet number space vs. multiple packet number spaces


qdeconinck:
This draft initially originates from a merging effort of previous Multipath proposals. While many points were found to be common between them, there remains one design point that still requires consensus: the number of packet number spaces that a Multipath QUIC connection should support (i.e., one for the whole connection vs. one per path).

The current draft enables experimentation with both variants, but in the final version we will certainly need to choose between one of the versions.

huitema:
The big debate is for the configuration with NULL CID on client, long CID on server. The packets from server to client do not carry a CID, and only the last bytes of the sequence number. The client will need some logic to infer the full sequence number before decrypting the packet. The client could maybe use the incoming 5 tuple as part of the logic, but it is not obvious. It is much simpler to assume a direct map from destination CID to number space. That means, if a peer uses a NULL CID, all packets sent to that peer are in the same number space.

huitema:
I think the above points to a possible "unified solution":

* if multipath negotiated, tie number spaces to destination CID, use default (N=0) if NULL CID.
* use multipath ack; in multipath ACK, identify number space by sequence number of DCID in received packets. Default that number to zero if received with NULL CID.
* in ABANDON path, identify path either by DCID of received packets, SCID of sent packet, or "this path" (sending path) if CID is NULL.

huitema:
If the APi does not support 96 bit sequence numbers, it should always be possible to create an encryption context per number space, using Key=current key and ID = current-ID + CID sequence number. Of course, that forces creation of multiple context, and that makes key rotation a bit harder to manage. But still, that should be possible.

mirjak:
Thanks for the summary @huitema. I think one point is missing in your list which is related to issue #87. Use of a single packet number space might not support ECN.

Regarding the unified solution: I think what you actually say is that we would specify both solutions and everybody would need to implement both logics. At least regarding the "simplicity of code" argument, that would be the worst choice.

If we can make the multiple packet number spaces solution work with one-sided CIDs, I'm tending toward such an approach. Use of multiple packet number spaces avoids ambiguity/needed "smartness" in ACK handling and packet scheduling which, as you say above, can make the implementation more complex and, moreover,  wrong decisions may have large impact on efficient (both local processing and on-the-wire). I don't think we want to leave these things to each implementation individually.

qdeconinck:
[...]

Issue #101, opened 2022-03-04
Server MUST validate paths before using them


obonaventure:
In section 3.1, we wrote 

When the multipath option is negotiated, clients that want to use an additional path MUST first initiate the Address Validation procedure with PATH_CHALLENGE and PATH_RESPONSE frames described in [Section 8](https://rfc-editor.org/rfc/rfc9000#section-8) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)]. After receiving packets from the client on the new paths, the servers MAY in turn attempt to validate these paths using the same mechanisms.[¶](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-3.1-1)

With multipath, we cannot simply write that the server MAY validate the new path. There is a risk of a malicious client that operates as follows:
- client creates path from its own address, A. Path is validate by server
- client starts long download from server
- client creates a second path to server using a spoofed address, B 
- client starts address validation for B but does not care about the response from the server
- client stops acknowledging packets over the first path to force the server to move packets to the second one and then opportunistically acknowledges packets that should have been sent by the server over that second packet (SPNS would probably make such an attack simpler than MPNS)

We probably need to start working on the security considerations section of the draft.

huitema:
This is already covered in the QUIC threat model. The server is supposed to validate new paths by sending a "path challenge" on that path, then waiting for a challenge response. If the client spoofs a path, it will never receive the challenge, and thus the server will never receive the response and will not validate the path. See section 8.2 of RFC 9000, path validation. Security section 21.1.3. Connection Migration states that "Path validation ([Section 8.2](https://www.rfc-editor.org/rfc/rfc9000.html#migrate-validate)) establishes that a peer is both willing and able to receive packets sent on a particular path."

huitema:
So yes, the current text should be "the servers MUST in turn attempt to validate these paths using the same mechanisms, as specified in {{ Section 8.2 of RFC9000 }}."

mirjak:
No, this should not be a MUST. It is true that the path MUST be validated before it can be used. However, just because the serve receives a new packet on a path, that doesn't automatically mean that the server has to start path validation. It could e.g. also decide to simply withdraw that packet.

mirjak:
Actually, this is also not right.

This is what RFC9000 says (section 9):

[...]

PR #103, opened 2022-03-04
First shot at a unified proposal

huitema:
This derives from the discussions on issue #96

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -137,7 +137,7 @@ possible to use different connection IDs for distinguishing packets
 sent to that node over different paths. All packets sent to a zero-length
 CID are numbered in the same number space. Out of order delivery is likely,
 which causes inflation of the number of acknowledgement ranges and of the
-the size of ACK frames. This proposal specifies algorithms forf{#
+the size of ACK frames. This proposal specifies algorithms for
 controlling the size of acknowledgement packets. Senders that accept to
 use multiple paths when sending to a node using zero-length CID will
 have to use special logic to minimize the impact of multipath
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -595,7 +595,7 @@ be controlled by the combination of one or several of the following:
    uses a series of consecutive sequence numbers without creating
    holes.
 
-### Zero-Length CID Loss and Congestion
+### Loss and Congestion Handling With Zero-Length CID {#zero-length-cid-loss-and-congestion}
 
 When sending to a zero-length CID
 receiver, senders will receive back acknowledgements that combine packet
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -629,9 +629,11 @@ about using the ACK Delay field of ACK frames, and
 {{ecn-and-zero-length-cid-considerations}} for issues on using ECN marks.)
 
 Loss detection as specified in {{QUIC-RECOVERY}} uses algorithms
-based on timers and on sequence numbers. When sending to zero-length CID receivers,
-senders expect that packets sent on different paths will not be received in
-order. They cannot directly use the packet sequence numbers to
+based on timers and on sequence numbers. When packets are sent over
+multiple paths, loss detection must be adapted to allow for different RTTs
+on different paths. When sending to zero-length CID receivers, packets sent
+on different paths may not be received in order. Therefore senders cannot
+directly use the packet sequence numbers to
 compute the Packet Thresholds defined in {{Section 6.1.1 of QUIC-RECOVERY}}.
 Relying only on Time Thresholds produces correct results, but is somewhat
 suboptimal. Some implementations have been getting good results by
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -134,7 +134,7 @@ loss recovery and congestion control mechanisms defined in
 Some deployments of QUIC use zero-length connection IDs.
 When a node selects to use zero-length connection IDs, it is not
 possible to use different connection IDs for distinguishing packets
-sent to that node over different paths. All packets sent to a zero-length
+sent to that node over different paths. All packets sent with a zero-length
 CID are numbered in the same number space. Out of order delivery is likely,
 which causes inflation of the number of acknowledgement ranges and of the
 the size of ACK frames. This proposal specifies algorithms for
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -550,7 +550,7 @@ packets sent on path 0.
 
 ## Using Zero-Length connection ID
 
-If a node elects to use zero-length connection ID, it MUST implement the
+If a node elects to use zero-length connection IDs, it MUST implement the
 handling of Acknowledgements defined in {{sending-acknowledgements-and-handling-ranges}}.
 The peer of a node using zero-length connection ID SHOULD implement specific
 logic to handle loss detection in the presence of multiple paths,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -210,7 +210,7 @@ PN spaces, available option values are listed in
 Option | Definition
 ---------|---------------------------------------
 0x0      | don't support multipath
-0x1      | supports multipath as defined in this proposal
+0x1      | supports multipath as defined in this document
 {: #param_value_definition title="Available value for enable_multipath"}
 
 If for any one of the endpoints the parameter is absent or set to 0,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -132,7 +132,7 @@ loss recovery and congestion control mechanisms defined in
 {{QUIC-RECOVERY}}.
 
 Some deployments of QUIC use zero-length connection IDs.
-When a node elects to use zero-length connection IDs, it is not
+When a node selects to use zero-length connection IDs, it is not
 possible to use different connection IDs for distinguishing packets
 sent to that node over different paths. All packets sent to a zero-length
 CID are numbered in the same number space. Out of order delivery is likely,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -664,9 +664,9 @@ Of course, the hosts will only take limited advantage from the multipath
 capability in these scenarios. Support for "make before break" migrations
 will improve, but load sharing between multiple paths will not work.
 
-## Using Multiple Packet Number Spaces
+## Using non-zero length CID and Multiple Packet Number Spaces
 
-If the multipath option is enabled with a value of 2, each path has
+If packets contain a non-zero CID, each path has
 its own packet number space for transmitting 1-RTT packets and a new
 ACK frame format is used as specified in {{ack-mp-frame}}.
 Compared to the QUIC version 1 ACK frame, the ACK_MP frames additionally
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -568,6 +568,11 @@ as explained in {{restricted-sending-to-zero-length-cid-peer}}.
 
 ### Sending Acknowledgements and Handling Ranges
 
+If zero-length CID and therefore also a single packet number space
+is used by the sender, the receiver MAY send ACK frames instead
+of ACK_MPframes to reduce overhead as the additional path ID field
+will anyway always carry the same value.
+
 If senders decide to send packets on paths with different
 transmission delays, some packets will very likely be received out
 of order.  This will cause the ACK frames to carry multiple ranges of
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -141,7 +141,7 @@ the size of ACK frames. This proposal specifies algorithms for
 controlling the size of acknowledgement packets. Senders that accept to
 use multiple paths when sending to a node using NULL length CID will
 have to use special logic to minimize the impact of multipath
-delivery on loss detection and congestion controls. If the receiver
+delivery on loss detection and congestion control. If the receiver
 reports ECN counters, they will contain the sum of marks received on all
 paths. Senders will need specific logic to handle such ECN marks.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -648,10 +648,14 @@ as described in {{QUIC-Timestamp}}.
 ### ECN and Zero-Length CID Considerations
 
 ECN feedback in QUIC is provided based on counters in the ACK frame
-(see {{Section  19.3.2. of QUIC-TRANSPORT}}). These counters are specific
-to a number space. When sending to zero-length CID receivers, the same number space is
-used for multiple paths, and the ECN feedback cannot unambiguously be assigned to
-a path.
+(see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK
+ frame acknowledges multiple packets, the ECN feedback cannot be accounted
+ to a specific packet.
+
+There are separate counters for each packet number space. However, sending
+to zero-length CID receivers, the same number space is used for multiple paths.
+Respectively, if an ACK frames acknowledges multiple packets from different paths,
+the ECN feedback cannot unambiguously be assigned to a path.
 
 If the sender marks its packets with the ECN capable flags, the network
 will expect standard reactions to ECN marks, such as slowing down
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -137,13 +137,16 @@ possible to use different connection IDs for distinguishing packets
 sent to that node over different paths. All packets sent to a zero-length
 CID are numbered in the same number space. Out of order delivery is likely,
 which causes inflation of the number of acknowledgement ranges and of the
-the size of ACK frames. This proposal specifies algorithms for
+the size of ACK frames. This proposal specifies algorithms forf{#
 controlling the size of acknowledgement packets. Senders that accept to
 use multiple paths when sending to a node using zero-length CID will
 have to use special logic to minimize the impact of multipath
 delivery on loss detection and congestion control. If the receiver
 reports ECN counters, they will contain the sum of markings received on all
 paths. Senders will need specific logic to handle such ECN marks.
+(The logic required to support zero-length CID at receivers and senders
+is discussed in {{using-zero-length}}.)
+
 
 This proposal does not cover address discovery and management. Addresses
 and the actual decision process to setup or tear down paths are assumed
@@ -548,7 +551,7 @@ ACK frames are received in 1-RTT packets while the state of multipath
 negotiation is ambiguous, they MUST be interpreted as acknowledging
 packets sent on path 0.
 
-## Using Zero-Length connection ID
+## Using Zero-Length connection ID {#using-zero-length}
 
 If a node elects to use zero-length connection IDs, it MUST implement the
 handling of Acknowledgements defined in {{sending-acknowledgements-and-handling-ranges}}.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -637,9 +637,11 @@ a path.
 
 If the sender marks its packets with the ECN capable flags, the network
 will expect standard reactions to ECN marks, such as slowing down
-transmission on the sending path. As the sending path is ambiguous,
-the sender must treat a CE marking as a congestion signal on all sending
-paths.
+transmission on the sending path. If zero-length CID is used, 
+the sending path is however ambiguous. Therefore, the sender MUST
+treat a CE marking as a congestion signal on all sending paths that
+have been by a packet that was acknowledged in the ACK frame signaling
+the CE counter increase.
 
 A host that is sending over multiple paths to a zero-length CID receiver
 MAY disable ECN marking and
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -645,7 +645,7 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
-### ECN and Zero-Length CID Considerations
+### ECN and Zero-Length CID Considerations {#ecn-handing}
 
 ECN feedback in QUIC is provided based on counters in the ACK frame
 (see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -602,7 +602,7 @@ be controlled by the combination of one or several of the following:
 ### Loss and Congestion Handling With Zero-Length CID {#zero-length-cid-loss-and-congestion}
 
 When sending to a zero-length CID
-receiver, senders will receive back acknowledgements that combine packet
+receiver, senders may receive acknowledgements that combine packet
 numbers received over multiple paths. Senders MUST be able to infer the
 sending path from the acknowledged packet numbers, for example by remembering
 which packet was sent on what path. The senders MUST use that information to
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -552,17 +552,30 @@ packets sent on path 0.
 
 ## Using Zero-Length connection ID {#using-zero-length}
 
-If a node elects to use zero-length connection IDs, it MUST implement the
-handling of Acknowledgements defined in {{sending-acknowledgements-and-handling-ranges}}.
-The peer node using zero-length connection IDs SHOULD implement a
-logic to handle loss detection in the presence of multiple paths,
-identify the path on which packets are acknowledged or determined
-lost for the purpose of congestion control as explained
-in {{zero-length-cid-loss-and-congestion}}, mitigate the ECN
-issues mentioned in {{ecn-and-zero-length-cid-considerations}}, and
-mitigate the RTT measurement issues explained
-in {{ack-delay-and-zero-length-cid-considerations}}. If a node
-does not support this logic, it MUST limit its use of multipath
+If a zero-length connection ID is used, one packet number space
+ for all paths. That means the packet sequence numbers are allocated
+ from the common
+ number space, so that, for example, packet number N could be sent
+ on one path and packet number N+1 on another.
+ 
+ In this case, ACK frames report the numbers of packets that have been
+ received so far,
+ regardless of the path on which they have been received. That means
+ the senders needs to maintain an association between sent packet numbers
+ and the path over which these packets were sent. This is necessary
+ to implement per path congestion control, as explained
+in {{zero-length-cid-loss-and-congestion}}.
+
+Further, the receiver of packets with zero-length connection IDs should
+implement handling of acknowledgements as defined in
+{{sending-acknowledgements-and-handling-ranges}}.
+
+ECN handing is specified in {{ecn-handling}}, and
+mitigation of the RTT measurement is further explained
+in {{ack-delay-and-zero-length-cid-considerations}}. 
+
+If a node
+does not want to implement this logic, it MAY instead limit its use of multiple paths
 as explained in {{restricted-sending-to-zero-length-cid-peer}}.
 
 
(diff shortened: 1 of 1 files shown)


Yanmei-Liu (approved):
Look good to me.

draft-ietf-quic-multipath.md
```diff
@@ -254,7 +248,7 @@ When the multipath option is negotiated, clients that want to use an
 additional path MUST first initiate the Address Validation procedure
 with PATH_CHALLENGE and PATH_RESPONSE frames described in
 {{Section 8 of QUIC-TRANSPORT}}. After receiving packets from the
-client on the new paths, the servers MAY in turn attempt to validate
+client on the new paths, the servers MUST in turn attempt to validate
```
qdeconinck, line 251: This seems more related to the discussion in #101, not sure if it should be part of this PR.

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +592,36 @@ be controlled by the combination of one or several of the following:
    uses a series of consecutive sequence numbers without creating
    holes.
 
-### ACK Delay Considerations
+### NULL-CID-Loss-and-Congestion
+
+Implementations MUST implement per-path congestion control. Congestion
```
qdeconinck, line 597: I don't think per-path congestion control is specific to NULL CID case.

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +592,36 @@ be controlled by the combination of one or several of the following:
...
+control regulates the amount of data in flight and the pacing of packets
+using estimates of path characteristics. The estimates can use a variety
+of signals, such as packet acknowledgements, loss detection, or measurements
+of round trip time. These events MUST be attributed to the path over
+which the corresponding packets were sent. When sending to a NULL CID
```
qdeconinck, line 602: Actually, the beginning of this paragraph is not specific to the NULL CID case.

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +592,36 @@ be controlled by the combination of one or several of the following:
...
+receiver, senders will receive back acknowledgements that combine packet
+numbers received over multiple paths. Senders MUST be able to infer the
+sending path from the acknowledged packet numbers, for example by remembering
+which packet was sent to what path. The MUST use that information to
```
qdeconinck, line 606:
```suggestion
which packet was sent to what path. The senders MUST use that information to
```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +630,41 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
+### ECN and NULL CID Considerations
+
+ECN feedback in QUIC is provided based on counters in the ACK frame
+(see {{Section  19.3.2. of QUIC-TRANSPORT}}). These counters are specific
+to a number space. When sending to NULL CID receivers, teh same number space is
```
qdeconinck, line 637:
```suggestion
to a number space. When sending to NULL CID receivers, the same number space is
```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +630,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+
+Hosts that are designed to support multipath using multiple number spaces
+MAY adopt a conservative posture after negotiating multipath support with
+a peer using NULL CID. The simplest posture is to elect to only send
+data on one path at a time, while accepting packets on all acceptable
+paths. In that case:
+
+* the attribution of packets to path discussed in {{null-cid-loss-and-congestion}}
+  are easy to solve because packets are sent on a single path,
+* the ACK Delays are mostly correct,
```
qdeconinck, line 661: why "mostly"?

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +592,36 @@ be controlled by the combination of one or several of the following:
    uses a series of consecutive sequence numbers without creating
    holes.
 
-### ACK Delay Considerations
+### NULL-CID-Loss-and-Congestion
```
qdeconinck, line 595:
```suggestion
### Congestion Control, Loss Recovery and NULL CID Considerations
```

qdeconinck (commented):
I believe this PR is a good first proposal for such a unified solution. If such a solution can fit all the use cases, I'm happy to further work on this.

draft-ietf-quic-multipath.md
huitema: Correct. Need to state that correctly. The point is, in the absence of CID, implementations must track the path over which the packet loss or the RTT measurement occurred, because the number space is shared.

draft-ietf-quic-multipath.md
huitema: Yes.

draft-ietf-quic-multipath.md
huitema: :-)
need to find another way of saying that without a number space per path this is harder to get right.

draft-ietf-quic-multipath.md
huitema: Let's first apply @mirjak's suggestion to replace NULL by zero-length...

draft-ietf-quic-multipath.md
huitema: Removed "mostly"

draft-ietf-quic-multipath.md
huitema: Fixed.

draft-ietf-quic-multipath.md
huitema: Fixed, per next comment.

draft-ietf-quic-multipath.md
huitema: Reverted.

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
huitema: Fixed

huitema:
The commit [a230bc1](https://github.com/quicwg/multipath/pull/103/commits/a230bc17db0eabf78dffb6e0d02224ad851197da) applies the changes suggested in the previous reviews. With those changes, I think we have a good basis for discussing a unified solution in the working group.

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,26 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+When multipath is negotiated,
+each destination connection ID is linked to a separate packet number space.
+Using multiple packet number spaces enables direct use of the
+loss recovery and congestion control mechanisms defined in
+{{QUIC-RECOVERY}}.
+
+Some deployments of QUIC use zero-length connection ID.
+When a node elects to use zero-length connection IDs, it is not
+possible to use different connections IDs for distinguishing packets
+sent to that node over different paths. All packets sent to a NULL
```
qdeconinck, line 137:
```suggestion
sent to that node over different paths. All packets sent to a zero-length
```

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,26 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+CID are numbered in the same number space. Out of order delivery is likely,
+which causes inflation of the number of acknowledgement ranges and of the
+the size of ACK frames. This proposal specifies algorithms for
+controlling the size of acknowledgement packets. Senders that accept to
+use multiple paths when sending to a node using NULL length CID will
```
qdeconinck, line 142:
```suggestion
use multiple paths when sending to a node using zero-length CID will
```

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,26 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+have to use special logic to minimize the impact of multipath
+delivery on loss detection and congestion controls. If the receiver
```
qdeconinck, line 144:
```suggestion
delivery on loss detection and congestion controller. If the receiver
```

qdeconinck (commented):
Some suggestions to remove NULL CID mentions, but overall this seems a good basis for further discussion.

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,26 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
-of multiple packet number spaces could be supported as well when
-the connection ID is only present in one direction.
+The proposal introduces support for multiple packet number spaces.
+When multipath is negotiated,
+each destination connection ID is linked to a separate packet number space.
+Using multiple packet number spaces enables direct use of the
+loss recovery and congestion control mechanisms defined in
+{{QUIC-RECOVERY}}.
+
+Some deployments of QUIC use zero-length connection ID.
```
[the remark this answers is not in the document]
boucadair: RFC9000 says the following:

```
  An endpoint SHOULD NOT initiate migration with a peer that has
   requested a zero-length connection ID, because traffic over the new
   path might be trivially linkable to traffic over the old one.
```

I wonder if the above behavior has implications on maintaining multiple paths when  zero-length connection IDs are in used.

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,26 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+sent to that node over different paths. All packets sent to a NULL
+CID are numbered in the same number space. Out of order delivery is likely,
+which causes inflation of the number of acknowledgement ranges and of the
+the size of ACK frames. This proposal specifies algorithms for
+controlling the size of acknowledgement packets. Senders that accept to
+use multiple paths when sending to a node using NULL length CID will
+have to use special logic to minimize the impact of multipath
+delivery on loss detection and congestion controls. If the receiver
+reports ECN counters, they will contain the sum of marks received on all
+paths. Senders will need specific logic to handle such ECN marks.
```
boucadair, line 146: Again, the "specific" thing is worth be exemplified.

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,26 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+possible to use different connections IDs for distinguishing packets
+sent to that node over different paths. All packets sent to a NULL
+CID are numbered in the same number space. Out of order delivery is likely,
+which causes inflation of the number of acknowledgement ranges and of the
+the size of ACK frames. This proposal specifies algorithms for
+controlling the size of acknowledgement packets. Senders that accept to
+use multiple paths when sending to a node using NULL length CID will
+have to use special logic to minimize the impact of multipath
+delivery on loss detection and congestion controls. If the receiver
+reports ECN counters, they will contain the sum of marks received on all
```
boucadair, line 145:
```suggestion
reports ECN counters, they will contain the sum of markings received on all
```

boucadair (changes requested):
Some very minor comments.

huitema:
Thanks @boucadair for the review. I have applied the editorial changes. You are asking for something more regarding the "specific logic" required when supporting zero-length CID. There are in fact two pieces to that: logic at the receiver, i.e., the node that chose to receive packets with zero-length CID; and logic at the sender, i.e., the node that accepts to send data on multiple paths towards a node that uses zero-length CID. This is explained in details in the section "Using Zero-Length connection ID", so I guess what we need is a reference to that section.

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,29 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+the size of ACK frames. This proposal specifies algorithms forf{#
+controlling the size of acknowledgement packets. Senders that accept to
+use multiple paths when sending to a node using zero-length CID will
+have to use special logic to minimize the impact of multipath
+delivery on loss detection and congestion control. If the receiver
+reports ECN counters, they will contain the sum of markings received on all
+paths. Senders will need specific logic to handle such ECN marks.
+(The logic required to support zero-length CID at receivers and senders
+is discussed in {{using-zero-length}}.)
+
```
huitema, line 150: This lines were added to address review by @boucadair

draft-ietf-quic-multipath.md
huitema: The new title is "Loss and Congestion Handling With Zero-Length CID". I hope it works.

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
huitema: The RFC 9000 text suggests that only the node that is choosing to use zero length CID will initiate a new path -- typically a client. That seems fine. Third parties will be able to observe packets flowing between client and server as follow:

* path 1: between address 1 and port 1 on client, and server address and server port on server.
* path 2: between address 2 and port 2 on client, and server address and server port on server.

The addresses and ports will be visible whether the client uses null-length CID or not. We may want to say to point out that using different port numbers for each path may remove 16 bits of possible correlation data, but that applies to RFC 9000 just as well.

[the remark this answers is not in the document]
boucadair: I agree this is OK in typical scenarios. 

I leave it up to you to decide if it is worth to include a mention about this in the draft. Thanks, @huitema.

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,29 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+loss recovery and congestion control mechanisms defined in
+{{QUIC-RECOVERY}}.
+
+Some deployments of QUIC use zero-length connection IDs.
+When a node elects to use zero-length connection IDs, it is not
+possible to use different connection IDs for distinguishing packets
+sent to that node over different paths. All packets sent to a zero-length
+CID are numbered in the same number space. Out of order delivery is likely,
+which causes inflation of the number of acknowledgement ranges and of the
+the size of ACK frames. This proposal specifies algorithms forf{#
```
qdeconinck, line 140:
```suggestion
the size of ACK frames. This proposal specifies algorithms for
```

qdeconinck (approved):
Small typo; for the remaining, LGTM.

draft-ietf-quic-multipath.md
```diff
@@ -554,39 +551,21 @@ ACK frames are received in 1-RTT packets while the state of multipath
...
+The peer node using zero-length connection IDs SHOULD implement a
+logic to handle loss detection in the presence of multiple paths,
+identify the path on which packets are acknowledged or determined
+lost for the purpose of congestion control as explained
+in {{zero-length-cid-loss-and-congestion}}, mitigate the ECN
+issues mentioned in {{ecn-and-zero-length-cid-considerations}}, and
+mitigate the RTT measurement issues explained
+in {{ack-delay-and-zero-length-cid-considerations}}. If a node
+does not support this logic, it MUST limit its use of multipath
+as explained in {{restricted-sending-to-zero-length-cid-peer}}.
```
mirjak, line 567: I don't think the MUST and SHOULD here make a lot of sense, given there are more/other normative language later in this section. Also ice think your removed too much.... here is an alternative proposal for this section intro part:
```suggestion
If a zero-length connection ID is used, one packet number space
 for all paths. That means the packet sequence numbers are allocated from the common
 number space, so that, for example, packet number N could be sent
 on one path and packet number N+1 on another.
 
 In this case, ACK frames report the numbers of packets that have been received so far,
 regardless of the path on which they have been received. That means
 the senders needs to maintain an association between sent packet numbers
 and the path over which these packets were sent. This is necessary
 to implement per path congestion control, as explained
in {{zero-length-cid-loss-and-congestion}}.

Further, the receiver of packets with zero-length connection IDs should
handling of acknowledgements defined in {{sending-acknowledgements-and-handling-ranges}}.

ECN handing is discussed in {{ecn-handling}}, and
mitigation of the RTT measurement is further explained
in {{ack-delay-and-zero-length-cid-considerations}}. 

If a node
does not support this logic, it MUST limit its use of multipath
as explained in {{restricted-sending-to-zero-length-cid-peer}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
+### ECN and Zero-Length CID Considerations
```
mirjak, line 631:
```suggestion
### ECN and Zero-Length CID Considerations {#ecn-handing}
```

mirjak (changes requested):
I made an editorial pass but am not done yet. @huitema maybe you can merge (or reject) already these changes and I finish my review tomorrow.

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+
+ECN feedback in QUIC is provided based on counters in the ACK frame
+(see {{Section  19.3.2. of QUIC-TRANSPORT}}). These counters are specific
+to a number space. When sending to zero-length CID receivers, the same number space is
+used for multiple paths, and the ECN feedback cannot unambiguously be assigned to
+a path.
```
mirjak, line 637:
```suggestion
ECN feedback in QUIC is provided based on counters in the ACK frame
(see {{Section  19.3.2. of QUIC-TRANSPORT}}). That means if an ACK
 frame acknowledges multiple packets, the ECN feedback cannot be accounted
 to a specific packet.

There are separate counters for each packet number space. However, sending to zero-length CID receivers, the same number space is used for multiple paths. Respectively, if an ACK frames acknowledges multiple
packets from different paths, the ECN feedback cannot unambiguously be assigned to
a path.
```

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+
+If the sender marks its packets with the ECN capable flags, the network
+will expect standard reactions to ECN marks, such as slowing down
+transmission on the sending path. As the sending path is ambiguous,
+the sender must treat a CE marking as a congestion signal on all sending
+paths.
```
mirjak, line 643:
```suggestion
transmission on the sending path. If zero-length CID is used, 
the sending path is however ambiguous. Therefore, the sender MUST
treat a CE marking as a congestion signal on all sending paths.
```

mirjak: Note that this change is not only editorial as I added the clause that you only MUST decrease those paths where you actually got a packet acknowledged. This is slightly more complicated but you should have saved the information where which packet was sent anyway and this allows for optimisation in the ACK creation. Given this is a MUST, I think we need this clause.

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+
+A host that is sending over multiple paths to a zero-length CID receiver
+MAY disable ECN marking and
+send all subsequent packets as Not-ECN capable.
+
+### Restricted Sending to Zero-Length CID Peer
+
+Hosts that are designed to support multipath using multiple number spaces
+MAY adopt a conservative posture after negotiating multipath support with
+a peer using zero-length CID. The simplest posture is to elect to only send
```
mirjak, line 653:
```suggestion
a peer using zero-length CID. The simplest posture is to only send
```

draft-ietf-quic-multipath.md
```diff
@@ -554,39 +551,21 @@ ACK frames are received in 1-RTT packets while the state of multipath
...
+lost for the purpose of congestion control as explained
+in {{zero-length-cid-loss-and-congestion}}, mitigate the ECN
+issues mentioned in {{ecn-and-zero-length-cid-considerations}}, and
+mitigate the RTT measurement issues explained
+in {{ack-delay-and-zero-length-cid-considerations}}. If a node
+does not support this logic, it MUST limit its use of multipath
+as explained in {{restricted-sending-to-zero-length-cid-peer}}.
+
 
 ### Sending Acknowledgements and Handling Ranges
```
mirjak, line 571:
```suggestion

If zero-length CID and therefore also a single packet number space
is used by the sender, the receiver MAY send ACK frames instead
of MP-ACK frames to reduce overhead as the additional path ID field
will anyway always carry the same value.

```

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +595,31 @@ be controlled by the combination of one or several of the following:
...
+perform congestion control on the relevant paths, and to correctly
+estimate the transmission delays on each path. (See
+{{ack-delay-and-zero-length-cid-considerations}} for specific considerations
+about using the ACK Delay field of ACK frames, and
+{{ecn-and-zero-length-cid-considerations}} for issues on using ECN marks.)
+
+Loss detection as specified in {{QUIC-RECOVERY}} uses algorithms
+based on timers and on sequence numbers. When sending to zero-length CID receivers,
+senders expect that packets sent on different paths will not be received in
+order. They cannot directly use the packet sequence numbers to
```
mirjak, line 614:
```suggestion
based on timers and on sequence numbers. When packets are sent over multiple paths,
loss detection must be adapted to allow for different RTTs on different paths.
When sending to zero-length CID receivers, packets sent on different paths may not be received in
order. Therefore senders cannot directly use the packet sequence numbers to
```

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +595,31 @@ be controlled by the combination of one or several of the following:
...
+compute the Packet Thresholds defined in {{Section 6.1.1 of QUIC-RECOVERY}}.
+Relying only on Time Thresholds produces correct results, but is somewhat
+suboptimal. Some implementations have been getting good results by
+remembering not just the path over which a packet was sent, but also
+the order of a packet on that path. They can then use this order in the
+Packet Threshold tests.
```
mirjak, line 620: I don't think I fully understand these last couple of sentences. Can you clarify?

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+* the ACK Delays are correct,
+* the vast majority of ECN marks relate to the current sending path.
+
+Of course, the hosts will only take limited advantage from the multipath
+capability in these scenarios. Support for "make before break" migrations
+will improve, but load sharing between multiple paths will not work.
+
 ## Using Multiple Packet Number Spaces
 
 If the multipath option is enabled with a value of 2, each path has
```
mirjak, line 668:
```suggestion
## Using non-zero length CID and Multiple Packet Number Spaces

If packets contain a non-zero CID, each path has
```

draft-ietf-quic-multipath.md
huitema: This is outdated.

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+Hosts that are designed to support multipath using multiple number spaces
+MAY adopt a conservative posture after negotiating multipath support with
+a peer using zero-length CID. The simplest posture is to elect to only send
+data on one path at a time, while accepting packets on all acceptable
+paths. In that case:
+
+* the attribution of packets to path discussed in {{zero-length-cid-loss-and-congestion}}
+  are easy to solve because packets are sent on a single path,
+* the ACK Delays are correct,
+* the vast majority of ECN marks relate to the current sending path.
```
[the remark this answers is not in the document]
huitema: There is always the occasional probe, etc.

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +612,36 @@ be controlled by the combination of one or several of the following:
...
+estimate the transmission delays on each path. (See
+{{ack-delay-and-zero-length-cid-considerations}} for specific considerations
+about using the ACK Delay field of ACK frames, and
+{{ecn-and-zero-length-cid-considerations}} for issues on using ECN marks.)
+
+Loss detection as specified in {{QUIC-RECOVERY}} uses algorithms
+based on timers and on sequence numbers. When packets are sent over
+multiple paths, loss detection must be adapted to allow for different RTTs
+on different paths. When sending to zero-length CID receivers, packets sent
+on different paths may not be received in order. Therefore senders cannot
```
mirjak, line 635:
```suggestion
on different paths may be received out of order. Therefore senders cannot
```

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,28 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+Using multiple packet number spaces enables direct use of the
+loss recovery and congestion control mechanisms defined in
+{{QUIC-RECOVERY}}.
+
+Some deployments of QUIC use zero-length connection IDs.
+When a node selects to use zero-length connection IDs, it is not
+possible to use different connection IDs for distinguishing packets
+sent to that node over different paths. All packets sent with a zero-length
+CID are numbered in the same number space. When using the same packet number space
+on multiple paths, out of order delivery is likely. This causes inflation of the number of
```
mirjak, line 139:
```suggestion
on multiple paths, out of order delivery is likely. This causes inflation of the number of
```

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,28 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+When a node selects to use zero-length connection IDs, it is not
+possible to use different connection IDs for distinguishing packets
+sent to that node over different paths. All packets sent with a zero-length
+CID are numbered in the same number space. When using the same packet number space
+on multiple paths, out of order delivery is likely. This causes inflation of the number of 
+acknowledgement ranges and therefore of the
+the size of ACK frames. Senders that accept to use a single number
+space on multiple paths when sending to a node using zero-length CID need
+to take special care to minimize the impact of multipath
+delivery on loss detection, congestion control, and ECN handling.
```
```suggestion
delivery on loss detection, congestion control, and ECN handling.
```

draft-ietf-quic-multipath.md
```diff
@@ -616,7 +612,36 @@ be controlled by the combination of one or several of the following:
...
+However, even if one packet number space is used on multiple path
+the sender MUST maintain separate congestion control state for each
+path. Therefore, senders MUST be able to infer the
+sending path from the acknowledged packet numbers, for example by remembering
+which packet was sent on what path. The senders MUST use that information to
+perform congestion control on the relevant paths, and to correctly
+estimate the transmission delays on each path. (See
+{{ack-delay-and-zero-length-cid-considerations}} for specific considerations
+about using the ACK Delay field of ACK frames, and
+{{ecn-and-zero-length-cid-considerations}} for issues on using ECN marks.)
```
mirjak, line 629:
```suggestion
{{ecn-handling}} for issues on using ECN marks.)
```

draft-ietf-quic-multipath.md
huitema: Fixed.

draft-ietf-quic-multipath.md
```diff
@@ -124,24 +124,29 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+each destination connection ID is linked to a separate packet number space.
+Using multiple packet number spaces enables direct use of the
+loss recovery and congestion control mechanisms defined in
+{{QUIC-RECOVERY}}.
+
+Some deployments of QUIC use zero-length connection IDs.
+When a node selects to use zero-length connection IDs, it is not
+possible to use different connection IDs for distinguishing packets
+sent to that node over different paths. This extension also specifies a way to use
+zero-length CID by using the same packet number space on all paths.
```
mirjak, line 138:
```suggestion
zero-length CID by using the same packet number space on all paths.
```

yfmascgy (approved):

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +625,41 @@ delay will most of the time relate to the path with the shortest latency.
...
+MAY disable ECN marking and
+send all subsequent packets as Not-ECN capable.
+
+### Restricted Sending to Zero-Length CID Peer
+
+Hosts that are designed to support multipath using multiple number spaces
+MAY adopt a conservative posture after negotiating multipath support with
+a peer using zero-length CID. The simplest posture is to elect to only send
+data on one path at a time, while accepting packets on all acceptable
+paths. In that case:
```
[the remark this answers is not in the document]
qdeconinck: That is what I understand from the current text, yes.

draft-ietf-quic-multipath.md
```diff
@@ -625,6 +628,41 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
```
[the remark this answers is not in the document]
qdeconinck: Maybe should we create an issue referencing this and create a specific PR afterwards?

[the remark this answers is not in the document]
mirjak: Created issue #125

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
mirjak: Yes, that's the meaning. For me the text is clear. I will merge the PR now, however, if you think further clarification is needed please propose a separate PR.

PR #117, opened 2022-04-06
[+] add stand-by status for path management

Yanmei-Liu:


draft-ietf-quic-multipath.md
```diff
@@ -1024,6 +1042,56 @@ field. If the Identifier Type if 0x02, the PATH_ABANDON frame MUST only
...
+
+PATH_STATUS Frames contain the following fields:
+
+Path Identifier: An identifier of the path, which is formatted 
+  as shown in {{fig-path-identifier-format}}. Exactly the same as 
+  the definition of Path Identifier in {#path-abandon-frame}.
+
+Path Status sequence number: A variable-length integer specifying 
+  the sequence number assigned for this PATH_STATUS frame.
+  There is a different path status sequence number space for each path.
```
mirjak, line 1070: Actually I don't think we need to maintain the sequence number per path on the sender side. As long as the sequence number is larger than the last received for that path, you should apply it.

Yanmei-Liu: It's ok that we maintain only one monotonically increasing sequence number for all the Path Status frame under the same connection.

Yanmei-Liu: Updated the definition of sequence number.

draft-ietf-quic-multipath.md
```diff
@@ -1091,6 +1109,57 @@ field. If the Identifier Type if 0x02, the PATH_ABANDON frame MUST only
 be sent on the path that is intended to be abandoned.
 
 
+## PATH_STATUS frame {#path-status-frame}
+
+PATH_STATUS Frame are used by endpoints to inform the peer of the current
```
qdeconinck, line 1114:
```suggestion
PATH_STATUS frames are used by endpoints to inform the peer of the current
```

draft-ietf-quic-multipath.md
```diff
@@ -307,6 +307,24 @@ to migrate to that path.  Instead, servers SHOULD consider new paths
 over which non-probing packets have been received as available
 for transmission.
 
+
+## Path State Management
+
+An endpoint uses PATH_STATUS frames to inform that the peer should
+send packets in the preference expressed by these frames.
```
qdeconinck, line 314: I think I would be useful to explicitly state that these are indications, i.e., the peer might not follow the endhost advertisements.

draft-ietf-quic-multipath.md
```diff
@@ -1091,6 +1109,57 @@ field. If the Identifier Type if 0x02, the PATH_ABANDON frame MUST only
...
+    Path Status (i),
+  }
+~~~
+{: #fig-path-status-format title="PATH_STATUS Frame Format"}
+
+PATH_STATUS Frames contain the following fields:
+
+Path Identifier: An identifier of the path, which is formatted
+  as shown in {{fig-path-identifier-format}}. Exactly the same as
+  the definition of Path Identifier in {#path-abandon-frame}.
```
qdeconinck, line 1133:
```suggestion
  the definition of Path Identifier in {{path-abandon-frame}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -1091,6 +1109,57 @@ field. If the Identifier Type if 0x02, the PATH_ABANDON frame MUST only
...
+
+Path Status sequence number: A variable-length integer specifying
+  the sequence number assigned for this PATH_STATUS frame. The sequence
+  number MUST be monotonically increasing for the sender of
+  the Path Status frame in the same connection.
```
qdeconinck, line 1138: Maybe we should state that the scope of the sequence number is relative to the Path Identifier? Having a single space for the whole connection might cause issues if we have different paths with different latencies.

draft-ietf-quic-multipath.md
```diff
@@ -1091,6 +1109,57 @@ field. If the Identifier Type if 0x02, the PATH_ABANDON frame MUST only
...
+
+Available values of Path Status field are:
+
+- 1: Standby
+- 2: Available
+
+Endpoints use PATH_STATUS frame to inform the peer whether it prefer to
+use this path or not. If an endpoint receives a PATH_STATUS frame
+containing 1-Standby status, it SHOULD stop sending non-probing packets
+on the corresponding path, until it receive a new PATH_STATUS frame
```
qdeconinck, line 1148:
```suggestion
on the corresponding path, until it receives a new PATH_STATUS frame
```

draft-ietf-quic-multipath.md
qdeconinck: Actually later this is explained, but I think it could be useful to explicitly state that here.

draft-ietf-quic-multipath.md
```diff
@@ -1091,6 +1109,57 @@ field. If the Identifier Type if 0x02, the PATH_ABANDON frame MUST only
...
+containing 2-Available status with a higher sequence number referring to
+the same path.
+
+Frames may be received out of order. A peer MUST ignore an incoming
+PATH_STATUS frame if it previously received another PATH_STATUS frame
+for the same Path Identifier with a sequence number equal to or
+higher than the sequence number of the incoming frame.
+
+PATH_STATUS frames SHOULD be acknowledged. If a packet containing
```
qdeconinck, line 1157: Do we want to state that PATH_STATUS is ack-eliciting?

qdeconinck (changes requested):

draft-ietf-quic-multipath.md
Yanmei-Liu: Actually the receiver should maintain the largest received sequence number for each path. The receiver should accept the Path Status Frame if and only if its sequence number is larger than the largest received sequence number of the Path Identifier. The latencies of different paths doesn't matter, because the receiver can distinguish Status Frames for different paths with Path Identifier field, and compares the sequence numbers independently.

draft-ietf-quic-multipath.md
Yanmei-Liu: The acknowledgement of PATH_STATUS frame is important for the sender to check whether the receiver got the path status change or not.

draft-ietf-quic-multipath.md
qdeconinck: I fully agree with your description, I just suggested to make it clearer here.
```suggestion
  the sequence number assigned for this PATH_STATUS frame. The sequence
  number is relative to the Path Identifier and MUST be monotonically increasing
  for the sender of the Path Status frame in the same connection.
```

draft-ietf-quic-multipath.md
qdeconinck:
```suggestion
PATH_STATUS frames are ack-eliciting. If a packet containing
```

draft-ietf-quic-multipath.md
Yanmei-Liu: I updated the description to make it clearer, and it's different for the sender and receiver side.

draft-ietf-quic-multipath.md
Yanmei-Liu: Added note for that.

qdeconinck (commented):
I think this is nearly ready. I just wonder what should be the behaviour of the PATH_STATUS receiver if all the existing paths are marked as Standby. Should we add something regarding this?

Yanmei-Liu:
> I think this is nearly ready. I just wonder what should be the behaviour of the PATH_STATUS receiver if all the existing paths are marked as Standby. Should we add something regarding this?

I agree with this point. For the receiver side, endpoint MAY ignore this frame if it would cause all the paths unavailable.
I updated the description.

huitema (approved):
Looks good.

qdeconinck (approved):
Sounds good to me.

Issue #95, opened 2022-02-09
Do we need a transport parameter to negotiate max path idle timeout?


mirjak:
This was discussed in issue #51 and PR #94.

yfmascgy:
Regarding the max_idle_timeout, RFC9000 says "To avoid excessively small idle timeout periods, endpoints MUST increase the idle timeout period to be at least three times the current Probe Timeout (PTO)." As in multi-path QUIC, the PTO is per-path, when max_idle_timeout is used to close a connection, it makes sense to set max_idle_timeout = 3*max(PTO_1, PTO_2). But when it comes to close a path, I am not sure if the timeout for path1 should also depend on the PTO of path 2?

qdeconinck:
I think there are two idle timeouts: the "connection idle timeout" and the "path idle timeout".

For the "connection idle timeout", I agree that it should be at least 3*PTO_max. Triggering this timeout closes the connection.

For the "path idle timeout", this might be a way to have a stronger guarantee to stop using paths (closing them) after some inactivity than the text proposed in #94. Still, if we define such a mechanism, we need to know whether all paths should have the same path timeout value or not (in the latter case, a transport parameter might not be the best way).

obonaventure:
I would avoid having a different timer for a path than for a connection. That would make the protocol more complex without a lot of benefits. If a host wants to use a shorter timeout per path than the connection timeout, it can simply define it and use a PATH_ABANDON frame to indicate that the path is abandoned. 

However, we need to be clearer in 3.2.3 to indicate that a path can be considered valid if either packets are received on that path or packets sent on that path are acknowledged (possibly on another path)

mirjak:
Presented at IETF-113 but no time for discussion.

yfmascgy:
I have submitted PR#124 to resolve this issue according to the above discussion.

mirjak:
PR was merged but forgot to close issue...

PR #124, opened 2022-07-07
fix max_idle_timeout issue in issue #95

yfmascgy:
Replace "max_idle_timeout" with "a path's idle_period" in the Idle_Timeout Section and make sure that the idle_period is larger than 3 times the path's PTO.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -397,18 +397,20 @@ as "no packet received on any path for the duration of the idle timeout".
 When only one path is available, servers MUST follow the specifications
 in {{QUIC-TRANSPORT}}.
 
-When more than one path is available, hosts shall monitor the arrival
-of non-probing packets and the acknowledgements
-for the packets sent over each path. Hosts SHOULD stop
-sending traffic on a path if for at least max_idle_timeout milliseconds
-(a) no non-probing packet was received or (b) no non-probing
-packet sent over this path was acknowledged, but MAY ignore that
-rule if it would disqualify
-all available paths. To avoid idle timeout of a path, endpoints can
-send ack-eliciting packets such as packets containing PING frames
-({{Section 19.2 of QUIC-TRANSPORT}}) on that path to keep it alive.
-Sending periodic PING frames also helps prevent middlebox timeout,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -397,18 +397,20 @@ as "no packet received on any path for the duration of the idle timeout".
...
+When more than one path is available, hosts shall monitor the arrival of
+non-probing packets and the acknowledgements for the packets sent over each
+path. Hosts SHOULD stop sending traffic on a path if for at least that path's
+idle_period milliseconds (a) no non-probing packet was received or (b) no
+non-probing packet sent over this path was acknowledged, but MAY ignore that
+rule if it would disqualify all available paths. If max_idle_timeout is
+specified, a path's idle_period is set to max_idle_timout if it is larger than
+three times the path's current PTO value.  To avoid excessively small idle
+timeout period, endpoints MUST increase a path's idle_peroid to be at least
+three times that path's current PTO. To avoid idle timeout of a path, endpoints
```
mirjak, line 409:
```suggestion
path. Hosts SHOULD stop sending traffic on a path if for at least the period of the
idle timeout as specified in {{Section 10.1. of QUIC-TRANSPORT}} 
(a) no non-probing packet was received or (b) no
non-probing packet sent over this path was acknowledged, but MAY ignore that
rule if it would disqualify all available paths. To avoid idle timeout of a path, endpoints
```

mirjak: Proposing to rather reference to RFC9000 instead of specifying this here separate again.

mirjak (changes requested):
Proposing to rather reference to RFC9000 instead of specifying this here separate again.

Yanmei-Liu (approved):
Just modified "...at least _max_idle_timeout milliseconds_ (a)..." 
into "... at least _the period of the idle timeout_  as specified in {{Section 10.1 of QUIC-TRANSPORT}} (a)..."

It's better to just modify the corresponding 2 lines, but it's ok for me.

mirjak (approved):

qdeconinck (approved):

PR #127, opened 2022-07-07
Add discussion about retiring DCID when there are still in-flight packets

### LINKED ISSUE #122: Handling incoming ACK_MP frames with a retired DCID sequence number
It might be possible to be in a state where a sender retires a DCID that was used by still in-flight packets. Hence, it is possible for the sender to receive a ACK_MP frame referencing the retired DCID. Depending on the situation:

- The sender can keep a way to track the sent packet with the retired DCID sequence number, allowing proper acknowledgement of the packets.
- The sender drops all the state related to the DCID, hence marking the in-flight packets as being lost and ignoring ACK_MP frames having a retired DCID sequence number.

While this should not affect interoperability, it might be worthwhile to add a sentence or two to mention the possible performance issue in such situation.

qdeconinck:
Fixes #122.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -387,6 +387,15 @@ acknowledgements. The peer MAY decide to keep sending data using
 the same IP addresses and UDP ports previously associated with
 the connection ID, but MUST use a different connection ID when doing so.
 
+Note that if the sender retires a Connection ID that is still used by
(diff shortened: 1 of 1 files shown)


Yanmei-Liu (commented):
Just little gramma fix

Yanmei-Liu (approved):

PR #129, opened 2022-07-07
Add some guidance for ACK scheduling

### LINKED ISSUE #123: Guidance on scheduling of packets containing ACK frames
This issue came up during the discussion of issue #33. We say that ACK or ACK_MP frames can be send over any path, however, maybe we should discuss some guidance and trade-offs. E.g. sending on the same path as the packets received also for RTT measurements; using the shortest delay path might have some performance benefits...?

qdeconinck:
Fixes #123.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -573,13 +573,15 @@ proposed and implemented, notably for Multipath TCP. A companion draft
 {{I-D.bonaventure-iccrg-schedulers}} provides several general-purpose
 packet schedulers depending on the application goals.
 
-The receiver plays an important role by scheduling ACK(_MP) frames. A
-natural ACK(_MP) scheduling strategy would be to send ACK(_MP) frames on
-the path showing the lowest latency to provide the fastest feedback to
-the peer. However, the sender can be interested by estimating the RTT of
-a path instead of the sum of the forward one-way delay of a path and the
-return one-way delay of another. To that end, a receiver should
-periodically send ACK(_MP) frames on the path they acknowledge packets.
+Note that the receiver could use a different scheduling strategy to send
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -573,6 +573,14 @@ proposed and implemented, notably for Multipath TCP. A companion draft
 {{I-D.bonaventure-iccrg-schedulers}} provides several general-purpose
 packet schedulers depending on the application goals.
 
+The receiver plays an important role by scheduling ACK(_MP) frames. A
+natural ACK(_MP) scheduling strategy would be to send ACK(_MP) frames on
+the path showing the lowest latency to provide the fastest feedback to
+the peer. However, the sender can be interested by estimating the RTT of
+a path instead of the sum of the forward one-way delay of a path and the
+return one-way delay of another. To that end, a receiver should
+periodically send ACK(_MP) frames on the path they acknowledge packets.
```
Yanmei-Liu, line 582: Sending ACK_(MP) frames on the fastest path is one of the ACK scheduling strategies, but IMO the most common strategy is sending ACK_(MP) frames on the same path of the packet was sent. By the way, as QUIC-TRANSPORT doesn't contain a mechanism to calculate one-way-delay on a certain path, it's very important that we need to add some guidance for one-way-delay calculation(e.g. [QUIC-Timestamp]) with the fastest path ACK scheduling strategy, and some guidance for the potential impact on congestion control.

Yanmei-Liu (changes requested):
Need to add guidance for one-way-delay calculation and impact on congestion control, if we want to recommend the fastest path ACK scheduling strategy.

Yanmei-Liu (approved):

mirjak:
I believe we can merge this now as it is. If we want to add more guidance later we should open a separate issue and another PR.

Issue #125, opened 2022-07-07
Add more explanation on ACK delay and zero-length CID


mirjak:
Comment by @yfmascgy (on PR #103):
I think we expand this section a bit as we found getting the delay right was non-trivial. QUIC time-stamp would work for sure. But if one decides not to use QUIC timestamp, we probably want to explain what is needed to be done here.

yfmascgy:
I have submitted PR #131 for this one.

qdeconinck:
Closing as #125 got merged.

PR #131, opened 2022-07-08
Add more explanation on ACK delay, RTT computation for single PNS

yfmascgy:
For issue #125 , add some considerations for RTT computation in our single pns implementation. Not sure if it is easy enough to understand. Please feel free to change.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -724,7 +724,7 @@ can be used:
   time of that packet. When an ACK is sent on that path, the ACK Delay field is
   re-calculated using that path's largest packet receive time.
 
-* Each path also keeps track of a list of sent packets that are acknowledged by
+* The sender also keeps track of a list of sent packets for each path that are acknowledged by
   ACKs from the same path. A path's RTT sample is generated on receving ACK
   that meets the following two conditions: (1) For that path, the largest
   same-path acknowledged packet number is updated. (2) One of newly same-path
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -715,7 +715,7 @@ If one chooses not to use time-stamps but wants to get reasonable estimation of
 RTTs on multiple paths with one packet number space, the following practices
 can be used:
 
-* Each path counts the number of ACK-eliciting packets received on that path,
+* The receiver side counts the number of ACK-eliciting packets received for each paths,
   and keeps a per-path ACK timer. An ACK from that path is triggered when the
   number of ACK-eliciting packets received on that path surpasses the path's
   ACK-eliciting threshold or the path's ACK-timer expires.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -714,17 +714,17 @@ packet was sent. However, when applying the above algorithm with SPNS, one may
 encounter the following issues: (1) RTT of some paths are not updated timely if
 ACKs are mostly returned from other paths, and (2) ACK frames depend on the
 largest received packet of the connection, not the path, and the resulted RTT
-sample may be the sum of the uplink delay of one path and the downlink delay of
-another path. One solution for accurate RTT measurements is to employ
-time-stamps as described in {{QUIC-Timestamp}}. If one chooses not to use
-time-stamps but wants to get reasonable estimation of RTTs on multiple paths
-with single packet number space, the following practices can be used:
+sample may be the sum of the one-way delays of two different paths. One
+solution for accurate RTT measurements is to employ time-stamps as described in
+{{QUIC-Timestamp}}. If one chooses not to use time-stamps but wants to get
+reasonable estimation of RTTs on multiple paths with single packet number
+space, the following practices can be used:
 
 For packet receiver (ACK sender):
 
 * Maintain an ACK threshold and an ACK timer for each path. A path should send
-an ACK when it receives ack-eliciting threshold number of ack-eliciting
-packets (e.g., two). And an ack-eliciting packet must be acknowledged within
+an ACK when it receives ack-eliciting-threshold number of ack-eliciting
+packets (e.g., two) on this path, and an ack-eliciting packet must be acknowledged within
 MAX_ACK_DELAY.
 
 * Write ACK frame based on the largest received packet of the path. Start the ACK
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -702,34 +702,45 @@ maintaining an order list of packets sent on each path. That ordered
 list can then be used to compute acknowledgement gaps per path in
 Packet Threshold tests.
 
-### ACK Delay, RTT, and Zero-Length CID Considerations {#ack-delay-and-zero-length-cid-considerations}
+### RTT Estimation Considerations when SPNS is Used {#ack-delay-and-zero-length-cid-considerations}
 
-The ACK Delay field of the ACK frame is relative to the largest
-acknowledged packet number (see {{Section 13.2.5 of QUIC-TRANSPORT}}).
-When using paths with different transmission delays, the reported host
-delay will most of the time relate to the path with the shortest latency.
-To collect ACK delays on all the paths, hosts should rely on time stamps
-as described in {{QUIC-Timestamp}}.
+When SPNS is in use, accurate RTT estimation requires more careful
+considerations. According to {{QUIC-RECOVERY}}, an endpoint generates an RTT
+sample on receiving an ACK frame that meets the following two conditions: (1)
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -711,6 +711,26 @@ delay will most of the time relate to the path with the shortest latency.
 To collect ACK delays on all the paths, hosts should rely on time stamps
 as described in {{QUIC-Timestamp}}.
 
+If one chooses not to use time-stamps but wants to get reasonable estimation of
+RTTs on multiple paths with one packet number space, the following practices
+can be used:
+
+* Each path counts the number of ACK-eliciting packets received on that path,
```
Yanmei-Liu, line 718:
```suggestion
* The receiver side counts the number of ACK-eliciting packets received for each paths,
```

Yanmei-Liu (changes requested):
The mechanism is ok, but we need to make it a little clearer by distinguishing actions for the sender and receiver side.

Yanmei-Liu (approved):
Look good to me.

draft-ietf-quic-multipath.md
```diff
@@ -711,6 +711,26 @@ delay will most of the time relate to the path with the shortest latency.
...
+* Each path records the largest packet received on that path and the receive
+  time of that packet. When an ACK is sent on that path, the ACK Delay field is
+  re-calculated using that path's largest packet receive time.
+
+* The sender also keeps track of a list of sent packets for each path that are acknowledged by
+  ACKs received from the same path. A path's RTT sample is generated on receving ACK
+  that meets the following two conditions: (1) For that path, the largest
+  same-path acknowledged packet number is updated. (2) One of newly same-path
+  acknowledged packets is ACK-eliciting.
+
```
huitema, line 732: I am trying to understand your algorithm. I guess it sums up as:

* For each path, the sender maintains knowledge of the largest PN acknowledged on path.
* The sender also keeps track of a list of sent packets for each path that are not yet acknowledged by
  ACKs received from the same path.
  
When a same path ACK is received:

* process the acknowledged ranges until finding a packet that:
      1) Was sent on the path
      2) Has a packet number larger than the largest PN acknowledged on path
* if several packets match the criteria above, retain only the one with the largest PN
* the remaining list may be empty, in which case do nothing
* if one packet matches the criteria above:
      1) Compute the RTT sample as the difference between the ACK received time and the packet sent time.
      2) Update RTT average, etc.
      3) Set the largest PN acknowledged on path to the PN that was just processed
      3) Remove all packets with PN lower or equal than this largest acknowledged PN from the list of sent packets for each path that are not yet acknowledged by ACKs received from the same path

The downside of that method is that the RTT sample may include some or all of the ACK Delay of the ACK frame. The possible rule is:

* if the largest acknowledged number in the ACK frame matches the largest acknowledged number on path, remove the ACK Delay from the RTT sample.
* if not, accept that the measurement may be imprecise.

Do I understand correctly?

huitema: Of course, we need to also specify the receiver behavior. Specifying that behavior may help deal with the "ACK Delay" issue. I assume that this why you write:

* The receiver counts the number of ACK-eliciting packets received for each paths,
  and keeps a per-path ACK timer. An ACK from that path is triggered when the
  number of ACK-eliciting packets received on that path surpasses the path's
  ACK-eliciting threshold or the path's ACK-timer expires.

I am not too sure about that, because the net effect of that rule is to get at least one new ACK **for each path** per ACK Timer. If the goal is to provide accurate measurement of RTT, we don't need that many "on path ACK". One per RTT would be enough. Maybe 2 per RTT to be safe. Or the max value of "ACK Timer" and "Path RTT / 2". Reset that timer for each path after sending an on path RTT. Then, trigger an on path ACK as soon as (1) the timer has expired and (2) a packet has been received. (Ack eliciting or not does not matter.)

That way, the sender can assume that for the on-path packet, the ACK delay is set to zero.

huitema: Correction. My concerned about ACK Delay is misplaced. We are speaking about ACK_(MP), so the largest packet acknowledged is always a packet sent on that same path. See the modification that I propose on PR #129, a simple guidance for always sending at least some ACK on the same path. If we accept that guidance, then the text becomes simpler -- no need to specify a receiver side strategy. So the entire addition becomes:


* For each path, the sender maintains knowledge of the largest PN acknowledged on path.

* The sender also keeps track of a list of sent packets for each path that are not yet acknowledged by ACKs received from the same path.

* When a same path ACK is received, the sender checks whether the largest packet acknowledged is larger than the largest PN acknowledged on path. If it is, the sender will:

    - Compute a new path RTT sample using the ACK arrival time, the ACK delay indication, and the time at which the packet with this largest PN was sent.
    - Set the largest PN acknowledged on path to the PN that was just processed
    - Remove all packets with PN lower or equal than this largest acknowledged PN from the list of packets sent on this path that are not yet acknowledged by ACKs received from the same path.

huitema: Of course, there is the issue of packets sent with zero-length CID. In that case, the simple algorithm above does not work very well. The ACK_(MP) will contain ack ranges received on many paths. Instead of using the largest number in the ACK_(MP) frame, the sender needs to filter the ACK ranges and retrieve the largest number acknowledged for the path.

draft-ietf-quic-multipath.md
```diff
@@ -711,6 +711,26 @@ delay will most of the time relate to the path with the shortest latency.
...
+can be used:
+
+* The receiver side counts the number of ACK-eliciting packets received for each paths,
+  and keeps a per-path ACK timer. An ACK from that path is triggered when the
+  number of ACK-eliciting packets received on that path surpasses the path's
+  ACK-eliciting threshold or the path's ACK-timer expires.
+
+* Each path records the largest packet received on that path and the receive
+  time of that packet. When an ACK is sent on that path, the ACK Delay field is
+  re-calculated using that path's largest packet receive time.
```
qdeconinck, line 725: It sounds odd to me to change the meaning of the ACK Delay field for a particular use case...

Yanmei-Liu (changes requested):
Need more discussion about the mechanism.

mirjak:
Yes, this needs more discussion. Also maybe we want to move this text rather in the "Implementation Considerations" section?

draft-ietf-quic-multipath.md
```diff
@@ -702,14 +702,45 @@ maintaining an order list of packets sent on each path. That ordered
...
+considerations. According to {{QUIC-RECOVERY}}, an endpoint generates an RTT
+sample on receiving an ACK frame that meets the following two conditions: (1)
+the largest acknowledged packet number is newly acknowledged, and (2) at least
+one of the newly acknowledged packets was ack-eliciting. The RTT sample,
+latest_rtt is calculated as the time elapsed since the largest acknowledged
+packet was sent. However, when applying the above algorithm with SPNS, one may
+encounter the following issues: (1) RTT of some paths are not updated timely if
+ACKs are mostly returned from other paths, and (2) ACK frames depend on the
+largest received packet of the connection, not the path, and the resulted RTT
+sample may be the sum of the uplink delay of one path and the downlink delay of
```
Yanmei-Liu, line 717: ... and the resulted RTT sample may be the sum of _the one way delay of two different paths._

draft-ietf-quic-multipath.md
```diff
@@ -702,14 +702,45 @@ maintaining an order list of packets sent on each path. That ordered
...
+another path. One solution for accurate RTT measurements is to employ
+time-stamps as described in {{QUIC-Timestamp}}. If one chooses not to use
+time-stamps but wants to get reasonable estimation of RTTs on multiple paths
+with single packet number space, the following practices can be used:
+
+For packet receiver (ACK sender):
+
+* Maintain an ACK threshold and an ACK timer for each path. A path should send
+an ACK when it receives ack-eliciting threshold number of ack-eliciting
+packets (e.g., two). And an ack-eliciting packet must be acknowledged within
```
Yanmei-Liu, line 727: A path should send an ACK _frame_ when it receives ack-eliciting threshold number of ack-eliciting
packets (e.g., two) _on the current path_.

Yanmei-Liu (approved):
I'm ok with the current version, just a few editorial hints.

mirjak:
@huitema, @yfmascgy updated this PR. Can we merge it?

Note that this PR is only for SPNS and part of that respective section. Based on his report, the strategy seems reasonable.

For ACK_MP we have the following text in the packet scheduling section (7.):

> Note that the receiver could use a different scheduling strategy to send ACK(_MP) frames. The recommended default behaviour consists in sending ACK(_MP) frames on the path they acknowledge packets.

huitema (approved):
Seems like a reasonable tradeoff, easy to implement.

PR #133, opened 2022-07-08
add some text to refuse path initiation

### LINKED ISSUE #120: active_connection_id_limit
Is there any reason why a new parameter isn't defined to negotiate max_paths rather than reusing this existing parameter? 

That design is much more cleaner. For example, the max paths wouldn't be ignored even when zero-length CIDs are in use. 

Unless I'm mistaken, the current design have to comply with the following from RFC9000:

>If an endpoint issues a zero-length connection ID... and therefore ignores the active_connection_id_limit value received from its peer.

BTW, how the following MUST will be honored for the zero-length CID case: 

```
Before an
   endpoint adds a new path by initiating path validation, it MUST check
   whether at least one unused Connection ID is available for each side.
```

qdeconinck:
Should fix #120.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -389,11 +389,13 @@ CONNECTION_CLOSE frame.
 ### Refusing a New Path
 
 An endpoint may deny the establishment of a new path initiated by its
-peer during the address validation procedure. To achieve this, the
-endpoint does not send a PATH_RESPONSE in response to the peer's
-PATH_CHALLENGE. The endpoint instead sends a PATH_ABANDON frame
-containing the Path Identifier of the refused path on another active path.
-
+peer during the address validation procedure. According to
+{{QUIC-TRANSPORT}}, the standard way to deny the establishment of a path
+is to not send a PATH_RESPONSE in response to the peer's PATH_CHALLENGE.
+An endpoint that has negotiated the usage of the multipath extension MAY
+use an explicit method by sending on another active path a PATH_ABANDON
+frame containing the Path Identifier of the refused path, but only if the
+PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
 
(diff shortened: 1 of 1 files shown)


huitema (commented):
Yes, good idea. Specifying "on another path" is indeed the key, because otherwise the responder needs to consume a CID. 

There is a corner case when the Path Challenge arrives with a NULL CID. What happens then?

Yanmei-Liu (commented):
I'm ok with the new text, but I agree with Christian that we need to explain what would happen when endpoints use NULL CID.

draft-ietf-quic-multipath.md
```diff
@@ -386,6 +386,15 @@ the server MAY wait for a short, limited time such as one RTO if a path
...
 CONNECTION_CLOSE frame.
 
+### Refusing a New Path
+
+An endpoint may deny the establishment of a new path initiated by its
+peer during the address validation procedure. To achieve this, the
+endpoint does not send a PATH_RESPONSE in response to the peer's
+PATH_CHALLENGE. The endpoint instead sends a PATH_ABANDON frame
+containing the Path Identifier of the refused path on another active path.
+
```
huitema, line 396: The standard way to deny the establishment of a path, according to {{RFC9000}}, is
to not send a PATH_RESPONSE  in response to the peer's
PATH_CHALLENGE. An endpoint that has negotiated use of multipath can
use a more direct method, but only if the PATH_CHALLENGE arrives with
a non-zero length CID. This endpoint MAY send a PATH_ABANDON frame
containing the Path Identifier of the refused path on another active path.

qdeconinck:
I rewrote the text based on the suggestion from @huitema to discuss the case of zero-length CIDs.

Yanmei-Liu (approved):
Looks good to me.

mirjak (approved):

PR #134, opened 2022-07-08
Do not imply

huitema:
Minor tweak. The text says that a path abandon "implies" retirement of the CID. Readers could be mislead and assume that they don't need to send a retire frame after sending abandon, when in fact they do.


huitema:
@Yanmei-Liu , @qdeconinck, @mirjak please review.

Yanmei-Liu (approved):
LGTM

qdeconinck (approved):
The proposed text is indeed clearer.

mirjak (approved):

PR #148, opened 2022-10-21
New implementation consideration on keep alives

### LINKED ISSUE #144: How to specify over which path to send PING frames?
Currently we don't mention PING frames in the draft, however, I guess there are use cases where an app wants to send a ping over one specific or both paths. We should probably say something about handling of PING frames if multiple path are available.

mirjak:
fixes #144

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1059,6 +1059,16 @@ each path. The benefit of such an approach is to simplify retransmission
 processing as the content of lost packets initially sent on one path can be sent
 on another path without further frame scheduling adaptations.
 
+## Keep Alive
+
+The QUIC specification defines an optional keep alive process, see {{Section 5.3 of RFC9000}}.
+Implementations of the multipath extension should map this keep alive process to a number of paths.
+Some applications may wish to ensure that one path remains active, while others could prefer to have
+two or more active paths during the connection lifetime. Different applications will likely require different strategies.
+Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
+on each of these paths before the idle timeout expires. Implementations should jitter their timers
+to avoid sending Ping frames at the same time over different paths.
+
 # New Frames {#frames}
 
 All the new frames MUST only be sent in 1-RTT packet, and MUST NOT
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1059,6 +1059,16 @@ each path. The benefit of such an approach is to simplify retransmission
 processing as the content of lost packets initially sent on one path can be sent
 on another path without further frame scheduling adaptations.
 
+## Keep Alive
+
+The QUIC specification defines an optional keep alive process, see {{Section 5.3 of RFC9000}}.
```
mirjak, line 1064:
```suggestion
The QUIC specification defines an optional keep alive process, see {{Section 5.3 of QUIC-TRANSPORT}}.
```

qdeconinck (approved):
I'm fine with the proposed addition.

huitema (commented):
Looks good. A bit redundant with other parts, but does not hurt.

draft-ietf-quic-multipath.md
```diff
@@ -1059,6 +1059,16 @@ each path. The benefit of such an approach is to simplify retransmission
...
 
+## Keep Alive
+
+The QUIC specification defines an optional keep alive process, see {{Section 5.3 of QUIC-TRANSPORT}}.
+Implementations of the multipath extension should map this keep alive process to a number of paths.
+Some applications may wish to ensure that one path remains active, while others could prefer to have
+two or more active paths during the connection lifetime. Different applications will likely require different strategies.
+Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
+on each of these paths before the idle timeout expires. Implementations should jitter their timers
+to avoid sending Ping frames at the same time over different paths.
```
mirjak, line 1070: > Implementations should jitter their timers to avoid sending Ping frames at the same time over different paths.

Do we really need this last sentence or can we remove it?

Yanmei-Liu: I'm ok with the previous parts.  We could remove the last sentence as it depends on the implementation.

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
mirjak:
```suggestion
on each of these paths before the idle timeout expires.
```

Issue #137, opened 2022-09-01
How should a receiver process a PATH_ABANDON having an unknown Path Identifier?


qdeconinck:
When using Identifier Types 0 or 1, the current text does not discuss how we should handle related path Identifier that the receiver cannot map to a path.

When using Identifier Type 1, the Path Identifier corresponds to the sequence number of the Destination CID used by the sender to send packets on the related path, hence a sequence number of a source Connection ID of the receiver. When using Identifier Type 0, the Path Identifier corresponds to the sequence number of the Source CID of the sender, hence the sequence number of the Destination CID used by the receiver to send packets on the related path.

While there are obvious values that could be considered as bad and may lead to, e.g., a MP_PROTOCOL_VIOLATION (e.g., a sequence number larger than any sequence number advertised), I don't think we should be as radical with regular values as they might have been retired meanwhile (also possible to face this with packet reordering). The worst edge cases I see are the followings.

With type 1, the sender might send a PATH_ABANDON with the seq of the CID in one packet, and then the RETIRE_CID retiring that CID in a subsequent one, but the RETIRE_CID reaches the receiver before the PATH_ABANDON one. If the receiver drops state upon RETIRE_CID reception, it won't be able to map the PATH_ABANDON to the path.

With type 0, the sender might send a PATH_ABANDON with the seq of the CID, but before it reaches the peer, the peer sends a RETIRE_CID retiring that CID. If the receiver drops the CID state upon RETIRE_CID sending, it won't be able to map the PATH_ABANDON to the path.

Handling these edge cases with connection closure is clearly too strong. Still, ignoring or even acknowledging these PATH_ABANDON may have bad consequences. Silently dropping packets (i.e., not acknowledge the packet having PATH_ABANDON) will make the peer retransmit the frame (hoping it updates its path identifier) but could cause the loss/non-acknowledgment of accompanying frames (e.g., STREAM ones), and processing packet content without acknowledging it sounds odd. Acknowledging PATH_ABANDON that cannot be mapped to a path would make some desynchronisation about the status of the path between the peers (the sender now assumes the path is closed, but the receiver not).

To address the type 1 issue, we could write some text stating that a sender cannot retire a DCID if a PATH_ABANDON containing the related sequence number is still in flight. But for the type 0, the fix is not obvious (I don't think we could have the same solution as in #135)...

yfmascgy:
[...]

PR #149, opened 2022-10-21
First proposal to "remove" single packet number space option

mirjak:
This PR removes the use of a single packet number space on multiple paths as a fully supported option. However, it currently doesn't remove it entirely.
Specifically, it moves the respective text for now into the appendix but still allows the reception of packets without a Connection ID on multiple validated paths, similarly as path migration is not strictly forbidden without connection ID.


draft-ietf-quic-multipath.md
```diff
@@ -1324,3 +1193,141 @@ one of the original proposals are:
 # Acknowledgments
 
 TBD
+
+--- back
+
+# Sending packet on multiple paths with zero length connection ID
```
mirjak, line 1199:
```suggestion
# Sending packets on multiple paths with zero length connection ID
```

draft-ietf-quic-multipath.md
```diff
@@ -1324,3 +1193,141 @@ one of the original proposals are:
...
+When using the same packet number space
+on multiple paths, out of order delivery is likely. This causes inflation of the number of
+acknowledgement ranges and therefore of the
+the size of ACK frames. Senders that accept to use a single number
+space on multiple paths when sending to a node using zero-length CID need
+to take special care to minimize the impact of multipath
+delivery on loss detection, congestion control, and ECN handling.
+This proposal specifies algorithms for
+controlling the size of acknowledgement packets and ECN handling in
+Section {{using-zero-length}} and {{ecn-handling}}.
```
mirjak, line 1216:
```suggestion
{{ecn-handling}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -663,162 +654,39 @@ the Destination Connection ID present in the packet header, as defined
...
-### Restricted Sending to Zero-Length CID Peer
-
-Hosts that are designed to support multipath using multiple number spaces
-MAY adopt a conservative posture after negotiating multipath support with
-a peer using zero-length CID. The simplest posture is to only send
-data on one path at a time, while accepting packets on all acceptable
-paths. In that case:
+Further, if the initial path uses zero-length Connection ID and
+packets without a Connection ID are successfully received over another validated path,
+these packet MUST either be acknowledged using ACK_MP frames with packet number space ID 0,
```
mirjak, line 682:
```suggestion
these packets MUST either be acknowledged using ACK_MP frames with packet number space ID 0,
```

draft-ietf-quic-multipath.md
```diff
@@ -663,162 +654,39 @@ the Destination Connection ID present in the packet header, as defined
...
-  are easy to solve because packets are sent on a single path,
-* the ACK Delays are correct,
-* the vast majority of ECN marks relate to the current sending path.
-
-Of course, the hosts will only take limited advantage from the multipath
-capability in these scenarios. Support for "make before break" migrations
-will improve, but load sharing between multiple paths will not work.
-
-## Using non-zero length CID and Multiple Packet Number Spaces
+However, even if packet without a Connection ID are successfully received
```
mirjak, line 685:
```suggestion
However, even if packets without a Connection ID are successfully received
```

draft-ietf-quic-multipath.md
```diff
@@ -663,162 +654,39 @@ the Destination Connection ID present in the packet header, as defined
...
-
-Of course, the hosts will only take limited advantage from the multipath
-capability in these scenarios. Support for "make before break" migrations
-will improve, but load sharing between multiple paths will not work.
-
-## Using non-zero length CID and Multiple Packet Number Spaces
+However, even if packets without a Connection ID are successfully received
+on multiple paths, it is RECOMMENDED to only use one of the active paths
+for sending of ack-elicting packets with zero-length Connection ID as this
+avoids large ACK frames and ambiguity if loss detection and RTT estimation
```
mirjak, line 688:
```suggestion
avoids large ACK frames and ambiguity in loss detection and RTT estimation
```

draft-ietf-quic-multipath.md
```diff
@@ -663,162 +654,39 @@ the Destination Connection ID present in the packet header, as defined
...
+To open a new path, an endpoint SHALL use different Connection IDs on different paths.
 Still, the receiver may observe the same Connection ID used on different
 4-tuples due to, e.g., NAT rebinding. In such case, the receiver reacts
 as specified in {{Section 9.3 of QUIC-TRANSPORT}}.
 
+If the initial path carries no Connection ID and a QUIC packet without
+a Connection ID is received on a 5-tuple of a non validated path,
+the receiver may try to decrypt the packet assuming the packet number
+space of the initial path with Path ID 0. This usually is only plausible
+if only the source address or port are different or if a specific
```
mirjak, line 666:
```suggestion
if only the source address or port are different or if a specific
```

draft-ietf-quic-multipath.md
```diff
@@ -873,7 +741,8 @@ In order to guarantee the uniqueness of the nonce, the nonce N is
 calculated by combining the packet protection IV with the packet number
 and with the path identifier.
 
-The path ID for 1-RTT packets is the sequence number of {{QUIC-TRANSPORT}}, or
+The path ID for 1-RTT packets is the sequence number of the Connection ID
```
mirjak, line 744:
```suggestion
The path ID for 1-RTT packets is the sequence number of the Connection ID
```

draft-ietf-quic-multipath.md
```diff
@@ -1324,3 +1193,141 @@ one of the original proposals are:
...
+
+*  Limiting the number of transmissions of a specific ACK range, on
+   the assumption that a sufficient number of transmissions almost
+   certainly ensures reception by the peer.
+
+*  Send multiple messages for a given path in a single socket
+   operation, so that a series of packets sent from a single path
+   uses a series of consecutive sequence numbers without creating
+   holes.
+
```
mirjak, line 1270:
```suggestion

```

huitema (commented):
I don't think this is ready yet. We may need to agree on the broad lines first.

mirjak:
@huitema yes I also opened issue #152 for discussion!

draft-ietf-quic-multipath.md
```diff
@@ -846,15 +696,15 @@ the server provided N Connection IDs, and the client is already actively
 using N paths, the limit is reached. If the client wants to start
 a new path, it has to retire one of the established paths.
 
-ACK_MP frame (defined in {{ack-mp-frame}}) can be returned via either a
-different path, or the same path identified by the Path Identifier, based on
-different strategies of sending ACK_MP frames.
+ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be send on the same path 
+as identified by the Path Identifier. However, an endpoint can be returned via either a
```
mirjak, line 700:
```suggestion
as identified by the Path Identifier. However, an ACK_MPfFrame can be returned via either a
```

mirjak:
I updated the PR and removed the appendix for SPNP. Further this PR threats the use of the multipath extension without CID as an error new text is the following:

```
If the parameter is set to 1, both endpoints MUST use non-zero connection
IDs. If an enable_multipath parameter set to 1 is received and the carrying packet
does not contain a non-zero legnth connection, the receiver MUST treat this as a connection error of type
TRANSPORT_PARAMETER_ERROR (specified in {{Section 20.1 of QUIC-TRANSPORT}})
and close the connection.
```

draft-ietf-quic-multipath.md
```diff
@@ -846,15 +696,15 @@ the server provided N Connection IDs, and the client is already actively
 using N paths, the limit is reached. If the client wants to start
 a new path, it has to retire one of the established paths.
 
-ACK_MP frame (defined in {{ack-mp-frame}}) can be returned via either a
-different path, or the same path identified by the Path Identifier, based on
-different strategies of sending ACK_MP frames.
+ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be send on the same path
```
mirjak, line 699:
```suggestion
ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be send on the same path
```

draft-ietf-quic-multipath.md
```diff
@@ -846,15 +696,15 @@ the server provided N Connection IDs, and the client is already actively
...
+as identified by the Path Identifier. However, an ACK_MP frame can be returned via either a
+different path, based on different strategies of sending ACK_MP frames.
```
mirjak, line 701:
```suggestion
as identified by the Path Identifier. However, an ACK_MP frame can be returned via a
different path, based on different strategies of sending ACK_MP frames.
```

draft-ietf-quic-multipath.md
```diff
@@ -176,12 +172,11 @@ We assume that the reader is familiar with the terminology used in
 - Path Identifier (Path ID): An identifier that is used to identify
   a path in a QUIC connection at an endpoint. Path Identifier is used
   in multipath control frames (etc. PATH_ABANDON frame) to identify
-  a path. By default, it is defined as the sequence number of
+  a path. The Path ID is defined as the sequence number of
   the destination Connection ID used for sending packets on that
-  particular path, but alternative definitions can be used if the length
-  of that connection ID is zero.
+  particular path.
```
mirjak, line 177:
```suggestion
  particular path.
```

draft-ietf-quic-multipath.md
```diff
@@ -250,6 +242,12 @@ the endpoints MUST fallback to
 {{QUIC-TRANSPORT}} with single active path and MUST NOT use any frame or
 mechanism defined in this document.
 
+If the parameter is set to 1, both endpoints MUST use non-zero connection
+IDs. If an enable_multipath parameter set to 1 is received and the carrying packet
+does not contain a non-zero legnth connection, the receiver MUST treat this as a connection error of type
```
mirjak, line 247:
```suggestion
does not contain a non-zero length connection ID, the receiver MUST treat this as a connection error of type
```

draft-ietf-quic-multipath.md
```diff
@@ -104,9 +104,14 @@ QUIC version 1)
...
 which usually also requires per-path RTT measurements
   * PMTU discovery should be performed per-path
+  * The use of this multipath extension requires the use of non-zero
+Connection IDs in both directions.
   * A path is determined by the 4-tuple of source and destination IP
 address as well as source and destination port. Therefore, there can be
 at most one active paths/connection ID per 4-tuple.
+  * If the 4-tuple changes without the use of a new connection ID (e.g.
+due to a NAT rebinding), this is considered as a migration events and not
+an new path.
```
qdeconinck, line 114:
```suggestion
due to a NAT rebinding), this is considered as a migration event and not
as a new path.
```

draft-ietf-quic-multipath.md
```diff
@@ -192,10 +187,7 @@ Identifier, is that the Path Identifier is used in multipath control
 frames to identify a path, and the Packet Number Space Identifier is
 used in 1-RTT packets and ACK_MP frames to distinguish packet number
 spaces for different paths. Both identifiers have the same value, which
-is the sequence number of the connection ID, if a non-zero connection ID
-is used. If the connection ID is zero length, the Packet Number Space
-Identifier is 0, while the Path Identifier is selected on path
-establishment.
+is the sequence number of the connection ID, when a new path is established.
```
qdeconinck, line 190: Maybe it can be addressed later, but given that the Path Identifier and the PNS ID are now always the same, we may simplify this section and only introduce one concept (i.e., the Path Identifier).

draft-ietf-quic-multipath.md
```diff
@@ -250,6 +242,12 @@ the endpoints MUST fallback to
 {{QUIC-TRANSPORT}} with single active path and MUST NOT use any frame or
 mechanism defined in this document.
```
qdeconinck, line 243: Given that we drop support for single packet number space and zero-length CID, we need to change the value of `enable_multipath`. Running a random generator gave me `0x0f739bbc1b666db9`, but we can of course choose another value (for instance, `0x0f739bbc1b666dXX` where `XX` is the draft version; the next version 04 of this document could use  `0x0f739bbc1b666d04`). The IANA table should be updated as well.

draft-ietf-quic-multipath.md
```diff
@@ -676,184 +671,8 @@ ACK frames are received in 1-RTT packets while the state of multipath
...
-capability in these scenarios. Support for "make before break" migrations
-will improve, but load sharing between multiple paths will not work.
-
-## Using non-zero length CID and Multiple Packet Number Spaces
-
-If packets contain a non-zero CID, each path has
-its own packet number space for transmitting 1-RTT packets and a new
-ACK frame format is used as specified in {{ack-mp-frame}}.
+A new ACK frame format is used as specified in {{ack-mp-frame}} that
+MUST be used for for acknolowing 1-RTT packets.
```
qdeconinck, line 675: The added text conflicts with the latter one saying "endpoints SHOULD use ACK_MP frames instead of ACK", and I don't think receiving ACK frames to acknowledge path 0 is actually an issue. Maybe this would be better?
```suggestion
The ACK_MP frame, as specified in {{ack-mp-frame}}, is used to
acknowledge 1-RTT packets.
```

draft-ietf-quic-multipath.md
```diff
@@ -863,9 +682,8 @@ Therefore, the packet number space for 1-RTT packets can be identified
 based on the Destination Connection ID in each packet.
 
 As soon as the negotiation of multipath support is completed,
-endpoints SHOULD use ACK_MP frames instead of ACK frames for
-acknowledgements of 1-RTT packets on path 0, as well as for 0-RTT packets
-that are acknowledged after the handshake concluded.
+endpoints SHOULD use ACK_MP frames instead of ACK frames also for 0-RTT packets
+that are acknowledged after the handshake concluded using the Packet Number Space Identifier 0.
```
qdeconinck, line 686: Maybe clearer?
```suggestion
endpoints SHOULD use ACK_MP frames instead of ACK frames to acknowledge application
data packets, including 0-RTT packets, received on path ID 0 after the handshake concluded.
```

draft-ietf-quic-multipath.md
```diff
@@ -877,15 +695,15 @@ the server provided N Connection IDs, and the client is already actively
 using N paths, the limit is reached. If the client wants to start
 a new path, it has to retire one of the established paths.
 
-ACK_MP frame (defined in {{ack-mp-frame}}) can be returned via either a
-different path, or the same path identified by the Path Identifier, based on
-different strategies of sending ACK_MP frames.
+ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be send on the same path
+as identified by the Path Identifier. However, an ACK_MP frame can be returned via a
```
qdeconinck, line 699: No strong opinion here, but should the "an ACK_MP frame can be returned" be instead "an ACK_MP frame MAY be returned"?

draft-ietf-quic-multipath.md
```diff
@@ -904,8 +722,8 @@ In order to guarantee the uniqueness of the nonce, the nonce N is
 calculated by combining the packet protection IV with the packet number
 and with the path identifier.
 
-The path ID for 1-RTT packets is the sequence number of {{QUIC-TRANSPORT}}, or
-zero if the Connection ID is zero-length.  {{Section 19 of QUIC-TRANSPORT}}
+The path ID for 1-RTT packets is the sequence number of the Connection ID
+as specfied in {{QUIC-TRANSPORT}}.  {{Section 19 of QUIC-TRANSPORT}}
```
qdeconinck, line 726:
```suggestion
as specified in {{QUIC-TRANSPORT}}.  {{Section 19 of QUIC-TRANSPORT}}
```

draft-ietf-quic-multipath.md
```diff
@@ -1139,14 +920,11 @@ path identifier.
     the control frame.
     Note that this is the connection identifier used by the peer
     when sending packets on the to-be-closed path.
-    This method SHOULD be used if this connection identifier is non-zero
-    length. This method MUST NOT be used if this connection identifier
-    is zero-length.
+    This method SHOULD be used by default.
```
qdeconinck, line 923: We can of course delegate this on a subsequent PR, but I think we should get rid of the `Identifier Type` for the PATH_ABANDON and PATH_STATUS frames. Also, I wonder it the value 1 should rather be the default (given #137).

qdeconinck (changes requested):
Globally good to go, a few nits to fix before merging. Also, given the interoperability changes it involves, I suggest to already change the transport parameter value in this PR (see my specific comment below).

draft-ietf-quic-multipath.md
mirjak: Yes, I was planning to raise a separate issue and PR after this one is merged. I think that would be cleaner.

draft-ietf-quic-multipath.md
mirjak: Yes, I agree that we need to change. Just need to decide with number.

draft-ietf-quic-multipath.md
mirjak: No, the normative part is the active part ("SHOULD send"). This other sentence should not be normative because it's just explaining that this can happen and needs to be considered.

draft-ietf-quic-multipath.md
mirjak: Yes, I was also planning to raise a separate issue and PR on this one after the merge.

draft-ietf-quic-multipath.md
qdeconinck: Just committed 13e6610576 to address this.

qdeconinck (approved):
With my added commit (13e661057), LGTM.

yfmascgy (approved):

huitema (approved):
Yes, I approve. It looks very much like I remember of my first review of this PR, with nice fixes.

PR #161, opened 2023-01-30
Move most text of former PNS/CID section to where it belongs

mirjak:
This is the first PR of 3-4 editorial passes I'm planning to do. This first PR only move existing text into the right place. I did not change anything in the text (except the section heading).

The next round will move some sections entirely (e.g. section on congestion control, RTT calculation, packet scheduling could maybe go into the implementation considerations) and then I'm planning to do a full edit pass at the end to check consistency. Would be great if we could review and merge this one soon, so I can move on with the next one. I'm keeping those PRs separately so it is easier to follow the changes.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -189,6 +189,13 @@ used in 1-RTT packets and ACK_MP frames to distinguish packet number
 spaces for different paths. Both identifiers have the same value, which
 is the sequence number of the connection ID, when a new path is established.
 
+The initial path that is used during the handshake (and multipath negotiation)
+has the path ID 0 and therefore all 0-RTT packets are also tracked and
+processed with the path ID 0.
+For 1-RTT packets, the path ID is the sequence number of
+the Destination Connection ID present in the packet header, as defined
+in {{Section 5.1.1 of QUIC-TRANSPORT}}.
+
 # High-level overview {#overview}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -189,6 +189,13 @@ used in 1-RTT packets and ACK_MP frames to distinguish packet number
 spaces for different paths. Both identifiers have the same value, which
 is the sequence number of the connection ID, when a new path is established.
 
+The initial path that is used during the handshake (and multipath negotiation)
+has the path ID 0 and therefore all 0-RTT packets are also tracked and
+processed with the path ID 0.
+For 1-RTT packets, the path ID is the sequence number of
+the Destination Connection ID present in the packet header, as defined
+in {{Section 5.1.1 of QUIC-TRANSPORT}}.
```
Yanmei-Liu, line 197: I'd suggest that we only keep Line 192 to 194, it's useful for readers to understand the initial path identifier.
Line 195 to 197 is very similar with the definition of Path Identifier in Line 175 to 177.

Yanmei-Liu (approved):
Basically very useful restructure. Looks good to me.

yfmascgy (approved):
Loos good to me:)

draft-ietf-quic-multipath.md
Yanmei-Liu: That's ok.

PR #165, opened 2023-02-14
Fix path closure example with path ID

### LINKED ISSUE #164: Unclear (inconsistent?) mapping of Path ID in PATH_STATUS and PATH_ABANDON frames
Since #156, we now define the Path ID as (Section 1.1)
```
the sequence number of the destination Connection ID
used for sending packets on that particular path.
```

And the definition of the `Path ID` as the field of PATH_ABANDON and PATH_STATUS frames directly maps to that identifier path (i.e., the sequence number of the destination Connection ID used for sending packets on that particular path, Sections 8.1 and 8.2).

But in Section 6.2, the example states that 
```
[the Path ID] identifies the path by the sequence number of the received packet's DCID over that path
```

I believe the current text is unclear and should be fixed. Yet, I see two possible ways of doing so. We should either state that

- the `Path ID` field of these frames refer to the Path ID used by the receiver of the frame, or
- add some information to indicate that it refers to the Path ID viewed from the sender's/receiver's viewpoint (that could be encoded as a bit in the frame's type)

mirjak:
fixes #164

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -744,11 +744,11 @@ second path, the server's 1-RTT packets use DCID C2, which has a sequence
 number of 2; the client's 1-RTT packets use DCID S3, which has a sequence number
 of 3. Note that the paths use different packet number spaces. In this case, the
 client is going to close the first path. It identifies the path by the sequence
-number of the received packet's DCID over that path,
-hence using the path_id 1. Optionally, the server confirms the path closure
+number of the DCID used for sending packets over that path,
+hence using the path_id 2. Optionally, the server confirms the path closure
 by sending an PATH_ABANDON frame using
-the sequence number of the received packet's DCID over that path as path
-identifier, which corresponds to the path_id 2. Both the client and
+the sequence number of the DCID used to send over that path as path
+identifier, which corresponds to the path_id 1. Both the client and
 the server can close the path after receiving the RETIRE_CONNECTION_ID frame
 for that path.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -744,26 +744,26 @@ second path, the server's 1-RTT packets use DCID C2, which has a sequence
 number of 2; the client's 1-RTT packets use DCID S3, which has a sequence number
 of 3. Note that the paths use different packet number spaces. In this case, the
 client is going to close the first path. It identifies the path by the sequence
-number of the received packet's DCID over that path,
-hence using the path_id 1. Optionally, the server confirms the path closure
+number of the DCID used for sending packets over that path,
```
qdeconinck, line 747:
```suggestion
number of the DCID it uses for sending packets over that path,
```

draft-ietf-quic-multipath.md
```diff
@@ -744,26 +744,26 @@ second path, the server's 1-RTT packets use DCID C2, which has a sequence
...
+hence using the path_id 2. Optionally, the server confirms the path closure
 by sending an PATH_ABANDON frame using
-the sequence number of the received packet's DCID over that path as path
-identifier, which corresponds to the path_id 2. Both the client and
+the sequence number of the DCID used to send over that path as path
```
qdeconinck, line 750:
```suggestion
the sequence number of the DCID it uses to send over that path as path
```

qdeconinck (commented):
I would also suggest to slightly update the text related to the `Path ID` field of the PATH_ABANDON and PATH_STATUS frames to indicate it refers to the sender's Path ID.

draft-ietf-quic-multipath.md
```diff
@@ -744,26 +744,26 @@ second path, the server's 1-RTT packets use DCID C2, which has a sequence
...
 (client tells server to abandon a path)
-1-RTT[X]: DCID=S2 PATH_ABANDON[path_id=1]->
+1-RTT[X]: DCID=S2 PATH_ABANDON[path_id=2]->
                            (server tells client to abandon a path)
-                      <-1-RTT[Y]: DCID=C1 PATH_ABANDON[path_id=2],
+                      <-1-RTT[Y]: DCID=C1 PATH_ABANDON[path_id=1],
                                                ACK_MP[Seq=2, PN=X]
 (client retires the corresponding CID)
-1-RTT[U]: DCID=S3 RETIRE_CONNECTION_ID[2], ACK_MP[Seq=1, PN=Y] ->
+1-RTT[U]: DCID=S3 RETIRE_CONNECTION_ID[1], ACK_MP[Seq=1, PN=Y] ->
```
yfmascgy, line 764:
```suggestion
1-RTT[U]: DCID=S3 RETIRE_CONNECTION_ID[2], ACK_MP[Seq=1, PN=Y] ->
```
Client wants to retire DCID S2, so the sequence number field in the retire_connection_ID should be 2.

draft-ietf-quic-multipath.md
```diff
@@ -744,26 +744,26 @@ second path, the server's 1-RTT packets use DCID C2, which has a sequence
...
                             (server retires the corresponding CID)
- <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[1], ACK_MP[Seq=3, PN=U]
+ <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[2], ACK_MP[Seq=3, PN=U]
```
yfmascgy, line 766:
```suggestion
 <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[1], ACK_MP[Seq=3, PN=U]
```
Server wants to retire DCID C1, so the sequence number field in the retire_connection_ID should be 1.

draft-ietf-quic-multipath.md
mirjak: Yes, I guess that was wrong in the example.

Yanmei-Liu (approved):
LGTM

PR #168, opened 2023-02-20
Clarify supported scenarios

### LINKED ISSUE #159: Clarify that there are no limitations on the use of server IP address
This draft only enables clients to open a new path, so the most common scenario is that the client has another interface and opens a new path from a new IP address. However, if the client knows another IP that the server is listing on to accept new paths for an existing quic connection (e.g. provided on the higher layer or out of band), that can be used as well. It would be useful to discussion this case more in the draft to ensure that it is clear that this case is supported.

qdeconinck:
Fix #159.


mirjak (approved):

Yanmei-Liu (approved):

Yanmei-Liu (approved):
LGTM

Issue #169, opened 2023-02-21
Does the Path ID change if the CID changes or not?


mirjak:
Currently it seems that the assumption is that the Path ID changes if the CID changes, however, this makes handling of MP frames that contain the path ID more complicated because you have to remember the old path ID to process these frames even when the CID is retired. Is there any good reason why we don't keep the Path ID constant over the life time of a path?

mirjak:
Also note if you retire an CID (e.g. because it's requested by the peer in the NEW_CONNECTION_ID Frame Retire Prior To field) and you don't have a new CID available anymore (yes, this should usually not happen), you cannot send a PATH_ABANDON frame anymore because you don't have valid Path ID (and we don't have the type field anymore).

huitema:
The sequence number is tied to the CID, because of encryption. If we tie it to something else, we are just moving the problem, because then we have to worry about mapping the CID to a path ID.

Yes, this has a couple of constraints. If a CID is actually retired, there is not way to acknowledge the packets that were sent using that CID. Effectively, they are treated as lost. The obvious solution is to write "don't do that" guidelines. Basically, wait a couple RTO before retiring a CID, so the ACK have time to arrive.

The other constraint is that ACK-MP carry the ID used when sending the packets. This is not a huge constraint, but it is an exception to the proposed "same path for ACK" policy. But the main argument for that same path policy is about computing timers. That argument does not really apply when renewing the CID.

Same issue for the path abandon. If packets are sent on the same four tuple with a new CID, just abandon that.

yfmascgy:
In addition to the encryption consideration as mentioned above. I remember there was another reason why the sequence number of CID was chosen to identify a path. It brings us back to the definition of what is a path and what is a path identifier.

By definition of the draft and RFC9000, a path is defined by a 4-tuple {source IP address, source port number, destination IP address, destination port number}, and a path ID is an identifier used to identify a path. That means we need to construct a mapping between a path ID and the associated 4-tuple.

However, there are couple of issues. First, the client's view and the server's view of the 4-tuple are not the same due to NAT.  Second, the 4-tuple can change due to (1) network path migration (think about the case where our phone's wifi is connected to a wifi router that has two outbound networks to the internet, one wireline and one satellite, and the router can switch between the two), and (2) NAT rebinding. If we use a constant path ID, then the path ID does not uniquely identify a 4-tuple, but instead, there will be a one-to-many mappings from a path ID to 4-tuples. 

[...]

PR #171, opened 2023-02-21
What happens when a CID is retired and no new CID is available anymore

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
 Receiving a RETIRE_CONNECTION_ID frame causes the endpoint to discard
-the resources associated with that connection ID. If the connection ID
+the resources associated with that Connection ID. If the connection ID
 was used by the peer to identify a path from the peer to this endpoint,
 the resources include the list of received packets used to send
-acknowledgements. The peer MAY decide to keep sending data using
+acknowledgements. The peer can keep sending data using
 the same IP addresses and UDP ports previously associated with
 the connection ID, but MUST use a different connection ID when doing so.
+E.g this can happen if the Connection ID issuer requests retirement of a
```
mirjak, line 457:
```suggestion
E.g this can happen if the Connection ID issuer requests retirement of a
```

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
```
qdeconinck, line 457:
```suggestion
This can happen if, e.g., the Connection ID issuer requests retirement of a
```

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+Connection ID using the Retire Prior To field in the NEW_CONNECTION_ID Frame.
```
qdeconinck, line 458:
```suggestion
Connection ID using the Retire Prior To field in the NEW_CONNECTION_ID frame.
```

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+If no new connection ID is available anymore, the endpoint cannot send on
+this path anymore but has to wait for the idle time-out before closing
```
qdeconinck, line 460: I'm not sure that the endpoint "has to wait" the idle timeout to close the path. I could also send a PATH_ABANDON to explicitly close the path, right? Instead, do we want to state here that an inactivity larger than the idle timeout implies an implicit path closure?

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+the path. If another active path is available, the enpoint can re-active
```
qdeconinck, line 461:
```suggestion
the path. If another active path is available, the enpoint can re-activate
```

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+the path before the timer expired if a new Connection ID gets issued
```
qdeconinck, line 462:
```suggestion
the path before the timer expires if a new Connection ID gets issued
```

qdeconinck (changes requested):
We definitely need to add some text to better explain what happens in such cases, but I think we need to be a bit clearer on the implicit path closure case (as we may want to explicitly close an available path before idle timeout).

draft-ietf-quic-multipath.md
mirjak: Currently it can't send an PATH_ABANDON because if we don't have a CID we don't have a valid path ID.

qdeconinck: Oh indeed, I went too quick on this one :-)

qdeconinck (approved):

draft-ietf-quic-multipath.md
mirjak: I do mention this issue in #169

mirjak: Actually we could say that the endpoint should first send the path abandoned frame and then the retire CID frame. However, those frames probably don't have to be processed in order and we need to address issue #137 properly. Or alternative you have to wait until the path abandoned frame is ack'ed and then send the retire CID frame but that would introduce a delay. Not sure how bad that is.

mirjak:
After author discussion: we need to also add a sentence about control frames

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,19 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+acknowledgements. The peer can keep sending data using
 the same IP addresses and UDP ports previously associated with
 the connection ID, but MUST use a different connection ID when doing so.
+This can happen if, e.g., the Connection ID issuer requests retirement of a
+Connection ID using the Retire Prior To field in the NEW_CONNECTION_ID frame.
+If no new connection ID is available anymore, the endpoint cannot send on
+this path anymore but has to wait for the idle time-out before closing
+the path. If another active path is available, the enpoint can re-activate
+the path before the timer expires if a new Connection ID gets issued
+by its peer.
```
mirjak, line 463:
```suggestion
this path and is not able to send control frames associated to this path anymore.
However, it can still acknowledge packets received on the path, by sending ACK_MP frames 
on another path, if available. If no data is received on this path, the idle time-out will close
the path. The endpoint can re-activate the path if a new Connection ID gets issued
by its peer over another path before the idle timer expires.
```

draft-ietf-quic-multipath.md
```diff
@@ -448,12 +448,20 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
 was used by the peer to identify a path from the peer to this endpoint,
 the resources include the list of received packets used to send
-acknowledgements. The peer MAY decide to keep sending data using
+acknowledgements. The peer can keep sending data using
 the same IP addresses and UDP ports previously associated with
 the connection ID, but MUST use a different connection ID when doing so.
+This can happen if, e.g., the Connection ID issuer requests retirement of a
+Connection ID using the Retire Prior To field in the NEW_CONNECTION_ID frame.
+If no new connection ID is available anymore, the endpoint cannot send on
+this path and is not able to send control frames associated to this path anymore.
```
qdeconinck, line 460: Given the clarification in #185, Multipath control frames include the DCID sequence number used by their peer. So if an endpoint sends a RETIRE_CONNECTION_ID, it cannot send on that path anymore if it does not have spare CIDs anymore. But when the receiver receives the RETIRE_CONNECTION_ID (which may have issued the NEW_CONNECTION_ID with the increased Retire Prior To beforehand), it cannot send control frames related to that path anymore. I think there is a difference here w.r.t. the current text.

draft-ietf-quic-multipath.md
```diff
@@ -447,15 +447,36 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+
+If the peer cannot send on a path and no data is received on the path, the idle time-out will close
+the path. If, before the idle timer expires, a new Connection ID gets issued
+by its peer over another path, the endpoint can re-activate the path by
+sending a packet with a new Connection ID on that path.
+
+If the sender retires a Connection ID that is still used by
 in-flight packets, it may receive ACK_MP frames referencing the retired
 Connection ID. If the sender stops tracking sent packets with retired
 Connection ID, these would be spuriously marked as lost. To avoid such
```
mirjak, line 482: This paragraph is still correct but doesn't fit perfectly anymore. I guess we could try to reword now or later...

draft-ietf-quic-multipath.md
```diff
@@ -447,15 +447,36 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
 ### Effect of RETIRE_CONNECTION_ID Frame {#retire-cid-close}
 
-Receiving a RETIRE_CONNECTION_ID frame causes the endpoint to discard
-the resources associated with that connection ID. If the connection ID
-was used by the peer to identify a path from the peer to this endpoint,
-the resources include the list of received packets used to send
-acknowledgements. The peer MAY decide to keep sending data using
+Receiving a RETIRE_CONNECTION_ID frame causes an endpoint to discard
+the resources associated with that Connection ID. Specifically, the endpoint
+should not use the sequence number of the retired connection ID anymore in
```
qdeconinck, line 452:
```suggestion
should not use the sequence number of the retired Connection ID anymore in
```

draft-ietf-quic-multipath.md
```diff
@@ -447,15 +447,36 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+any control frames, as the peer will not be able to associate those frames to
+a path and will therefore ingore them. This means an endpoint is also not required
```
qdeconinck, line 454:
```suggestion
a path and will therefore ignore them. This means an endpoint is also not required
```

draft-ietf-quic-multipath.md
```diff
@@ -447,15 +447,36 @@ PATH_CHALLENGE arrives in a packet using a non-zero length Connection ID.
...
+Note that even if a peer cannot send on a path anymore because it does not have
+a valid Connection ID to use, it can still acknowledge packets received on the path,
+by sending ACK_MP frames on another path, if available. But also note that,
+as there is no valid CID associated with the path, the other end cannot send
+multipath control frames that contain the sequence number of a Connection ID, such
+as PATH_ABANDON or PATH_STATUS.
+
+If the peer cannot send on a path and no data is received on the path, the idle time-out will close
+the path. If, before the idle timer expires, a new Connection ID gets issued
+by its peer over another path, the endpoint can re-activate the path by
```
qdeconinck, line 476: Actually, the peer could still be able to send packets on the path the endpoint does not have DCID anymore. The "over another path" is hence not necessarily true.

```suggestion
by its peer, the endpoint can re-activate the path by
```

qdeconinck (commented):
I think we are nearly good, I just suggest to remove a (maybe wrong) precision.

mirjak:
This seems ready to merge. Plan is to merge end of tomorrow (Friday).

huitema (approved):
Yes, this looks good.

PR #172, opened 2023-02-21
Clarifications on connection migration

### LINKED ISSUE #160: Clarify support of migration
RFC9000 allows endpoints to change the CID any time on a path. So it's not automatically a new path if a new CID is seen. This should be noted explicitly in the draft.

Further, also migration can happened without endpoint control (NAT rebinding). In this case the server will see a new 4-tuple but no CID change. In line with RFC9000 the server should start path validation and then the client should use a new CID. In case of multipath support we need to clarify if and when this is treated as a new path. If seen as a new path, we would probably need to close the old one (which however is not functional anymore). However, if e.g. only the port number changes, RFC9000 also would also to not reset the congestion control. In this case it would make sense to not treat it as a new path.

mirjak:
fixes #160

As I added this text into the implementation consideration section, I on purpose did not use normative language. However, we might want to double check that the normative language we used elsewhere in the document is covering this appropriately.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -894,7 +894,7 @@ In this case the server performs path validation (see {{Section 9 of QUIC-TRANSP
 Moreover, {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
 can change the connection ID it uses for a peer to another available one
 at any time during the connection. As such a sole change of the Connection
-ID without any change in the address does indiacte a path. 
+ID without any change in the address does not indicate a new change. 
 
 If an endpoint uses a new Connection ID after an idle period
 and a NAT binding leads to a 4-tuple changes on the same packet,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -912,7 +912,7 @@ the receiving endpoint may not be able to associate the packet to
 an existing path and will therefore consider this as a new path.
 This leads to an inconsistent view of open paths at both peers,
 however, as the "old" path will not work anymore, it will be silently
-closed after the idle timeout expires. 
+closed after the idle timeout expires.
 
 
 # New Frames {#frames}
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -897,7 +897,7 @@ networks events such as NAT rebinding may make the packet's receiver
 observe a different 4-tuple. Servers observing a 4-tuple change will
 performs path validation (see {{Section 9 of QUIC-TRANSPORT}}).
 If the path validation process succeeds, the endpoints need to reset
-their congestion controller and round-trip time estimator for he new
+their congestion controller and round-trip time estimator for the new
 path as specficied in  {{Section 9.4 of QUIC-TRANSPORT}}.
 
 {{Section 9.3 of QUIC-TRANSPORT}} allows an endpoint to skip validation of
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -885,6 +885,37 @@ two or more active paths during the connection lifetime. Different applications 
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
+## Migration
+
+Even when the multipath extension is used, migration can happen.
+E.g. the client's source port and/or IP address may change due to changes
+in network topology or address mappings, such as caused by NAT rebinding.
+In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}}
+and both peers are expected to use new connection IDs during this process.
+
+Even is the path validation is causes by a migration event, when
+the multipath extension is used, the
+endpoint treats the receipt of a PATH_CHALLENGE frame as a new path
+and also initiates path validation to open the path.
+This is important to ensure
+that both peers have the same view on the number of active paths. However,
+this also implicitly means that every migration event will cause a congestion
+control reset, even if e.g. only the source port changed. Further, if the
+old path after a migration event is not usable anymore, with the multipath
+extension the endpoint might send a PATH_ABANDON frame to close the path.
+
+{{Section 9.3 of QUIC-TRANSPORT}} allows an endpoint to skip validation of
+a peer address if that address has been seen recently. However, when the
+multipath extension is used and an endpoint has multiple addresses that
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead. Otherwise switching paths without path validation
+could cause an inconsistent view of open paths at both peer.
+
+Moreover, {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the connection ID it uses for a peer to another available one
+at any time during the connection. As such a sole change of the Connection
+ID without any change in the address does open a new path.
+
 # New Frames {#frames}
 
 All the new frames MUST only be sent in 1-RTT packet, and MUST NOT
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -884,31 +884,35 @@ two or more active paths during the connection lifetime. Different applications 
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
-## Migration
+## Connection ID Changes and NAT Rebindings
 
-Even when the multipath extension is used, migration can happen.
-E.g. the client's source port and/or IP address may change due to changes
-in network topology or address mappings, such as caused by NAT rebinding.
-In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}}).
-
-Moreover, {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
-can change the connection ID it uses for a peer to another available one
+{{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the connection ID it uses for to another available one
 at any time during the connection. As such a sole change of the Connection
-ID without any change in the address does not indicate a new change. 
+ID without any change in the address does not indicate a path change and
+the endpoint will keep the same congestion control and RTT measurement state.
 
-If an endpoint uses a new Connection ID after an idle period
-and a NAT binding leads to a 4-tuple changes on the same packet,
-the receiving endpoint may not be able to associate the packet to
-an existing path and will therefore consider this as a new path.
-This leads to an inconsistent view of open paths at both peers,
-however, as the "old" path will not work anymore, it will be silently
-closed after the idle timeout expires. 
+While endpoints assign a Connection ID to a specific sending 4-tuple,
+networks events such as NAT rebinding may make the packet's receiver
+observe a different 4-tuple. Servers observing a 4-tuple change will
+performs path validation (see {{Section 9 of QUIC-TRANSPORT}}).
+If the path validation process succeeds, the endpoints need to reset
+their congestion controller and round-trip time estimator for he new
+path as specficied in  {{Section 9.4 of QUIC-TRANSPORT}}.
 
 {{Section 9.3 of QUIC-TRANSPORT}} allows an endpoint to skip validation of
 a peer address if that address has been seen recently. However, when the
 multipath extension is used and an endpoint has multiple addresses that
 could lead to switching between different paths, it should rather maintain
 multiple open paths instead.
+
+If an endpoint uses a new Connection ID after an idle period
+and a NAT rebinding leads to a 4-tuple changes on the same packet,
+the receiving endpoint may not be able to associate the packet to
+an existing path and will therefore consider this as a new path.
+This leads to an inconsistent view of open paths at both peers,
+however, as the "old" path will not work anymore, it will be silently
+closed after the idle timeout expires. 
 
 
 # New Frames {#frames}
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -890,10 +890,10 @@ on each of these paths before the idle timeout expires.
 Even when the multipath extension is used, migration can happen.
 E.g. the client's source port and/or IP address may change due to changes
 in network topology or address mappings, such as caused by NAT rebinding.
-In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}}
+In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}})
 and both peers are expected to use new connection IDs during this process.
 
-Even is the path validation is causes by a migration event, when
+Even is the path validation is caused by a migration event, when
 the multipath extension is used, the
 endpoint treats the receipt of a PATH_CHALLENGE frame as a new path
 and also initiates path validation to open the path.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -110,8 +110,7 @@ Connection IDs in both directions.
 address as well as source and destination port. Therefore, there can be
 at most one active paths/connection ID per 4-tuple.
   * If the 4-tuple changes without the use of a new connection ID (e.g.
-due to a NAT rebinding), this is considered as a migration event and not
-as a new path.
+due to a NAT rebinding), this is considered as a migration event.
 
 The path management specified in {{Section 9 of QUIC-TRANSPORT}}
 fulfills multiple goals: it directs a peer to switch sending through
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -889,31 +889,27 @@ on each of these paths before the idle timeout expires.
 Even when the multipath extension is used, migration can happen.
 E.g. the client's source port and/or IP address may change due to changes
 in network topology or address mappings, such as caused by NAT rebinding.
-In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}})
-and both peers are expected to use new connection IDs during this process.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -885,6 +885,37 @@ two or more active paths during the connection lifetime. Different applications
...
+multipath extension is used and an endpoint has multiple addresses that
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead. Otherwise switching paths without path validation
+could cause an inconsistent view of open paths at both peer.
+
+Moreover, {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the connection ID it uses for a peer to another available one
+at any time during the connection. As such a sole change of the Connection
+ID without any change in the address does open a new path.
+
```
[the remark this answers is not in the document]
mirjak: First, I don't think it is true that this is not aligned with RFC9000 because there is always just one path in RFC9000. However, the problem is that in RFC9000 both ends don't need a common understanding of the path but now they do, so we need to address this. In our definition in this draft every 4-tuple is a different path and sticking to that makes it less complicated. The problem with NAT rebinding is that only one end see a new 4-tuple. As such we need an explicit signal to tell the other end to create a new path. I propose to simple use path validation for this (to keep change minimal).

If you are concerned about resetting the congestion window, we can say that the new path can be initialisierend with the cc values from the old path if only the source port changes.

qdeconinck (changes requested):
I don't think the added text is right.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +885,37 @@ two or more active paths during the connection lifetime. Different applications
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
+## Migration
+
+Even when the multipath extension is used, migration can happen.
+E.g. the client's source port and/or IP address may change due to changes
+in network topology or address mappings, such as caused by NAT rebinding.
+In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}}
+and both peers are expected to use new connection IDs during this process.
```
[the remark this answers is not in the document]
mirjak: My thinking was that if you receive a path challenge from the server, you should treat this as a "request" to open a new path and thus use a new CID even if the 4-tuple seem to correspond to an existing path.

However, thinking further I guess you also have to close the existing path at the same time as you should not have two paths with the same 4-tuple. Thus the client MUST send a path abandon frame for the old path.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +885,37 @@ two or more active paths during the connection lifetime. Different applications
...
+
+Even is the path validation is causes by a migration event, when
+the multipath extension is used, the
+endpoint treats the receipt of a PATH_CHALLENGE frame as a new path
+and also initiates path validation to open the path.
```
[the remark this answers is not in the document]
mirjak: I think this is something we need to decide.

The problem is that the server sees a new 4-tuple but the client not, so we need to make clear if this creates a new path or not in such a way that both ends come to the same conclusion.

You could also say it's not a new path if on server side the 4-tuple changes but not the CID and on client side a path challenge (with new CID) is received on an existing 4-tuple. However, you still need to reset congestion control in most cases, so I thought it was easier to just create a new path.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +885,37 @@ two or more active paths during the connection lifetime. Different applications
...
+This is important to ensure
+that both peers have the same view on the number of active paths. However,
```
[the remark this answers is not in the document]
mirjak: If both the 4-tuple and the CID changes, the client should send a path challenge in that packet to explicitly open a new path. Otherwise the server should probably ignore that packet as it can't associate it to an exiting path. Maybe something we should also spell out more clearly.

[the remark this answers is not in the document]
mirjak: Actually ignoring might not be right to address this change where you restart after an idle period with a new CID and then also the NAT rebinding happens... however, it would be the same procedure as in the case when the CID not changes after a NAT rebinding: the server sends a path challenge frame and that would be the signal for the client to assign a new path and use a new CID + close the old path on the presumedly same 4-tuple from the client's point of view.

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
mirjak: And you still need to address the case where both the CID and the 4-tuple changes. As you can't decide to which existing path you should assign this packet, I guess you have to assume it's a new path. Therefore it better if the client also opens a new path when it received the path challenge and explicitly closes the old path.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,33 @@ two or more active paths during the connection lifetime. Different applications
...
+at any time during the connection. As such a sole change of the Connection
+ID without any change in the address does indiacte a path. 
+
+If an endpoint uses a new Connection ID after an idle period
+and a NAT binding leads to a 4-tuple changes on the same packet,
+the receiving endpoint may not be able to associate the packet to
+an existing path and will therefore consider this as a new path.
+This leads to an inconsistent view of open paths at both peers,
+however, as the "old" path will not work anymore, it will be silently
+closed after the idle timeout expires.
```
mirjak, line 905: I updated the text but I'm not really happy here. The server has to assume a new path if it can't associate a packet with new CID and new 4-tuple to an existing path but it might still try to use the old path to send data. Those packets will then be lost and it takes a timeout to detect that this is a the case. Isn't there a way for the client to say: ups I received a path validation even though I din't know my address change, so please don't send data using this old CID anymore because it's actually the same path...?

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
huitema: The node that renewed the CID will deduce the path change from invoming traffic on same (incoming) CID and new 4-tuple. It can at any point abandon the old path or retire the old CID.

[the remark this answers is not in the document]
mirjak: The problem is that after removing the type we used the sender’s cid in the abandon frame and then this doesn’t work as the client only sees one path. However I think we need to fix that an then recommend to send an abandon frame for the old received cid.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,33 @@ two or more active paths during the connection lifetime. Different applications
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
+## Migration
+
```
huitema, line 888: Migration in RFC 9000 means "move from sending and receiving on path 1 to sending and receiving on path 2". I think that multipath implementation just replaces the RFC 9000 concept to "migration", replacing it by a multiphase process in which path 2 is explicitly tried and validated, and then path 1 might be kept going, or might be explicitly abandoned if not useful any more.

We are discussing here two exceptions to that process: NAT Rebinding, and CID renewal. Mabe we should make that into two sections, instead of a single migration section.

huitema (commented):
I don't think that this is ready. I would suggest to not have a migration section, but instead have two sections for NAT Rebinding and CID renewal.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,33 @@ two or more active paths during the connection lifetime. Different applications
...
+however, as the "old" path will not work anymore, it will be silently
+closed after the idle timeout expires. 
+
+{{Section 9.3 of QUIC-TRANSPORT}} allows an endpoint to skip validation of
+a peer address if that address has been seen recently. However, when the
+multipath extension is used and an endpoint has multiple addresses that
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead.
+
+
```
qdeconinck, line 913: I wonder if the following would make it more clear by avoiding the notion of "migration".

Note that I think we could have a better behaviour in the case "both 4-tuple and CID change", but I would prefer delegating this to a later issue.
```suggestion
## NAT Rebinding {nat-rebinding}

While endpoints assign a Connection ID to a specific sending 4-tuple,
networks events such as NAT rebinding may make the packet's receiver
perceives a different 4-tuple. Servers observing 4-tuple changes must
initiate path validation. If the path validation process succeeds, the
endpoint follows considerations of {{Section 9.4 of QUIC-TRANSPORT}}.

## CID Renewal

At any time, endpoints may change the Connection ID they use to send
packets on a path. Such a path can keep the same congestion control
state and RTT measurements.

If an endpoint uses a new Connection ID after an idle period
and a NAT rebinding leads to a 4-tuple changes on the same packet,
the receiving endpoint may not be able to associate the packet to
an existing path and will therefore consider this as a new path.
This leads to an inconsistent view of open paths at both peers,
however, as the "old" path will not work anymore, it will be silently
closed after the idle timeout expires.

```

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,33 @@ two or more active paths during the connection lifetime. Different applications
...
+Even when the multipath extension is used, migration can happen.
+E.g. the client's source port and/or IP address may change due to changes
+in network topology or address mappings, such as caused by NAT rebinding.
+In this case the server performs path validation (see {{Section 9 of QUIC-TRANSPORT}}).
+
+Moreover, {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the connection ID it uses for a peer to another available one
+at any time during the connection. As such a sole change of the Connection
+ID without any change in the address does indiacte a path.
```
mirjak, line 897:
```suggestion
ID without any change in the address does not indicate a new change. 
```

draft-ietf-quic-multipath.md
mirjak: Sorry I did another change and lost you proposal but will edit the PR directly now accordingly. Thx!

draft-ietf-quic-multipath.md
mirjak: I kept it in one section but am not talking about migration anymore but CID changes and NAT rebinding instead. Better?

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,37 @@ two or more active paths during the connection lifetime. Different applications
...
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead.
+
+If an endpoint uses a new Connection ID after an idle period
+and a NAT rebinding leads to a 4-tuple changes on the same packet,
+the receiving endpoint may not be able to associate the packet to
+an existing path and will therefore consider this as a new path.
+This leads to an inconsistent view of open paths at both peers,
+however, as the "old" path will not work anymore, it will be silently
+closed after the idle timeout expires.
```
mirjak, line 915:
```suggestion
closed after the idle timeout expires.
```

huitema (approved):
The updated text works for me.

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,37 @@ two or more active paths during the connection lifetime. Different applications
...
+at any time during the connection. As such a sole change of the Connection
+ID without any change in the address does not indicate a path change and
+the endpoint will keep the same congestion control and RTT measurement state.
+
+While endpoints assign a Connection ID to a specific sending 4-tuple,
+networks events such as NAT rebinding may make the packet's receiver
+observe a different 4-tuple. Servers observing a 4-tuple change will
+performs path validation (see {{Section 9 of QUIC-TRANSPORT}}).
+If the path validation process succeeds, the endpoints need to reset
+their congestion controller and round-trip time estimator for he new
```
mirjak, line 900:
```suggestion
their congestion controller and round-trip time estimator for the new
```

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,37 @@ two or more active paths during the connection lifetime. Different applications
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
+## Connection ID Changes and NAT Rebindings
+
+{{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the connection ID it uses for to another available one
```
mirjak, line 890:
```suggestion
can change the Connection ID it uses for to another available one
```

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,37 @@ two or more active paths during the connection lifetime. Different applications
...
+multiple open paths instead.
+
+If an endpoint uses a new Connection ID after an idle period
+and a NAT rebinding leads to a 4-tuple changes on the same packet,
+the receiving endpoint may not be able to associate the packet to
+an existing path and will therefore consider this as a new path.
+This leads to an inconsistent view of open paths at both peers,
+however, as the "old" path will not work anymore, it will be silently
+closed after the idle timeout expires.
+
```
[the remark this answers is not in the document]
huitema: I think overloading the meaning of PATH_CHALLENGE is a bad idea, see comments on #188. But sending a "RETIRE_CONNECTION_ID" too soon is also a bad idea, because the peer needs some time to send acknowledgements. Maybe the peer doing the CID renewal could send PATH_ABANDON(old_path, reason_code = CID_replaced), with a clear statement explaining that the 4-tuple is still valid, but the CID is not?

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,37 @@ two or more active paths during the connection lifetime. Different applications
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
+## Connection ID Changes and NAT Rebindings
+
+{{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the Connection ID it uses for to another available one
+at any time during the connection. As such a sole change of the Connection
+ID without any change in the address does not indicate a path change and
+the endpoint will keep the same congestion control and RTT measurement state.
```
qdeconinck, line 893: Given that we give implementation considerations, we should not restrict the behaviour too much IMHO.

```suggestion
the endpoint can keep the same congestion control and RTT measurement state.
```

draft-ietf-quic-multipath.md
```diff
@@ -885,6 +884,37 @@ two or more active paths during the connection lifetime. Different applications
...
+
+While endpoints assign a Connection ID to a specific sending 4-tuple,
+networks events such as NAT rebinding may make the packet's receiver
+observe a different 4-tuple. Servers observing a 4-tuple change will
+performs path validation (see {{Section 9 of QUIC-TRANSPORT}}).
+If the path validation process succeeds, the endpoints need to reset
+their congestion controller and round-trip time estimator for the new
+path as specficied in  {{Section 9.4 of QUIC-TRANSPORT}}.
```
qdeconinck, line 901: Why not simply refers to that section directly, that also states that servers may keep their current state if they suspect the event to be a NAT rebinding?
```suggestion
If the path validation process succeeds, the endpoint
follows considerations of {{Section 9.4 of QUIC-TRANSPORT}}.
```

qdeconinck (commented):
Two comments, but then I think the text will be ready.

draft-ietf-quic-multipath.md
mirjak: I would prefer to keep this because I find it annoying to provide a vague reference without giving an indication what this reference is for. Why do you think the current text is not good?

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
mirjak: Let's open an issue to add a reason code, or a replaced by field, or both. Maybe we should also rename path_abandon to cid_abandon?

Question is should we merge this text for now or also open a new issue instead? I think we can merge because it shouldn't hurt to send a path_abandon frame, right?

mirjak:
The current text is ready to merge. There is a remaining question about proposed additional text. If we don't conclude until end of tomorrow (Friday), we will merge as is (without the addition) and create a new PR/issue for that addition instead.

draft-ietf-quic-multipath.md
qdeconinck: Because it somewhat contradicts what Section 9.4 states.

> On confirming a peer's ownership of its new address, an endpoint MUST immediately reset the congestion controller and round-trip time estimator for the new path to initial values (see Appendices [A.3](https://www.rfc-editor.org/rfc/rfc9002#appendix-A.3) and [B.3](https://www.rfc-editor.org/rfc/rfc9002#appendix-B.3) of [[QUIC-RECOVERY](https://www.rfc-editor.org/rfc/rfc9000.html#QUIC-RECOVERY)]) unless the only change in the peer's address is its port number. Because port-only changes are commonly the result of NAT rebinding or other middlebox activity, the endpoint MAY instead retain its congestion control state and round-trip estimate in those cases instead of reverting to initial values.

We could just adapt the text to be like "the endpoint updates the path's congestion controller and round-trip time estimator according to {{Section 9.4 of QUIC-TRANSPORT}}".

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
qdeconinck: I think this issue is somewhat orthogonal, and discussed in #188.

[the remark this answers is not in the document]
huitema: We do have an issue with the server continuing sending data on the old path, now a dead end. I am concerned that we will not have an agreement before Monday. I left comments in issue #188. If we resolve that issue before Friday 3/10, let's incorporate the agreed solution in this PR. If we do not, let's check in the PR as is, without the extra text.

draft-ietf-quic-multipath.md
mirjak:
```suggestion
If path validation process succeeds, the endpoints set
the path's congestion controller and round-trip time
estimator according to {{Section 9.4 of QUIC-TRANSPORT}}.
```

qdeconinck (approved):

PR #173, opened 2023-02-21
Clarify path close and waiting for at least 3 PTOs

mirjak:
The current text says that one has to wait for a the retire frame or 3 PTOs. However (as also indicated in the states figure) this should be at least 3 PTOs. This PR fixes this inconsistency. Further is add a reference to the definition to PTO and replaces RTO with PTO where the wrong term is used. I believe all these fixes address straightforward mistakes, however, if discussion is needed, we can of course also open an issue.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -384,9 +384,6 @@ Both endpoints, namely the client and the server, can close a path,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
 abandons the path with a corresponding Path Identifier.
 
-The endpoint sending the PATH_ABANDON frame can consider a path as
-abandoned when the packet that contained the PATH_ABANDON frame is
-acknowledged.
 
 The receiver of a PATH_ABANDON frame should not release its resources
 immediately, but SHOULD wait for at least three times the current
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -398,7 +398,7 @@ if any.
 
 The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
 immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs but at least three times the current 
+frame for the used connection IDs but at least three times the current
 Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}.
 This is inline with the requirement of {{Section 10.2 of QUIC-TRANSPORT}}
 to ensure that paths close cleanly and that delayed or reordered packets
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -391,7 +391,7 @@ acknowledged.
 The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
 immediately, but SHOULD wait for at least three times the current
 Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}
-after the last time a packet with the corresponding CID or until it receives
+after receive of the PATH_ABANDON frame or until it receives
 a RETIRE_CONNECTION_ID frame for that CID, whichever happens sooner.
 
 Similarly, an endpoint SHOULD wait for three PTOs after the last sent packet
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -53,11 +53,11 @@ normative:
   RFC2119:
   QUIC-TRANSPORT: rfc9000
   QUIC-TLS: rfc9001
+  QUIC-RECOVERY: rfc9002
 
 informative:
   RFC6356:
   I-D.bonaventure-iccrg-schedulers:
-  QUIC-RECOVERY: rfc9002
   QUIC-Invariants: rfc8999
   QUIC-Timestamp: I-D.huitema-quic-ts
   OLIA:
@@ -398,7 +398,11 @@ if any.
 
 The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
 immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs or 3 RTOs.
+frame for the used connection IDs but at least three times the current 
+Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}.
+This is inline with the requirement of {{Sectuion 10.2 of QUIC-TRANSPORT}}
+to ensure that paths close cleanly and that delayed or reordered packets
+are properly discarded.
 
 Usually, it is expected that the PATH_ABANDON frame is used by the client
 to indicate to the server that path conditions have changed such that
@@ -430,7 +434,7 @@ and enters the closing state. If the client received a PATH_ABANDON
 frame for the last open path, it MAY instead try to open a new path, if
 available, and only initiate connection closure if path validation fails
 or a CONNECTION_CLOSE frame is received from the server. Similarly
-the server MAY wait for a short, limited time such as one RTO if a path
+the server MAY wait for a short, limited time such as one PTO if a path
 probing packet is received on a new path before sending the
 CONNECTION_CLOSE frame.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -400,7 +400,7 @@ The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
 immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
 frame for the used connection IDs but at least three times the current 
 Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}.
-This is inline with the requirement of {{Sectuion 10.2 of QUIC-TRANSPORT}}
+This is inline with the requirement of {{Section 10.2 of QUIC-TRANSPORT}}
 to ensure that paths close cleanly and that delayed or reordered packets
 are properly discarded.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -388,17 +388,16 @@ The endpoint sending the PATH_ABANDON frame can consider a path as
 abandoned when the packet that contained the PATH_ABANDON frame is
 acknowledged.
 
-The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
+The receiver of a PATH_ABANDON frame should not release its resources
 immediately, but SHOULD wait for at least three times the current
 Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}
-after receive of the PATH_ABANDON frame or until it receives
-a RETIRE_CONNECTION_ID frame for that CID, whichever happens sooner.
-
-Similarly, an endpoint SHOULD wait for three PTOs after the last sent packet
-before sending the RETIRE_CONNECTION_ID frame for the corresponding CID.
+after the last sent packet before sending the RETIRE_CONNECTION_ID frame
+for the corresponding CID.
 This is inline with the requirement of {{Section 10.2 of QUIC-TRANSPORT}}
 to ensure that paths close cleanly and that delayed or reordered packets
 are properly discarded.
+The effect of receiving a RETIRE_CONNECTION_ID frame is specified in the
+next section.
 
 Usually, it is expected that the PATH_ABANDON frame is used by the client
 to indicate to the server that path conditions have changed such that
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -405,7 +405,7 @@ the path is or will be not usable anymore, e.g. in case of a mobility
 event. The PATH_ABANDON frame therefore recommends to the receiver
 that no packets should be sent on that path anymore.
 In addition, the RETIRE_CONNECTION_ID frame is used indicate to the receiving peer
-that the sender does not intend to send any packets assiciated to the
+that the sender will not send any packets associated to the
 Connection ID used on that path anymore.
 The receiver of an PATH_ABANDON frame MAY also send
 an PATH_ABANDON frame to indicate its own unwillingness to receive
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -373,7 +373,7 @@ Both endpoints, namely the client and the server, can initiate path closure,
 by sending a PATH_ABANDON frame (see {{path-abandon-frame}}) which
 requests the peer to stop sending packets with the corresponding Destination Connection ID.
 
-The receiver of a PATH_ABANDON frame should not release its resources
+The sender and receiver of a PATH_ABANDON frame should not release its resources
 immediately, but SHOULD wait for at least three times the current
 Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}
 after the last sent packet before sending the RETIRE_CONNECTION_ID frame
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -398,7 +398,11 @@ if any.
 
 The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
 immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs or 3 RTOs.
+frame for the used connection IDs but at least three times the current 
+Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}.
+This is inline with the requirement of {{Sectuion 10.2 of QUIC-TRANSPORT}}
```
mirjak, line 403:
```suggestion
This is inline with the requirement of {{Section 10.2 of QUIC-TRANSPORT}}
```

draft-ietf-quic-multipath.md
```diff
@@ -398,7 +398,11 @@ if any.
 
 The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
 immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs or 3 RTOs.
+frame for the used connection IDs but at least three times the current
```
mirjak, line 401:
```suggestion
frame for the used connection IDs but at least three times the current
```

yfmascgy (approved):
Looks good to me.

draft-ietf-quic-multipath.md
```diff
@@ -398,7 +398,11 @@ if any.
...
+Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}.
+This is inline with the requirement of {{Section 10.2 of QUIC-TRANSPORT}}
+to ensure that paths close cleanly and that delayed or reordered packets
+are properly discarded.
```
huitema, line 406: This does not read very well: "SHOULD wait for the reception of the RETIRE_CONNECTION_ID
frame for the used connection IDs but at least three times the current
Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}." I don't know how to interpret it.

I think what you mean is: "the receiver of a PATH_ABANDON frame SHOULD NOT release its resources
immediately, but SHOULD wait for at least 3 RTOs after the last time a packet was sent using the corresponding CID or until it receives a RETIRE_CONNECTION_ID frame for that CID, whichever happens sooner."

Then there is the separate question of the sender behavior. I think we should says something like: "Nodes that Abandon a PATH SHOULD wait for 3 RTOs before sending a RETIRE_CONNECTION_ID frame for the corresponding CID. However, they MAY issue the RETIRE_CONNECTION_ID frame sooner, for example because of resource constraints or because the peer asked to retire all connection identifier before a specific sequence number. This may cause the peer node to conclude that no acknowledgement will be received for packets sent with that CID."

huitema (changes requested):

draft-ietf-quic-multipath.md
mirjak: What you propose in your text is exactly what we need to clarify. I think based on {{Section 6.2. of QUIC-RECOVERY}} you always need to wait for at least 3 PTO. If we say it that way, I don't think there is a need to add the second part you propose to wait with the sending of RETIRE_CONNECTION_ID frame . That would be another solution but more complex.

mirjak: Yes, agree. I'll give it a try. But we do agree that's the right thing?

qdeconinck: I think we agree on the meaning, but we should indeed improve the presentation. What about replacing the first (very) long sentence by this?

```
Upon the reception of a PATH_ABANDON, the endpoint SHOULD delay the release
of the associated path resource until either a) the peer has retired the
Connection ID it used on the abandoned path, or b) at least three times the
current Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}.
```

qdeconinck (changes requested):
I think the meaning is right, but I think we can clarify it (suggestion provided).

draft-ietf-quic-multipath.md
mirjak: Okay this means we do not agree. I believe you always have to wait at least threes PTOs because of re-ordered packets. I guess if you receive the retired CID frame that means that the peer will not send any new packets but you can only be sure that there is no reordering if you received all packets before the packet with the retired CID frame, however, you don't really know how far back you have to look or there could be gaps introduced by the sender. So checking that is not really practical and always waiting for 3 PTOs is therefore just simpler. Or is there is strong reason why it would be needed to release resources on receipt of the retired CID frame already?

I note that this discussion is also related to the question if we change the path ID with the CID or not. If we retire the path ID together with the CID, it makes more sense to act on the retired CID frame because in this case even if you receive reordered packets, you would not be able to process them as you don't have a valid path ID.

Yanmei-Liu: I agree that endpoint should always wait at least three PTOs before it releases resources of the corresponding path, but the sentence need to be fixed. Maybe we need 2 states "Closing Path State" and "Draining Path State": 
- After the endpoint sends the PATH_ABANDON frame, it enters into the "Closing Path State", and needs to wait for at least three PTOs before releasing the CID resource of the corresponding path. 
- The "Draining Path State" is entered once an endpoint receives a PATH_ABANDON frame, which indicates that its peer is closing or draining on this path. "Draining Path State" last for at least three PTOs, before it ends, the endpoint SHOULD NOT release the corresponding CID resouce, until it receives a RETIRE_CONNECTION_ID frame for that CID.

Yanmei-Liu (changes requested):

draft-ietf-quic-multipath.md
mirjak: This PR is actually also connects to the question on path  ID.

First of all it is not clear to me what an endpoints need to do if it first receives a retired CID frame and then a (re-order) packet with that CID. In order word is an endpoint supposed to retire/forget the CID immediately when it received the retire CID frame and discard any packets with the old CID or might it also what 1-3 PTOs for re-order packet? I didn't find anything in RFC9000 but maybe somebody else that is better in finding this has a pointer to right part...?

If the former, meaning you remove it immediately and ignore old packets, that would also mean that you don't have a valid path ID anymore and can't process any ACKs or other control frame with the old path ID anymore. As such you also don't have to wait for 3 PTOs as you might receive packet/frames but can process them anymore. In this case (assuming that the path ID is connected to the CID), one solution could be to recommend to wait 3PTOs before you send the retired CID frame (or just accepting that this might happens and you have time retransmit some packets unnecessarily if you change the CID).

If you have a stable path ID, you can retire immediately and ignore packets with the old CID but still process any control frames that are received in packets with the new CID, and respectively have to wait 3 PTO before you can also remove the path id/close the path entirely.

Or alternatively, you always have to 3 PTOs after the abandon farm before you remove any state including the last used CID but that also means that the retired CID frames doesn't have any impact. And it also means whenever you received a retired CID frame during an active path use, you also need to wait 3 PTOs before you remove it.

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,51 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
-path.
+abandons the path with a corresponding Path Identifier.
 
-The endpoint sending the PATH_ABANDON frame SHOULD consider a path as
+The endpoint sending the PATH_ABANDON frame can consider a path as
 abandoned when the packet that contained the PATH_ABANDON frame is
-acknowledged. When releasing resources of a path, the endpoint SHOULD
-send a RETIRE_CONNECTION_ID frame for the connection IDs used on the path,
-if any.
+acknowledged.
```
mirjak, line 389: not sure we need this sentence at all...

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,51 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
 
 The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
-immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs or 3 RTOs.
+immediately, but SHOULD wait for at least three times the current
+Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}
+after the last time a packet with the corresponding CID or until it receives
```
mirjak, line 394:
```suggestion
after receive of the PATH_ABANDON frame or until it receives
```

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,50 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
+The receiver of a PATH_ABANDON frame should not release its resources
+immediately, but SHOULD wait for at least three times the current
+Probe Timeout (PTO) interval as defined in {{Section 6.2. of QUIC-RECOVERY}}
+after the last sent packet before sending the RETIRE_CONNECTION_ID frame
+for the corresponding CID.
+This is inline with the requirement of {{Section 10.2 of QUIC-TRANSPORT}}
+to ensure that paths close cleanly and that delayed or reordered packets
+are properly discarded.
+The effect of receiving a RETIRE_CONNECTION_ID frame is specified in the
+next section.
```
mirjak, line 400: This is a new normative requirement now but I think this is what is actually meant by "releasing resources". Particularly you should only retire you CID if you are sure you don't need it anymore. Hence the new normative requirement is to wait 3 PTOs before you sent the retired CID frame after receiving the abandon frame.

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,50 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
 Both endpoints, namely the client and the server, can close a path,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
-abandons the path with a corresponding Path Identifier. Once a path is
-marked as "abandoned", it means that the resources related to the path,
-such as the used connection IDs, can be released.
-However, information related to data delivered over that path SHOULD
-not be released immediately as acknowledgments can still be received
-or other frames that also may trigger retransmission of data on another
-path.
+abandons the path with a corresponding Path Identifier.
```
mirjak, line 385: I didn't change this first sentence because that already covered in PR #185 (which we should probably merge first)

draft-ietf-quic-multipath.md
qdeconinck: Agree.

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,50 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
 the path is or will be not usable anymore, e.g. in case of a mobility
-event. The PATH_ABANDON frame therefore indicates to the receiving peer
-that the sender does not intend to send any packets on that path anymore
-but also recommends to the receiver that no packets should be sent in
-either direction. The receiver of an PATH_ABANDON frame MAY also send
-an PATH_ABANDON frame to signal its own willingness to not send
+event. The PATH_ABANDON frame therefore recommends to the receiver
+that no packets should be sent on that path anymore.
+In addition, the RETIRE_CONNECTION_ID frame is used indicate to the receiving peer
+that the sender does not intend to send any packets assiciated to the
```
qdeconinck, line 408: The RETIRE_CONNECTION_ID frame is not an indication, it is stronger.
```suggestion
that the sender will not send any packets associated to the
```

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,50 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
+Connection ID used on that path anymore.
+The receiver of an PATH_ABANDON frame MAY also send
```
qdeconinck, line 410:
```suggestion
The receiver of a PATH_ABANDON frame MAY also send
```

draft-ietf-quic-multipath.md
```diff
@@ -382,55 +382,50 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
+an PATH_ABANDON frame to indicate its own unwillingness to receive
```
qdeconinck, line 411:
```suggestion
a PATH_ABANDON frame to indicate its own unwillingness to receive
```

qdeconinck (commented):
Text is fine, but we should merge #185 before this one IMHO.

draft-ietf-quic-multipath.md
mirjak:
```suggestion
```

draft-ietf-quic-multipath.md
mirjak: agreed

draft-ietf-quic-multipath.md
huitema: Yes, this is good. There are two big motivations: packets may be delivered out of order, with packets in transit arriving after the PATH_ABANDON; and, acknowledgements still need to be sent. Waiting 3 PTO is the standard mitigation for such issues.

huitema (approved):
Looks good.

qdeconinck (approved):
By fixing the conflict, I also fixed #194.

draft-ietf-quic-multipath.md
huitema: The question about "what to do with packets where DCID is retired" is not specific to multipath. The same "out of order delivery issues" could happen with QUIC as well. After a CID is retired, we expect the issuer of the CID to "forget about it". If that CID is forgotten, the load balancer or the QUIC server will not be able to associate it with a QUIC connection context. The QUIC connection will never have to process these packets.

The QUIC server will assume that the client is trying to use a connection that does not exist anymore and could issue a "stateless reset" packet, but since the CID has been retired the token in the stateless reset will not match any valid CID for the connection and will have no effect.

Which is a long winded way to say that packets arriving after the DCID has been retired will be automatically ignored. There is no need to specify a processing for them, because they will never be processed at all.

draft-ietf-quic-multipath.md
```diff
@@ -372,58 +372,45 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
-abandoned when the packet that contained the PATH_ABANDON frame is
-acknowledged. When releasing resources of a path, the endpoint SHOULD
-send a RETIRE_CONNECTION_ID frame for the connection IDs used on the path,
-if any.
 
-The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
-immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs or 3 RTOs.
+The receiver of a PATH_ABANDON frame should not release its resources
+immediately, but SHOULD wait for at least three times the current
```
huitema, line 377:
~~~suggestion
The sender and the receiver of a PATH_ABANDON frame should not release their resources
~~~

draft-ietf-quic-multipath.md
huitema: The suggestion above should address the point raised by @Yanmei-Liu in the resolved conversation below.

draft-ietf-quic-multipath.md
```diff
@@ -372,58 +372,45 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
-The endpoint sending the PATH_ABANDON frame SHOULD consider a path as
-abandoned when the packet that contained the PATH_ABANDON frame is
-acknowledged. When releasing resources of a path, the endpoint SHOULD
-send a RETIRE_CONNECTION_ID frame for the connection IDs used on the path,
-if any.
 
-The receiver of a PATH_ABANDON frame SHOULD NOT release its resources
-immediately, but SHOULD wait for the reception of the RETIRE_CONNECTION_ID
-frame for the used connection IDs or 3 RTOs.
+The receiver of a PATH_ABANDON frame should not release its resources
```
mirjak, line 376:
```suggestion
The sender and receiver of a PATH_ABANDON frame should not release its resources
```

PR #174, opened 2023-02-21
Clarify path usage for PATH_ABANDON frame

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -382,7 +382,15 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
 Both endpoints, namely the client and the server, can close a path,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
-abandons the path with a corresponding Path Identifier. Once a path is
+abandons the path with a corresponding Path Identifier. The PATH_ABANDON
+frame contains the Path Identifier and therefore can be sent on any path.
+Respectively, if the to-be-closed path is assumed to be not working anymore,
+the endpoint might consider to send the PATH_ABANDON frame on another path.
+If the expiration of the PTO causes a retransmission of the PATH_ABANDON frame,
+this might be an inidcation that the path is not usable anymore and the
+endpoint respectively might consider to retransmit on another path.
```
mirjak, line 391:
```suggestion
endpoint respectively might consider to retransmit on another path.
```

draft-ietf-quic-multipath.md
```diff
@@ -382,7 +382,15 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
```
qdeconinck, line 391:
```suggestion
this might be an indication that the sending path is not usable anymore and the
endpoint might consider to retransmit the frame on another path.
```

qdeconinck (approved):
Minor editorial suggestion.

draft-ietf-quic-multipath.md
```diff
@@ -382,7 +382,15 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
 Both endpoints, namely the client and the server, can close a path,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
-abandons the path with a corresponding Path Identifier. Once a path is
+abandons the path with a corresponding Path Identifier. The PATH_ABANDON
+frame contains the Path Identifier and therefore can be sent on any path.
+Respectively, if the to-be-closed path is assumed to be not working anymore,
+the endpoint might consider to send the PATH_ABANDON frame on another path.
+If the expiration of the PTO causes a retransmission of the PATH_ABANDON frame,
+this might be an indication that the sending path is not usable anymore and the
+endpoint might consider to retransmit the frame on another path.
```
Yanmei-Liu, line 391: How about: "If the PATH_ABANDON frame is considered lost in the current path, the endpoint might choose to retransmit the frame on another path if it indicates the path is not usable anymore."

The keypoint is the expiration of the PTO might not be an indication of path blackhole in most cases, but the endpoint could choose to use another path if it indicates the path is unavailable.

mirjak: The reason why I mention PTO here is because if you, receives ACK that indicate a lost, that means the path is probably still usable. It's only if the PTO expires. However, maybe it better to says this more generically like:

"If the PATH_ABANDON frame is considered lost in the current path and another other packet has been received on that path indication that the past is still usable, the endpoint might choose to retransmit the frame on another path."

Or should we even say something stronger than "might"?

yfmascgy: I think retransmitting PATH_ABANDON on another path when PTO happens is reasonable.

yfmascgy (approved):

draft-ietf-quic-multipath.md
mirjak, line 391:
```suggestion
If the PATH_ABANDON frame is considered lost in the current path and 
another other packet has been received on that path indication that the 
past is still usable, the endpoint might choose to retransmit the frame on another path.
```

mirjak, line 391:
```suggestion
```

Issue #183, opened 2023-03-03
Provide more guidance on which path to use for control frames


mirjak:

qdeconinck:
We probably need to add some text in the scheduling considerations where control frames (PATH_STATUS, PATH_ABANDON, but also MAX_DATA,...) should be sent on a good working path, ideally with low latency.

huitema:
I would say the same applies to ACK_MP

mirjak:
For ACK we recommend (SHOULD) to send it on the same path because of RTT calculations and I believe the scheduling text discusses at already.

huitema:
Thinking about this some more. I believe we should say *NOTHING*. The point of the standard is to specify expected behavior when a node receives protocol messages from its peer. The basic statement is that most frames, including control frames, can be sent on any path -- the only exception being PATH CHALLENGE frames and PATH_RESPONSE frames.

So my preference would be:

1) say nothing, because the purpose of standards is to specify requirements for Interop, not provide guidance to implementors.

2) if we do say something, just say that these frames can be sent on any valid path, based on preferences of the application and the implementation.

huitema:
As for ACK_MP frames, the text says "ACK_MP frame (defined in [Section 8.3](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#ack-mp-frame)) SHOULD be sent on the same path as identified by the Path Identifier. However, an ACK_MP frame can be returned via a different path, based on different strategies of sending ACK_MP frames." That's an OK compromise, leaving open the possibility for implementations to use any strategy they want.

LPardue:
Alternatively, spell out the considerations to make when picking a path? Does it matter, if so, why? This would be useful to know when new frames come along in the future and need to think for themselves how multipath might affect them.

huitema:
Sure, but at this stage we have preciously little experience. For example, I have a user request to specify affinity between a stream and a path. I can certainly provide them an API to do that, but this is basically research. And then, affinity applies to stream frames, but whether it should be applied to control frames is debatable.

I am concerned that the guidance that we provide will be so generic as not to be useful, while still requiring long debates and delaying the publication of the draft. I think guidance is best provided in research papers, and possibly in text books.

LPardue:
I agree we shoukd avoid the potential time sink.

[...]

PR #185, opened 2023-03-03
Remove mentions of the "Path Identifier"

### LINKED ISSUE #169: Does the Path ID change if the CID changes or not?

### LINKED ISSUE #181: Regular ACK frames when multipath has been negotiated
# Introduction

It should be allowed to send regular ACK frames also after multipath has been negotiated.

# Reasoning

Sending acknowledgments over other paths than where the packet was received is complex, which is recognised in the Packet Scheduling section of the draft. A straight-forward implementation may therefore never send cross-path acknowledgments. Allowing it to use regular ACK frames would reduce the special-case code for multipath mode and also save at least one byte on every acknowledgment.

The specification should be updated to (explicitly) say that an endpoint MAY send regular ACK frames also after multipath has been negotiated.

qdeconinck:
In lot of places in the draft, we mention the "Path Identifier", which we define as the Destination Connection ID sequence number used over that path. It seems that maintaining the notion of "Path Identifier" causes a lot of confusion, so better keep things simple and mention CID sequence numbers directly.

Note that this implies that the related fields in
multipath-specific have been renamed to their "explicit" form. Some of these names are maybe a bit long, so we may shorten them if this does not introduce ambiguity.

Fix #169. Rewriting related section may also answer #181.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -206,7 +206,7 @@ In this version of the document, a QUIC server does not initiate the creation
 of a path, but it can validate a new path created by a client.
 A new path can only be used once the associated 4-tuple has been validated
 by ensuring that the peer is able to receive packets at that address
-(see {{Section 8 of RFC9000}}). The Destination Connection ID is used
+(see {{Section 8 of QUIC-TRANSPORT}}). The Destination Connection ID is used
 to associate a packet to a packet number space that is used on a valid path. Further, the
 sequence number of Destination Connection ID is used as numerical identifier
 in control frames. E.g. an endpoint sends a PATH_ABANDON frame to request its peer to
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -204,10 +204,10 @@ To add a new path to an existing multipath QUIC connection, a client starts a pa
 the chosen path, as further described in {{setup}}.
 In this version of the document, a QUIC server does not initiate the creation
 of a path, but it can validate a new path created by a client.
-A new path can only be used once it has been validated. Each endpoint
-identifies its paths by the sequence number of the Destination
-Connection ID it uses over them. Such numerical identifier is notably
-used when an endpoint sends a PATH_ABANDON frame to request its peer to
+A new path can only be used once it has been validated. The Destination
+Connection ID is used to associate a packet to a valid path. Further, the
+sequence number of Destination Connection ID is used as numerical identifier
+in control frames. E.g. an endpoint sends a PATH_ABANDON frame to request its peer to
 abandon the path on which the sender uses the Destination Connection ID
 with the sequence number contained in the PATH_ABANDON frame.
 
@@ -580,8 +580,7 @@ The ACK_MP frame, as specified in {{ack-mp-frame}}, is used to
 acknowledge 1-RTT packets.
 Compared to the QUIC version 1 ACK frame, the ACK_MP frame additionally
 contains the Receiver's Destination Connection ID Sequence Number field
-to distinguish the Connection ID-specific packet number space
-acknowledged.
+to distinguish the Connection ID-specific packet number space.
 
 Acknowledgements of Initial and Handshake packets MUST be carried using
 ACK frames, as specified in {{QUIC-TRANSPORT}}. The ACK frames, as defined
@@ -592,7 +591,7 @@ the Connection ID having sequence number 0.
 
 As soon as the negotiation of multipath support is completed,
 endpoints SHOULD use ACK_MP frames instead of ACK frames to acknowledge application
-data packets, including 0-RTT packets, using the Connection ID with
+data packets, including 0-RTT packets, using the initial Connection ID with
 sequence number 0 after the handshake concluded.
 
 ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be sent on the path
@@ -961,7 +960,7 @@ PATH_STATUS Frames contain the following fields:
 Sender's Destination Connection ID Sequence Number:
 : The sequence number of the Destination Connection ID used by the
   sender of this frame to send packets over the path the status update
-  corresponds.
+  corresponds to.
 
 Path Status sequence number:
 : A variable-length integer specifying
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -539,7 +539,7 @@ In Active state, hosts MUST also track the following information:
 
 - Associated Source Connection ID: The Connection ID used to receive
 packets over the path. The endpoint relies on its sequence number to
-receive and acknowledge packets belonging to that Connection ID-specific
+send path control information and specifically acknowledge packets belonging to that Connection ID-specific
 packet number space.
 
 A path in the "Validating" state performs path validation as described
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -340,8 +340,8 @@ PATH_STATUS frame describes 2 kinds of path states:
 - Mark a path as "standby", i.e., suggest that no traffic should be sent
   on that path if another path is available.
 
-Endpoints use Sender's Destination Connection ID Sequence Number field
-in PATH_STATUS frame to identify which path’s state is going to be
+Endpoints use Destination Connection ID Sequence Number field
+in PATH_STATUS frame to identify which path state is going to be
 changed. Notice that PATH_STATUS frame
 can be sent via a different path. An Endpoint MAY ignore the PATH_STATUS frame
 if it would make all the paths unavailable in a single connection.
@@ -375,7 +375,7 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
 Both endpoints, namely the client and the server, can close a path,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
 abandons the path with the corresponding Destination Connection ID.
-The PATH_ABANDON frame contains the Sender's Destination Connection
+The PATH_ABANDON frame contains the Destination Connection
 ID Sequence Number and therefore can be sent on any path.
 
 Once a path is
@@ -575,13 +575,13 @@ the ACK frame as well as packet protection as described in the following subsect
 The ACK_MP frame, as specified in {{ack-mp-frame}}, is used to
 acknowledge 1-RTT packets.
 Compared to the QUIC version 1 ACK frame, the ACK_MP frame additionally
-contains the Receiver's Destination Connection ID Sequence Number field
+contains the receiver's Destination Connection ID Sequence Number field
 to distinguish the Connection ID-specific packet number space.
 
 Acknowledgements of Initial and Handshake packets MUST be carried using
 ACK frames, as specified in {{QUIC-TRANSPORT}}. The ACK frames, as defined
-in {{QUIC-TRANSPORT}}, do not carry the Receiver's Destination
-Connection ID Sequence Number field. If multipath has been successfully
+in {{QUIC-TRANSPORT}}, do not carry the Destination Connection ID
+Sequence Number field. If multipath has been successfully
 negotiated, ACK frames in 1-RTT packets acknowledge packets sent with
 the Connection ID having sequence number 0.
 
@@ -728,11 +728,12 @@ second path, the server's 1-RTT packets use DCID C2, which has a sequence
 number of 2; the client's 1-RTT packets use DCID S3, which has a sequence number
 of 3. Note that the paths use different packet number spaces. In this case, the
 client is going to close the first path. It identifies the path by the sequence
-number of the DCID it uses for sending packets over that path,
-hence using the path_id 2. Optionally, the server confirms the path closure
+number of the DCID its peer uses for sending packets over that path,
+hence using the DCID sequence number 1 (which relates to C1). Optionally, the
+server confirms the path closure
 by sending an PATH_ABANDON frame by indicating the sequence number of the DCID
-it uses to send over that path, which corresponds to the sequence number
-1 (of C1). Both the client and
+the client uses to send over that path, which corresponds to the sequence number
+2 (of S2). Both the client and
 the server can close the path after receiving the RETIRE_CONNECTION_ID frame
 for that path.
 
@@ -740,9 +741,9 @@ for that path.
 Client                                                      Server
 
 (client tells server to abandon a path)
-1-RTT[X]: DCID=S2 PATH_ABANDON[sender_dcid_seq_num=2]->
+1-RTT[X]: DCID=S2 PATH_ABANDON[dcid_seq_num=1]->
                            (server tells client to abandon a path)
-               <-1-RTT[Y]: DCID=C1 PATH_ABANDON[sender_dcid_seq_num=1],
+               <-1-RTT[Y]: DCID=C1 PATH_ABANDON[dcid_seq_num=2],
                                                ACK_MP[PID=2, PN=X]
 (client retires the corresponding CID)
 1-RTT[U]: DCID=S3 RETIRE_CONNECTION_ID[2], ACK_MP[PID=1, PN=Y] ->
@@ -895,7 +896,7 @@ PATH_ABANDON frames are formatted as shown in {{fig-path-abandon-format}}.
 ~~~
   PATH_ABANDON Frame {
     Type (i) = TBD-02 (experiments use 0xbaba05),
-    Sender's Destination Connection ID Sequence Number (i),
+    Destination Connection ID Sequence Number (i),
     Error Code (i),
     Reason Phrase Length (i),
     Reason Phrase (..),
@@ -905,9 +906,9 @@ PATH_ABANDON frames are formatted as shown in {{fig-path-abandon-format}}.
 
 PATH_ABANDON frames contain the following fields:
 
-Sender's Destination Connection ID Sequence Number:
+Destination Connection ID Sequence Number:
 : The sequence number of the Destination Connection ID used by the
-  sender of this frame to send packets over the path to abandon.
+  receiver of the frame to send packets over the path to abandon.
 
 Error Code:
 : A variable-length integer that indicates the reason for abandoning
@@ -944,7 +945,7 @@ PATH_STATUS frames are formatted as shown in {{fig-path-status-format}}.
 ~~~
   PATH_STATUS Frame {
     Type (i) = TBD-03 (experiments use 0xbaba06),
-    Sender's Destination Connection ID Sequence Number (i),
+    Destination Connection ID Sequence Number (i),
     Path Status sequence number (i),
     Path Status (i),
   }
@@ -953,9 +954,9 @@ PATH_STATUS frames are formatted as shown in {{fig-path-status-format}}.
 
 PATH_STATUS Frames contain the following fields:
 
-Sender's Destination Connection ID Sequence Number:
+Destination Connection ID Sequence Number:
 : The sequence number of the Destination Connection ID used by the
-  sender of this frame to send packets over the path the status update
+  receiver of this frame to send packets over the path the status update
   corresponds to.
 
 Path Status sequence number:
@@ -964,7 +965,7 @@ Path Status sequence number:
   number MUST be monotonically increasing generated by the sender of
   the PATH_STATUS frame in the same connection. The receiver of
   the PATH_STATUS frame needs to use and compare the sequence numbers
-  separately for each Sender's Destination Connection ID Sequence
+  separately for each Destination Connection ID Sequence
   Number.
 
 Available values of Path Status field are:
@@ -981,7 +982,7 @@ the same path.
 
 Frames may be received out of order. A peer MUST ignore an incoming
 PATH_STATUS frame if it previously received another PATH_STATUS frame
-for the same Sender's Destination Connection ID Sequence Number with a
+for the same Destination Connection ID Sequence Number with a
 Path Status sequence number equal to or higher than the Path Status
 sequence number of the incoming frame.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -212,9 +204,12 @@ To add a new path to an existing multipath QUIC connection, a client starts a pa
...
 of a path, but it can validate a new path created by a client.
-A new path can only be used once it has been validated. Each endpoint associates a
-Path identifier to each path. This identifier is notably used when a peer sends a PATH_ABANDON frame
-to indicate that it has closed the path whose identifier is contained in the PATH_ABANDON frame.
+A new path can only be used once it has been validated. Each endpoint
+identifies its paths by the sequence number of the Destination
+Connection ID it uses over them. Such numerical identifier is notably
+used when an endpoint sends a PATH_ABANDON frame to request its peer to
+abandon the path on which the sender uses the Destination Connection ID
+with the sequence number contained in the PATH_ABANDON frame.
```
mirjak, line 212:
```suggestion
A new path can only be used once it has been validated. The Destination
Connection ID is used to associate a packet to a valid path. Further, the
sequence number of Destination Connection ID is used as numerical identifier
in control frames. E.g. an endpoint sends a PATH_ABANDON frame to request its peer to
abandon the path on which the sender uses the Destination Connection ID
with the sequence number contained in the PATH_ABANDON frame.
```

mirjak: Maybe we can also move the text about the status frame here.

draft-ietf-quic-multipath.md
```diff
@@ -536,19 +529,18 @@ In non-final states, hosts have to track the following information.
 destination port) used by the endhost to send packets over the path.
 
 - Associated Destination Connection ID: The Connection ID used to send
-packets over the path.
+packets over the path. The endpoint relies on its sequence number to
+maintain a Destination Connection ID-specific packet number space for
+sending packets over this path. Packet number considerations as
+described in {{Section 12.3 of QUIC-TRANSPORT}} apply within the given
+path.
```
mirjak, line 536: I don't think we need this additional text.

You don't need the sequence number to identify the packet space; you can also just use the CID directly. You only need the sequence number really in control frames.

Also what consideration in RFC9000 do you want to point to?

draft-ietf-quic-multipath.md
```diff
@@ -536,19 +529,18 @@ In non-final states, hosts have to track the following information.
...
-identified by the path ID. Packet number considerations as described in
-{{Section 12.3 of QUIC-TRANSPORT}} apply within the given path.
-
-In the "Active" state, hosts MUST also track the following information.
-
 - Associated Source Connection ID: The Connection ID used to receive
-packets over the path.
+packets over the path. The endpoint relies on its sequence number to
+receive and acknowledge packets belonging to that Connection ID-specific
+packet number space.
```
mirjak, line 543:
```suggestion
packets over the path. The endpoint relies on its sequence number to
send oath control information and specifically acknowledge packets belonging to that Connection ID-specific
packet number space.
```

draft-ietf-quic-multipath.md
```diff
@@ -587,26 +579,25 @@ the ACK frame as well as packet protection as described in the following subsect
...
 Acknowledgements of Initial and Handshake packets MUST be carried using
 ACK frames, as specified in {{QUIC-TRANSPORT}}. The ACK frames, as defined
-in {{QUIC-TRANSPORT}}, do not carry path identifiers. If for any reason
-ACK frames are received in 1-RTT packets while the state of multipath
-negotiation is ambiguous, they MUST be interpreted as acknowledging
-packets sent on path 0.
+in {{QUIC-TRANSPORT}}, do not carry the Receiver's Destination
+Connection ID Sequence Number field. If multipath has been successfully
+negotiated, ACK frames in 1-RTT packets acknowledge packets sent with
+the Connection ID having sequence number 0.
```
mirjak, line 591:
```suggestion
Connection ID Sequence Number field. If multipath has been successfully
negotiated, ACK frames in 1-RTT packets MUST be interpreted as acknowledging
packets on the path they the packet with the ACK frames been received.
```

draft-ietf-quic-multipath.md
```diff
@@ -587,26 +579,25 @@ the ACK frame as well as packet protection as described in the following subsect
...
 As soon as the negotiation of multipath support is completed,
 endpoints SHOULD use ACK_MP frames instead of ACK frames to acknowledge application
-data packets, including 0-RTT packets, received on path ID 0 after the handshake concluded.
+data packets, including 0-RTT packets, using the Connection ID with
+sequence number 0 after the handshake concluded.
 
-ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be sent on the same path
-as identified by the Path Identifier. However, an ACK_MP frame can be returned via a
+ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be sent on the path
+it perceives the Connection ID of the packet number space it acknowledges.
```
mirjak, line 599:
```suggestion
ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be sent on same the path
as the data was sent that it acknowledged.
```

draft-ietf-quic-multipath.md
```diff
@@ -966,16 +958,19 @@ PATH_STATUS frames are formatted as shown in {{fig-path-status-format}}.
 
 PATH_STATUS Frames contain the following fields:
 
-Path Identifier:
-: The identifier of the path that the status update correspondes to.
+Sender's Destination Connection ID Sequence Number:
+: The sequence number of the Destination Connection ID used by the
+  sender of this frame to send packets over the path the status update
+  corresponds.
```
mirjak, line 964:
```suggestion
  corresponds to.
```

draft-ietf-quic-multipath.md
huitema: Typo in suggested change: oath -> path

draft-ietf-quic-multipath.md
huitema: No, please do not apply this change. This PR is editorial in nature, and should not change the consensus so far, i.e., the old ACK frame acknowledges CID number 0. Quentin's original text is fine.

draft-ietf-quic-multipath.md
```diff
@@ -1028,16 +1024,16 @@ ACK_MP frame is formatted as shown in {{fig-ack-mp-format}}.
 Compared to the ACK frame specified in {{QUIC-TRANSPORT}}, the following
 field is added.
 
-Path Identifier:
-: The identifier of the path to identify the packet number
+Receiver's Destination Connection ID Sequence Number:
+: The sequence number of the Connection ID identifying the packet number
   space of the 1-RTT packets which are acknowledged by the ACK_MP frame.
```
huitema, line 1030: I think we have an issue of consistency. In the Abandon and Path Status frames, the DCID field is called "Sender's DCID". Here, it is called "Receiver's DCID". Yet, they refer to exactly the same value. My suggestion:

* Remove "Sender's" from the DCID descriptions in PATH_ABANDON and PATH_STATUS
* Remove "Receiver's" from the DCID descriptions in ACK_MP
* Otherwise, keep the text as is, because the text is correct.

draft-ietf-quic-multipath.md
```diff
@@ -1028,16 +1024,16 @@ ACK_MP frame is formatted as shown in {{fig-ack-mp-format}}.
...
+If an endpoint receives an ACK_MP frame with a Connection ID sequence
+number which was never issued (i.e., with a sequence number
 larger than the largest one advertised), it MUST treat this as a connection
 error of type MP_PROTOCOL_VIOLATION and close the connection.
-If an endpoint receives an ACK_MP frame with a Path ID
-which is no more active (e.g., retired by a RETIRE_CONNECTION_ID
+If an endpoint receives an ACK_MP frame with a Connection ID sequence
+number which is no more active (e.g., retired by a RETIRE_CONNECTION_ID
 frame or belonging to closed paths), it MUST ignore the ACK_MP frame
 without causing a connection error.
```
huitema, line 1039: The reference to "closed paths" is a bit ambiguous, because it is in fact OK to send an ACK_MP after and ABANDON_PATH -- or even in the same packet, but QUIC does not specify the order of processing of frames in a packet. And it is also OK to process it, rather than spuriously repeat data that has been actually received. I would be tempted to just strike out "or belonging to closed paths". Or maybe strike out the whole parenthesis. Or maybe strike out the whole paragraph about not active anymore. For comparison, Section 19.16 of RFC 9000 just says:

Receipt of a RETIRE_CONNECTION_ID frame containing a sequence number greater than any previously sent to the peer MUST be treated as a connection error of type PROTOCOL_VIOLATION.

huitema (commented):
A few nits, but this PR does improve the draft!

draft-ietf-quic-multipath.md
mirjak: fixed

draft-ietf-quic-multipath.md
qdeconinck: Indeed, let's remove that added text.

draft-ietf-quic-multipath.md
mirjak: Yes, this changes goes maybe a bit far for this PR but Quentin's text is not right because he removed the "If for any reason  ACK frames are received in 1-RTT packets while the state of multipath negotiation is ambiguous" part. In this ambiguous state all acks are mapped to the seq number 0. However, this is not true later. My proposal is not to change the consensus but rather say the more general thing.

draft-ietf-quic-multipath.md
mirjak: No they do not refer to the same value. The Sender's CID is the one used when you sent a packet. The Receivers's CID is the one used by the other end to send - hence when you receive packets.

draft-ietf-quic-multipath.md
qdeconinck: Yes, let's avoid introducing design changes here. Also, I think relying on the reception 4-tuple may lead to ambiguous state at the receiver.

draft-ietf-quic-multipath.md
qdeconinck: While I understand the suggestion, I'm not sure this makes the text clearer, though I don't have strong opposition here.

draft-ietf-quic-multipath.md
qdeconinck: Actually the PATH_ABANDON and PATH_STATUS frame includes the sequence number of the DCID of the sender of the frame, while in ACK_MP the sequence number corresponds to the DCID used by its peer, hence the (verbose) name convention. But if this introduces confusion, I can remove "Sender's" and "Receiver's" mentions.

draft-ietf-quic-multipath.md
qdeconinck: I think your point about the whole parenthesis is right. Also, this issue about seeing retired DCID also applies in PATH_ABANDON and PATH_STATUS (i.e., just ignore the frame), so we may just want to add a sentence at the beginning of the presentation of the multi path-specific frames?

draft-ietf-quic-multipath.md
qdeconinck: For the ambiguous part, I'm not sure to understand when the negotiation of multi path could be ambiguous. But so far the less ambiguous approach (as I understood so far) is that ACK frames (always) relate to CID number 0.

mirjak: I was assuming an ACK frame always corresponds to the path it was sent on. However, it seem this is simply not defined in the draft and we probably need to open an issue to clarify this.

draft-ietf-quic-multipath.md
mirjak: Actually we did a mistake here when I remove the type field. The path abandon frame should also have the receivers cid as you tell the other end to not use this for sending data anymore. Same I probably true for the path status frame.

draft-ietf-quic-multipath.md
huitema: No, we specifically do not want implementations to continue sending ACK frames after negotiating multipath. The notion of "current path" is ambiguous -- for example, it does change in case of NAT rebinding. Adding the CID-sequence-number in the ACK_MP guarantee that there is no such ambiguity. As for nake ACK, the equivalence to "cid-number=0" is straightforward.

draft-ietf-quic-multipath.md
huitema: I think aligning ACK, ABANDON and STATUS makes sense. ABANDON is in many respects just the symmetric of the ACK. For example, a node receives data on a cellular path that charges per byte when a "free" path is available. The node will send the ABANDON immediately, telling its peer to "stop doing that." Using the DCID "used by the peer to send 1-RTT packets" makes a lot of sense.

draft-ietf-quic-multipath.md
qdeconinck: I fully agree here. Sending the ACK_MP on the same path is a reasonable default behaviour, but in some network configurations, this is clearly suboptimal and there is no point forcing endpoints to do that. I think having a  "SHOULD" is fine, but not a "MUST".

Rethinking further, I'm wondering if this paragraph could even be removed based on discussions in #183. We already discuss the ACK_MP scheduling in implementation considerations, and that should be enough. Maybe add at some place that frames can effectively be sent and received on any valid/active path.

draft-ietf-quic-multipath.md
qdeconinck: Yeah, this would make sense. Let me update the text and the examples accordingly.

draft-ietf-quic-multipath.md
```diff
@@ -212,9 +204,12 @@ To add a new path to an existing multipath QUIC connection, a client starts a pa
 the chosen path, as further described in {{setup}}.
 In this version of the document, a QUIC server does not initiate the creation
 of a path, but it can validate a new path created by a client.
-A new path can only be used once it has been validated. Each endpoint associates a
-Path identifier to each path. This identifier is notably used when a peer sends a PATH_ABANDON frame
-to indicate that it has closed the path whose identifier is contained in the PATH_ABANDON frame.
+A new path can only be used once it has been validated. The Destination
+Connection ID is used to associate a packet to a valid path. Further, the
```
mirjak, line 208:
```suggestion
A new path can only be used once the associated 4-tuple has been validated
by ensuring that the peer is able to receive packets at that address
(see {{Section 8 of RFC9000}}). The Destination Connection ID is used
to associate a packet to a packet number space that is used on a valid path. Further, the
```

draft-ietf-quic-multipath.md
```diff
@@ -378,8 +374,9 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
 
 Both endpoints, namely the client and the server, can close a path,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
-abandons the path with a corresponding Path Identifier. The PATH_ABANDON
-frame contains the Path Identifier and therefore can be sent on any path.
+abandons the path with the corresponding Destination Connection ID.
```
mirjak, line 377:
```suggestion
which request the peer to stop sending packets with the corresponding Destination Connection ID.
```

draft-ietf-quic-multipath.md
```diff
@@ -378,8 +374,9 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
 
 Both endpoints, namely the client and the server, can close a path,
```
mirjak, line 375:
```suggestion
Both endpoints, namely the client and the server, can initiate path closure,
```

draft-ietf-quic-multipath.md
```diff
@@ -176,21 +176,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
...
-  in multipath control frames (e.g., PATH_ABANDON frame) to identify
-  a path. The initial path that is used during the handshake (and multipath negotiation)
-  has the path ID 0 and therefore all 0-RTT packets are also tracked and
-  processed with the path ID 0. For 1-RTT packets, the path ID is the
-  sequence number of the Destination Connection ID present in the packet
-  header, as defined in {{Section 5.1.1 of QUIC-TRANSPORT}} that is used
-  for sending packets on that particular path.
+  "network path" used in {{QUIC-TRANSPORT}}. Endpoints identify a path
+  by the sequence number of the Destination Connection ID they use over
+  that path.
```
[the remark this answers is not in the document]
qdeconinck: While the path actually refers to the (perceived) 4-tuple, not sure we say the path only depends on the CID. The CID (sequence number) is used between endpoints to identify a path in PATH_ABANDON/PATH_STATUS frames. However, we can remove this definition if this brings confusion.

draft-ietf-quic-multipath.md
```diff
@@ -587,26 +575,24 @@ the ACK frame as well as packet protection as described in the following subsect
...
+to distinguish the Connection ID-specific packet number space.
 
 Acknowledgements of Initial and Handshake packets MUST be carried using
 ACK frames, as specified in {{QUIC-TRANSPORT}}. The ACK frames, as defined
-in {{QUIC-TRANSPORT}}, do not carry path identifiers. If for any reason
-ACK frames are received in 1-RTT packets while the state of multipath
-negotiation is ambiguous, they MUST be interpreted as acknowledging
-packets sent on path 0.
+in {{QUIC-TRANSPORT}}, do not carry the Destination Connection ID
+Sequence Number field. If multipath has been successfully
```
mirjak, line 584:
```suggestion
Sequence Number field to identify the packet number space.
If the multipath extension has been successfully
```

draft-ietf-quic-multipath.md
mirjak:
```suggestion
ACK_MP frame (defined in {{ack-mp-frame}}) SHOULD be sent on the path
it received packet with the Connection ID of the packet number space it acknowledges.
```

mirjak: On the editorial part: I think "perceives" is just the wrong word. I was also trying to make it shorter but feel free to take any of the two suggestions.

On the text itself, if we use normative language it should stay here and not move to the implementation guidance section. let keep this for now and maybe open a separate issue in case we need more discussion or actually want to change it.

My view is: I think we need some normative guidance because of the RTT calculation issue. However, there might be room for refining it or provide further explanation in the implementation guidance scheduling section.

draft-ietf-quic-multipath.md
mirjak: I see that this text was remove, which is fine. But was it re-added somewhere else?

draft-ietf-quic-multipath.md
huitema: "receive packet with" is a better text than "perceived".

And yes, let's merge this PR as is, and discuss path affinity of MP_ACK separately.

draft-ietf-quic-multipath.md
```diff
@@ -376,10 +374,11 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
 
 ### Use PATH_ABANDON Frame to Close a Path {#path-abandon-close}
 
-Both endpoints, namely the client and the server, can close a path,
+Both endpoints, namely the client and the server, can initiate path closure,
 by sending PATH_ABANDON frame (see {{path-abandon-frame}}) which
```
mirjak, line 378:
```suggestion
by sending a PATH_ABANDON frame (see {{path-abandon-frame}}) which
```

draft-ietf-quic-multipath.md
```diff
@@ -376,10 +374,11 @@ Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
...
-abandons the path with a corresponding Path Identifier. The PATH_ABANDON
-frame contains the Path Identifier and therefore can be sent on any path.
+which requests the peer to stop sending packets with the corresponding Destination Connection ID.
+The PATH_ABANDON frame contains the Destination Connection
+ID Sequence Number and therefore can be sent on any path.
```
mirjak, line 381: This is entirely my fault with PR #174 but this is actually specified already later in the text and therefore redundant here. However, we can also just leave this for now and make another editorial pass later.

draft-ietf-quic-multipath.md
huitema: I started a discussion of the "ACK on same path" in issue #190

draft-ietf-quic-multipath.md
qdeconinck: Not yet. We may add a sentence to clarify that we should just ignore Multipath frames related to retired DCIDs, but I would suggest to add such a sentence at the beginning of the section introducing these frames, as this applies to all of them.

mirjak: Okay, maybe let's open an issue, so we don't forget!

draft-ietf-quic-multipath.md
mirjak: Okay, should we open an (editorial) issue, so we don't forget? Or we can create a new PR right after this one is merged (without an issue).

mirjak:
I think this PR is ready to merge. Let's wait until end of today to give others another change to review. However would be good to merge to, as we need to merge this one before we can merge the others!

draft-ietf-quic-multipath.md
qdeconinck: Filled #193.

draft-ietf-quic-multipath.md
qdeconinck: Filled #194.

Yanmei-Liu (approved):
Looks good to me.

PR #189, opened 2023-03-08
Add descriptions on resetting packet number  when a new CID is used

yfmascgy:
A new sentence is added to clarify that when a new CID is used, packet number space's largest_ack should be reset to None for the packet number encoding algorithm specified in RFC9000.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -582,6 +582,13 @@ This also means that the same packet number can occur on each path and the
 packet number is not a unique identifier anymore. This requires changes to
 the ACK frame as well as packet protection as described in the following subsections.
 
+When multipath is negotiated,
+each destination connection ID is linked to a separate packet number space.
+When an existing path switches to a new CID or a new CID is used to open a new path,
+the largest packet number (largest_acked) that has been acknowledged by the
+peer in this new CID's packet number space SHOULD be reset to "None"
+for packet number encodings, as specified in ({{Section A.2 of QUIC-TRANSPORT}}).
+
 ## Sending Acknowledgements
 
 The ACK_MP frame, as specified in {{ack-mp-frame}}, is used to
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -584,10 +584,10 @@ the ACK frame as well as packet protection as described in the following subsect
 
 When multipath is negotiated,
 each Destination Connection ID is linked to a separate packet number space.
-When an existing path switches to a new CID or a new CID is used to open a new path,
+Each CID-specific packet number space starts at packet number 0. When following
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -582,6 +582,13 @@ This also means that the same packet number can occur on each path and the
 packet number is not a unique identifier anymore. This requires changes to
 the ACK frame as well as packet protection as described in the following subsections.
 
+When multipath is negotiated,
+each destination connection ID is linked to a separate packet number space.
```
qdeconinck, line 586:
```suggestion
each Destination Connection ID is linked to a separate packet number space.
```

draft-ietf-quic-multipath.md
```diff
@@ -582,6 +582,13 @@ This also means that the same packet number can occur on each path and the
...
+When an existing path switches to a new CID or a new CID is used to open a new path,
+the largest packet number (largest_acked) that has been acknowledged by the
+peer in this new CID's packet number space SHOULD be reset to "None"
+for packet number encodings, as specified in ({{Section A.2 of QUIC-TRANSPORT}}).
```
qdeconinck, line 590: I think discussing the notion of path here adds confusion. Packet number spaces are directly related to the CIDs. Maybe this would be clearer?
```suggestion
Each CID-specific packet number space starts at packet number 0. When following
the packet number encoding algorithm described in {{Section A.2 of QUIC-TRANSPORT}},
the largest packet number (largest_acked) that has been acknowledged by the
peer in this new CID's packet number space is initially set to "None".
```

qdeconinck (commented):
I think we should not discuss the notion of path here.

draft-ietf-quic-multipath.md
huitema: +1 @qdeconinck . The word "reset" in the original PR suggests that the number space is "per path". It is not. It is per CID.

huitema (commented):
Please use Quentin's text!

draft-ietf-quic-multipath.md
mirjak: Yes, "reset" would be problematic as you still need to keep state also for any packet that you may receive with the old CID.

How does changing the CID on the same path actually impact loss detection? Do we also need to say something about that?

qdeconinck: @mirjak For the changing CID, I suggest to adapt the text in #172 (basically, if you change the CID on a path, you can keep your path's congestion control state and RTT measurements).

yfmascgy: Yeah, my bad. The word "reset" has problems. I will adopt Quentin's suggestions.

mirjak (approved):

PR #191, opened 2023-03-09
Add retire_cid in path status figure in section 4.4

yfmascgy:


Yanmei-Liu (approved):

mirjak (approved):

mirjak:
I believe this needs updating again, as we now say in the text that you wait for 3 PTOs after the Abandon frame before you send the retire_CID frame. Does that mean we need an additional state here?

PR #195, opened 2023-03-10
Make DCID sequence number considerations common to all multipath-specific frames

### LINKED ISSUE #137: How should a receiver process a PATH_ABANDON having an unknown Path Identifier?

### LINKED ISSUE #193: Clarify endpoints should ignore multipath frames referencing retired DCID sequence number
We may add a sentence to clarify that we should just ignore Multipath frames related to retired DCIDs, but I would suggest to add such a sentence at the beginning of the section introducing these frames, as this applies to all of them.

qdeconinck:
More specifically, if the receiver cannot process an old DCID seq number, it MUST ignore it (i.e., no error).

Fix #193. Also fix #137.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -964,6 +964,12 @@ If an endpoint receives multipath-specific frames from packets of
 other encryption levels, it MUST return MP_PROTOCOL_VIOLATION
 as a connection error and close the connection.
 
+All these multipath-specific frames relate to a Destination Connection
+ID sequence number. If an endpoint receives a multipath-specific frame
+with a Destination Connection ID sequence number that it cannot process
+anymore (e.g., because the Connection ID might have been retired), it
+MUST silently ignore the frame.
+
 ## PATH_ABANDON Frame {#path-abandon-frame}
 
 The PATH_ABANDON frame informs the peer to abandon a
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -964,7 +964,7 @@ If an endpoint receives multipath-specific frames from packets of
 other encryption levels, it MUST return MP_PROTOCOL_VIOLATION
 as a connection error and close the connection.
 
-All these multipath-specific frames relate to a Destination Connection
+All multipath-specific frames relate to a Destination Connection
 ID sequence number. If an endpoint receives a multipath-specific frame
 with a Destination Connection ID sequence number that it cannot process
 anymore (e.g., because the Connection ID might have been retired), it
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -965,7 +965,10 @@ other encryption levels, it MUST return MP_PROTOCOL_VIOLATION
 as a connection error and close the connection.
 
 All multipath-specific frames relate to a Destination Connection
-ID sequence number. If an endpoint receives a multipath-specific frame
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -964,6 +964,12 @@ If an endpoint receives multipath-specific frames from packets of
 other encryption levels, it MUST return MP_PROTOCOL_VIOLATION
 as a connection error and close the connection.
 
+All these multipath-specific frames relate to a Destination Connection
```
mirjak, line 967:
```suggestion
All multipath-specific frames relate to a Destination Connection
```

draft-ietf-quic-multipath.md
```diff
@@ -964,6 +964,12 @@ If an endpoint receives multipath-specific frames from packets of
...
+ID sequence number. If an endpoint receives a multipath-specific frame
+with a Destination Connection ID sequence number that it cannot process
+anymore (e.g., because the Connection ID might have been retired), it
+MUST silently ignore the frame.
```
mirjak, line 971: I think we should cover two cases here. If the seq number is smaller than the largest issued, you should ignore. If it's larger you have a connection error...?

qdeconinck:
Updated the PR to also include the case of a DCID sequence number greater than any sent (connection error).

Yanmei-Liu (approved):
Look good to me

PR #198, opened 2023-03-10
Send Abandon frame after NAT rebinding

mirjak:
This is the proposed text more the other PR to start working this. The proposed solution in further discussion was to also extend the path abandon path with a specific reason code of this case.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1012,6 +1012,11 @@ This leads to an inconsistent view of open paths at both peers,
 however, as the "old" path will not work anymore, it will be silently
 closed after the idle timeout expires.
 
+If an endpoint receives a PATH_CHALLENGE frame with a new CID on a existing path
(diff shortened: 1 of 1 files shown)


mirjak:
See issue #188

huitema (changes requested):
That PR is premature. We do not yet have consensus on how to solve issue #188, and we should keep the discussion there.

mirjak:
Yes, this is not ready to merge and discuss should take place in the issue (#188). I just wanted to capture this text as a starting point before we lose it.

huitema:
I think we need clarity on PR #292 before we progress this draft. We have an issue there, because the "Natted" path and the "before-NAT" path use the same number space. With PR #292, that gives us a strong constraint. I think I explained a plausible solution in https://github.com/quicwg/multipath/pull/292#discussion_r1493966132, but with that solution if rebinding is accepted the new (natted) path will have the same path_id as before NAT rebinding.

mirjak:
OBE after merge of #292

PR #199, opened 2023-03-13
editorial pass on frames section

### LINKED ISSUE #196: Make an editorial pass in the "New Frames" section
Just to not forget this, but this is mostly editorial.

- Probably present first the ACK_MP frame, and then PATH_ABANDON and PATH_STATUS
- Drop reference to I-D.liu-multipath-quic as the PATH_STATUS frame is now integrated in the draft.
- Based on #183 and #190, and If we agree on #192, we could also drop the text in PATh_ABANDON stating that

> PATH_ABANDON frames MAY be sent
on any path, not only the path on which the referenced Destination
Connection ID is used.

qdeconinck:
Fix #196.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1018,12 +1018,41 @@ with a Destination Connection ID sequence number that it cannot process
 anymore (e.g., because the Connection ID might have been retired), it
 MUST silently ignore the frame.
 
+## ACK_MP Frame {#ack-mp-frame}
+
+The ACK_MP frame (types TBD-00 and TBD-01; experiments use 0xbaba00..0xbaba01)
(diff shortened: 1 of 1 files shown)


qdeconinck:
The ACK_MP text does not change, it just comes before PATH_ABANDON and PATH_STATUS. The rationale is that 1) the new order follows the encoding numbering (lowest to highest), and 2) the ACK_MP is a critical frame, the PATH_ABANDON is an important one yet still "indicative" (it suggests the peer to abandon this path) and the PATH_STATUS frame is an indicative one.

Yanmei-Liu (approved):
OK. It's reasonable.

mirjak (approved):

PR #217, opened 2023-04-06
Proposed text for RTT computation and ACK_MP scheduling

### LINKED ISSUE #132: Can we be more specific about RTT computation? The word "statistical" is a bit broad.
The current draft says "If timestamps are not available, implementations could estimate one way delays using statistical techniques". I find the word "statistical" a bit too broad. Can we elaborate what is the exact statistical method recommended to use? Moreover, "statistical" also implies variation and randomness, what happens for applications that do not expect such randomness?

huitema:
fixes #132 

This proposed text updates the discussion of RTT computation and ACK_MP scheduling in the Example section, which is not normative. The proposed text reflects the discussion during the working group meeting in Yokohama, and aligns with the apparent consensus that emerged during the discussion.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -915,7 +915,7 @@ the new values. There is however an exception: some congestion
 control functions rely on estimates of the minimum RTT. It might be prudent
 for nodes to remember the path over which the ACK MP that produced
 the minimum RTT was received, and to restart the minimum RTT computation
-if that path is abandoned. 
+if that path is abandoned.
 
 ## Packet Scheduling
 
@@ -938,7 +938,7 @@ the standard algorithm reflect both the characteristics of the
 path and the scheduling algorithm of ACK_MP frames. The estimates will converge
 faster if the scheduling strategy is stable, but besides that
 implementations can choose between multiple strategies such as sending
-ACK_MP frames on the path they acknowledge packets, or sending 
+ACK_MP frames on the path they acknowledge packets, or sending
 ACK_MP frames on the shortest path, which results in shorter control loops
 and thus better performance.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -907,7 +907,7 @@ for example the PTO timeout: if an ACK_MP is not received after more
 than 350ms, either the data packet or its ACK_MP were probably lost.
 
 The simplest implementation is to compute smoothedRTT and RTTvar per
-{{section 5.3 of QUIC-RECOVERY}} regardless of the path through which MP_ACKs are
+{{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which MP_ACKs are
 received. This algorithm will provide good results,
 except if the set of paths changes and the ACK_MP sender
 revisits its sending preferences. This is not very
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -889,31 +889,33 @@ Terrestrial | 100ms  | 350ms
 Satellite   | 350ms  | 600ms
 {: #fig-example-ack-delay title="Example of ACK delays using multiple paths"}
 
-Using the default algorithm specified in {{QUIC-RECOVERY}} would result
-in suboptimal performance, computing average RTT and standard
-deviation from series of different delay measurements of different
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -889,31 +889,33 @@ Terrestrial | 100ms  | 350ms
...
+
+The computed values reflect both the state of the network path and the
+scheduling decisions by the sender of the ACK_MP frames. In the example
+above, we may assume that the ACK_MP will be sent over the terrestrial
+link, because that provides the best response time. In that case, the
+computed RTT value for the satellite path will be about 350ms. This
+lower than the 600ms that would be measured if the ACK_MP came over
+the satellite channel, but it is still the right value for computing
+for example the PTO timeout: if an ACK_MP is not received after more
+than 350ms, either the data packet or its ACK_MP were probably lost.
```
mirjak, line 907: Isn't the PTO 3xRTT?

mirjak: However, I'm not sure about this. In this examples you assume that the terrestrial path will used in a stable way for ACK. In the paragraph above you say if different path are used which is a different case. So if I send half my ACK over each pass, I will see an average RTT of 475ms which is too low for those ACKs that are sent over the satellite path.

draft-ietf-quic-multipath.md
```diff
@@ -889,31 +889,33 @@ Terrestrial | 100ms  | 350ms
...
+
+In general, using the algorithm above will provide good results,
```
mirjak, line 909: Which algorithm exactly?

draft-ietf-quic-multipath.md
huitema: Commenting on this PR after the merge: in the example that you cite, the smoothed RTT will actually be very close to the shortest path RTT + maybe 1 ACK delay. ACK_MP frames are redondant. If you send ACK alternately on short and long path, the ACK on the short path will arrive before the one on the long path. When the ACK on the long path arrives, its "highest" packet will already have been acked as part of a range carried on the short path, so it will not contribute to RTT computation.

mirjak: Only if you send ACKs multiple time per RTT and the latency difference between the path is large.

So maybe we should rather not make any default recommendations but just explain the problems and non-problems in different alternatives...?

(btw. which merge are you talking about? This PR is not merged yet)

huitema: From RFC 9002:
```
PTO = smoothed_rtt + max(4*rttvar, kGranularity) + max_ack_delay
```

draft-ietf-quic-multipath.md
huitema: Should probably say "computing smoothedRTT and RTTvar per section 5.3 of RFC9002 regardless of the path on which MP_ACK are received". But that's a handful, so maybe:

"The simplest implementation is to compute `smoothedRTT` and `RTTvar` per section 5.3 of RFC9002 regardless of the path on which MP_ACKs are received. This algorithm will provide good results,"

draft-ietf-quic-multipath.md
```diff
@@ -931,9 +933,14 @@ implementation.
 
 Note that this implies that an endpoint may send and receive ACK_MP
 frames on a path different from the one that carried the acknowledged
-packets. A reasonable default consists in sending ACK_MP frames on the
-path they acknowledge packets, but the receiver must not assume its
-peer will do so.
+packets. As noted in {{compute-rtt}} the values computed using
+the standard algorithm reflect both the characteristics of the
+path and the scheduling algorithm of ACK_MP frames. The estimates will converge
+faster if the scheduling strategy is stable, but besides that
```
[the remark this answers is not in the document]
huitema: I think "convergence" generally means that the algorithm has produced results that reflect the current network state, and that the smoothed RTT and RTTvar values will reflect that.

draft-ietf-quic-multipath.md
huitema: Latest commit fixes that, hopefully.

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
huitema: Just pushed a fix in the latest commit.

Yanmei-Liu (approved):

PR #250, opened 2023-07-05
Add Chinese names

### LINKED ISSUE #201: Use authors actual names
Now, I might be making an assumption, but I believe that some of the authors/editors on the draft are currently using latin transliterations of their names.  We have the ability to use real names, can we do that?

(Note: we need to ask; I have had some people express a desire to only use the latin version of their name.  The most important thing is to respect individual wishes.)

mirjak:
fixes #201

This is what MT proposed. When I build this locally, it doesn't change the title page but it has this in the author information section

```
   Yanmei Liu (editor)
   Alibaba Inc.
   Email: miaoji.lym@alibaba-inc.com

   Additional contact information:

      马云飞 (editor)
      Alibaba Inc.
```

This is the right way to do this?

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -16,7 +16,9 @@ pi: [toc, sortrefs, symrefs]
 author:
 -
   ins: Y. Liu
-  name: Yanmei Liu
+  fullname:
+      :: "马云飞"
+      ascii: "Yanmei Liu"
   role: editor
   org: Alibaba Inc.
   email: miaoji.lym@alibaba-inc.com


draft-ietf-quic-multipath.md
```diff
@@ -16,7 +16,9 @@ pi: [toc, sortrefs, symrefs]
 author:
 -
   ins: Y. Liu
```
LPardue, line 18:
```suggestion
```

If you remove this line, the author names at the top of the document will properly render

draft-ietf-quic-multipath.md
```diff
@@ -15,14 +15,16 @@ pi: [toc, sortrefs, symrefs]
 
 author:
 -
-  ins: Y. Liu
-  name: Yanmei Liu
+  fullname:
+      :: :: "刘彦梅"
```
mirjak, line 19:
```suggestion
      :: "刘彦梅"
```

Yanmei-Liu (approved):

martinthomson (commented):
There is a bug on line 48 still, but I'm glad to see this happen.  (As a rule, "fullname" is the only thing to use, but Quentin probably needs to specify some so that his surname doesn't get busted up by xml2rfc.)

PR #263, opened 2023-07-10
ACK_MP: fix typo about experiments number

alagoutte:
remove last 0xbaba experimental number

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1018,7 +1018,7 @@ MUST silently ignore the frame.
 
 ## ACK_MP Frame {#ack-mp-frame}
 
-The ACK_MP frame (types TBD-00 and TBD-01; experiments use 0x15228c00-0x15228c01)
+The ACK_MP frame (types TBD-00 and TBD-01)
 is an extension of the ACK frame defined by {{QUIC-TRANSPORT}}. It is
 used to acknowledge packets that were sent on different paths using
 multiple packet number spaces. If the frame type is TBD-01, ACK_MP frames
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1018,7 +1018,7 @@ MUST silently ignore the frame.
 
 ## ACK_MP Frame {#ack-mp-frame}
 
-The ACK_MP frame (types TBD-00 and TBD-01; experiments use 0xbaba00..0xbaba01)
+The ACK_MP frame (types TBD-00 and TBD-01; experiments use 0x15228c00-0x15228c01)
```
marten-seemann, line 1021: It might make sense to only have a single point in the document that defines the code point.

huitema: @marten-seemann Yes. Case in point: I thought I had fixed that everywhere, PR #254 was reviewed and approved 4 times, yet we missed that. Fix would be to replace all references to experimental types by a pointer to the IANA section. But we missed the IETF deadline for that.

huitema (approved):
Thanks for fixing that, @alagoutte. As @marten-seemann suggests, we could do something more radical, but merging that PR would at least fix the editor copy.

draft-ietf-quic-multipath.md
alagoutte: need to update the commit for remove code point ?

qdeconinck: Let's be even more radical for now.
```suggestion
The ACK_MP frame (types TBD-00 and TBD-01)
```

qdeconinck (approved):

PR #264, opened 2023-07-19
Fix typo and enable multipath between only one client and only one server IP address

muxamilian:


Yanmei-Liu (approved):
I agree that there are scenarios which need to use multi-path with same IP address and different ports. BTW, this also help implementations test with each other when running on most devices without two or more IP addresses.

mirjak:
Yes, it is true that multiple port number can be used and we should not that. However, the chance proposed some editorially not correct as the list is only focused on IP addresses.

muxamilian:
My idea was that both the client and the server can only use one IP address but multiple ports. Then it's still multi-path. In what way do you mean my suggestion is not about IP addresses? I'm not sure I understood.

mirjak:
The text before the list says

```
Each endhost may use several IP addresses to serve the connection. In
particular, the multipath extension supports the following scenarios.
```

So I think the point that you added (is correct) but doesn't find editorially in that list. We could simply add this as  sentence after the list or make a slightly larger editorial change (to this sentence above) to make sure that is clear.

muxamilian:
@mirjak Tried to rephrase it according to your suggestion.

draft-ietf-quic-multipath.md
```diff
@@ -153,6 +153,9 @@ particular, the multipath extension supports the following scenarios.
     several ones.
   * The client uses multiple IP addresses and the server listens on
     several ones.
+    
+The client can use one or several ports per IP address and the
+server can listen on one ore several ports per IP address.
```
qdeconinck, line 158:
```suggestion
    * The client uses only one IP address and the server
      listens on only one.

Note that in the last scenario, it still remains possible to have
multiple paths over the connection, given that a path is not only
defined by the IP addresses being used, but also the port numbers.
In particular, the client can use one or several ports per IP
address and the server can listen on one or several ports per IP
address.
```

qdeconinck (changes requested):
In my opinion, explicitly adding a fourth scenario and explaining why multiple paths can be present there makes the idea clearer in my humble opinion.

draft-ietf-quic-multipath.md
muxamilian: As additional context, which I forgot earlier, having multipath from the same IP but different ports is useful for congestion control algorithms that make use of fair queuing: https://github.com/muxamilian/fair-queuing-aware-congestion-control

PR #287, opened 2023-11-03
send PATH_ABANDON after time-out on another path

### LINKED ISSUE #210: path-specific idle timeout logic
We introduced the idle timeout in RFC 9000 to enable endpoints to close idle connections without sending any packets (and potentially powering up the radio for that). Idle timeouts are inherently racy: the start time will vary slightly between both endpoints (by about half an RTT), and a packet sent just before expiration of the idle timeout period might be received just after the expiration.
This is ok because running into the idle timeout will completely free up all resources associated with that connection, and stray packets will elicit stateless resets, allowing the other node to do the same.

I’m not sure I fully understand how this applies to paths. It seems like there needs to be a clear signal that a path has been abandoned due to idleness. Sending this signal will be very cheap if it’s bundled with frames that would be sent anyway on a different path.

That said, I’m wondering if the idle timeout is really the right concept for paths to begin with. Peers usually apply some kind of keep-alive logic to keep a path alive, and that keep-alive period is based on the (presumed) characteristics of the path, such as the (presumed) timeouts applied by NATs, as well as the cost of sending a packet on that particular path. Would it make sense to introduce the concept of a path timeout, which could be attached to the PATH_ frames?

mirjak:
fixes #210

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -441,6 +441,8 @@ any packet on this path anymore.
 PATH_ABANDON frames can be sent on any path,
 not only the path that is intended to be closed. Thus, a path can
 be abandoned even if connectivity on that path is already broken.
+Respectively, if there is still an active path, it is RECOMMENDED to
+send an PATH_ABANDON frame after an idle time on another path.
 
 If a PATH_ABANDON frame is received for the only active path of a QUIC
 connection, the receiving peer SHOULD send a CONNECTION_CLOSE frame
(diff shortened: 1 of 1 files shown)


huitema (approved):

draft-ietf-quic-multipath.md
```diff
@@ -441,6 +441,8 @@ any packet on this path anymore.
 PATH_ABANDON frames can be sent on any path,
 not only the path that is intended to be closed. Thus, a path can
 be abandoned even if connectivity on that path is already broken.
+Respectively, if there is still an active path, it is RECOMMENDED to
+send an PATH_ABANDON frame after an idle time on another path.
```
qdeconinck, line 445:
```suggestion
send a PATH_ABANDON frame after an idle time on another path.
```

qdeconinck (approved):
With minor nit.

PR #331, opened 2024-04-17
Use even numbers for path ID

### LINKED ISSUE #328: Designing odd/even path-id, or not.
During the processing of PR #272, we discussed separating the path identifiers in two classes: client initiated path identifiers, with even numbers, and server initiated path identifiers, with odd numbers.

Separating the identifiers in two classes allows clients and servers to both initiate paths, without risking identifier collisions. The current single-class design only works because only the client initiates paths. Otherwise, we would have the risk of both client and servers picking the same path identifiers, and simultaneously sending "path challenges" with the same identifier, but using different IP addresses and UDP ports. This would create confusion.

The design itself is fairly simple: both endpoints advertise odd and even path identified connection identifiers. When the client creates a path, the client picks an available even numbered identifier advertised by the server; when the server creates a path, it picks an odd numbered identifier. When receiving a packet for a new path, each endpoint verifies that the new number is authorized. For example, if the client receives from the server a message on a new path, it verifies that the path identifier of the connection ID chosen by the server is odd. If not, it MAY signal an MP protocol violation. When responding, the endpoint picks a connection ID with the same path identifier as that of the incoming packet.

That much is clear, but then we have to look at the impact of the odd/even design on the number of paths in the `initial_max_paths` transport parameter, and in the `MAX_PATHS` frame. If we follow the design of the max streams design in RFC 9000, we should replace the `initial_max_paths` parameters by two separate parameters:

* initial_max_paths_client: maximum number of path that the client may inititiate,
* initial_max_paths_server: maximum number of path that the server may inititiate.

For each of those, the maximum would be the minimum of the values advertised by client and server. Instead of writing that the multipath option is not used if the `initial_max_paths` parameter is not present, we would have something a bit more complicated, i.e., stating that the frames are only used if at least one of the parameters is present.

Then we would also have to split the `MAX_PATHS` frame into `MAX_PATHS_CLIENT` and `MAX_PATHS_SERVER`.

All that is doable, but I wonder whether this complexity is justified. Maybe what we need instead is just a way to resolve the "path identifier collision". For example, assuming just on path identifier space, we could have a "client win" rule: if the client creates path N with a specific 4 tuple and then receives a path message with identifier N from the server using a different 4 tuple, it just ignores it; if the server sends a path challenge four path N with a four tuple, and then receives a path message from the client with a different 4 tuple, it consider its initial challenge as lost and only responds to the client's message.

So, I have a question: bite the complexity of 2 parallel path identifier sequences; or, simply add a "collisions" rule. What do people think?

initial_max_stream_data_bidi_local
initial_max_stream_data_bidi_remote

huitema:
This PR specifies using even number path identifiers. It anticipates the possibility of using the odd numbered identifiers for server-initiated paths, but does not specify how to do that. As a result, it is much simpler than PR #329.

Close #328

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -336,7 +336,7 @@ that future extensions will lift the requirement in {{path-initiation}} that
 paths are only initiated by clients. In that case, to avoid collisions,
 we will need separate path IDs number spaces for client-initiated and
 server-initiated paths, such as using even numbered path IDs for client-initiated paths
-and odd numbered path IDs for server initiated paths.
+and odd numbered path IDs for server-initiated paths.
 
 
 # Path Setup and Removal {#setup}
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1402,7 +1402,7 @@ last MAX_PATHS frame received from the peer, or the value received
 from the peer in
 the initial_max_paths transport parameters if no MAX_PATHS frames have
 been received. MP_NEW_CONNECTION_ID received as
-with a laue larger than authorized MUST be treated as
+with Path ID value larger than authorized MUST be treated as
 a connection error of type MP_PROTOCOL_VIOLATION.
 
 The Sequence Number field and Retire Prior To field is allocated
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -335,7 +335,7 @@ This specification requires path IDs to be even numbers. We anticipate
 that future extensions will lift the requirement in {{path-initiation}} that
 paths are only initiated by clients. In that case, to avoid collisions,
 we will need separate path IDs number spaces for client-initiated and
-server initiated spaces, such as using even numbered path IDs for client initiated paths
+server-initiated paths, such as using even numbered path IDs for client-initiated paths
 and odd numbered path IDs for server initiated paths.
 
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1484,7 +1484,7 @@ For example, if an endpoint received a Maximum Paths value of 2,
 it may send MP_NEW_CONNECTION_ID with Path ID 0 and 2, but not higher.
 
 The Maximum Paths value cannot exceed 2^31-1, as it is not
-possible to encode Path IDs larger than 2^32-2. Receipt of a 
+possible to encode Path IDs larger than 2^32-2. Receipt of a
 Path Identifier larger than this limit MUST be treated
 as a connection error of type FRAME_ENCODING_ERROR.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -294,7 +294,7 @@ TRANSPORT_PARAMETER error.
 
 # Path Identifier {#pathid}
 
-The explicit Path Identifier is an integer between 0 and 2^32 - 1 (inclusive).
+The explicit Path Identifier is an even integer between 0 and 2^32 - 2 (inclusive).
 The Path Identifier is pre-allocated when endpoints provide new Connection IDs
 with MP_NEW_CONNECTION_ID frames {{mp-new-conn-id-frame}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -329,6 +329,15 @@ Endpoints use PATH_ABANDON frame to inform the peer of the retirement of associa
 Path Identifier. When there is not enough unused Path Identifiers, endpoints SHOULD
 send MAX_PATHS frame to inform the peer that new Path Identifiers are available.
 
+## Using even numbers for path ID
+
+This specification requires path IDs to be even numbers. We anticipate
+that future extensions will lift the requirement in {{path-initiation}} that
+paths are only initiated by clients. In that case, to avoid collisions,
+we will need separate path IDs number spaces for client-initiated and
+server initiated spaces, such as using even numbered path IDs for client initiated paths
```
Yanmei-Liu, line 338:
```suggestion
server-initiated paths, such as using even numbered path IDs for client-initiated paths
```

draft-ietf-quic-multipath.md
```diff
@@ -329,6 +329,15 @@ Endpoints use PATH_ABANDON frame to inform the peer of the retirement of associa
...
+and odd numbered path IDs for server initiated paths.
```
Yanmei-Liu, line 339:
```suggestion
and odd numbered path IDs for server-initiated paths.
```

draft-ietf-quic-multipath.md
```diff
@@ -1383,6 +1392,19 @@ Stateless Reset Token:
...
+The Path Identifier MUST be an even number. MP_NEW_CONNECTION_ID received as
+with an odd numbered path identifier MUST be treated as
+a connection error of type MP_PROTOCOL_VIOLATION.
+
+The Path Identifier MUST NOT be larger than the value authorized in the
+last MAX_PATHS frame received from the peer, or the value received
+from the peer in
+the initial_max_paths transport parameters if no MAX_PATHS frames have
+been received. MP_NEW_CONNECTION_ID received as
+with a laue larger than authorized MUST be treated as
```
Yanmei-Liu, line 1405:
```suggestion
with Path ID value larger than authorized MUST be treated as
```

draft-ietf-quic-multipath.md
```diff
@@ -1383,6 +1392,19 @@ Stateless Reset Token:
...
+
+The Path Identifier MUST be an even number. MP_NEW_CONNECTION_ID received as
+with an odd numbered path identifier MUST be treated as
+a connection error of type MP_PROTOCOL_VIOLATION.
+
+The Path Identifier MUST NOT be larger than the value authorized in the
+last MAX_PATHS frame received from the peer, or the value received
+from the peer in
+the initial_max_paths transport parameters if no MAX_PATHS frames have
+been received. MP_NEW_CONNECTION_ID received as
```
Yanmei-Liu, line 1404:
```suggestion
been received. MP_NEW_CONNECTION_ID received
```

Yanmei-Liu (commented):
Thanks a lot for this new PR!
I totally support it, try to help fix a few typo issue.

huitema:
Thanks for the fixes, @Yanmei-Liu

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -1452,21 +1474,29 @@ MAX_PATHS Frame {
...
 : A count of the cumulative number of path that can be opened
-over the lifetime of the connection. This value cannot exceed 2^32-1, as it is not
-possible to encode Path IDs larger than 2^32-1. Receipt of a frame that permits
-opening of a path with Path Identifier larger than this limit MUST be treated
+over the lifetime of the connection.
+
+The endpoint receiving this frame MAY send MP_NEW_CONNECTION_ID frames
+with Path IDs strictly lower than twice the Maximum Paths.
+For example, if an endpoint received a Maximum Paths value of 2,
+it may send MP_NEW_CONNECTION_ID with Path ID 0 and 2, but not higher.
```
qdeconinck, line 1484: A small suggestion to make this clearer (and avoid the potential reader ambiguity of the "2" value)
```suggestion
For example, if an endpoint received a Maximum Paths value of 3,
it may send MP_NEW_CONNECTION_ID with Path ID 0, 2 and 4, but not higher.
```

qdeconinck (approved):
So I suppose then that receiving a MAX_PATHS with a Maximum Paths value of 0 leads to a connection error. A bit too bad, but I can live with that.

A small suggestion to make the example clearer, otherwise this looks good to me.

draft-ietf-quic-multipath.md
huitema: Yes, good idea.

mirjak:
Do we generally have consensus for the even/odd split? Or do we need to confirm this on the list/in the interim before we merge?

As I just wrote in PR #329 I don't think we fully have consensus yet if we want to specify the server-initiated paths part in this doc or separately. However, I think think we can already merge this PR and keep discussing in PR #329.

huitema:
Closing this PR after the 6/4/2024 interim.

Issue #317, opened 2024-03-17
explicit path IDs: Should server preferred address have its own path ID?


marten-seemann:
It's kind of weird that for the migration to the server preferred address path, the path ID is defined as 0. It might be advantageous to keep the anycast and the unicast path open for a limited amount of time.

We could consider defining this path to have path ID 1?

huitema:
+1. I was under the impression that the preferred path was path-id=1 from the start.

huitema:
Let's tackle that in conjunction with the server/client odd/even split. With that split, we have two options:

1) Stay strictly compatible with 9000. The client performs a path migration, very similar in nature to a NAT rebinding, and the path ID remains 0.

2) Create a specific path for the "preferred" address, in which case the client initiates a new path to that address. Since this is a client initiated new path, the path ID should be an even number, probably 2.

mirjak:
I think this is a migration event as also in RFC9000. What would be the benefit of using a new path ID? When would it be useful to have the anycast and unicast address in use at the same time?

mirjak:
Double-checked that this is clearly specified with path ID 0 as discussed here:

```
Also, the Path ID for the connection ID specified in the "preferred address" transport parameter is 0. Use of the "preferred address" is considered as a migration event that does not change the Path ID.
```

Closing now.

PR #334, opened 2024-05-14
Editorial pass on explicit Path ID

mirjak:
I mainly merged the "Path identifier" section into the right place in the rest of the document. Having such a separate section did repeat the same things multiple time. That is usually very unfortunate because that means you have to change things in multiple places.

Further, some smaller editorial issue in the rest of the document.

This is only a first pass, I think we need more iterations.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -132,7 +132,7 @@ integer between 0 and 2^32 - 1 (inclusive). Path identifies are generated
 monotonically increasing and cannot be reused.
 
 The Path ID is used to
-address a path the new multipath control frames,
+address a path in the new multipath control frames,
 such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
 PATH_AVAILABLE {{path-available-frame}} as well as ACK_MP {{ack-mp-frame}}.
 Further, Connection IDs are issued per Path ID.
@@ -140,23 +140,23 @@ Each connection ID is associated with one path identifier
 but multiple connection IDs can be associated with the same path identifier.
 
 The Path ID of the initial path is 0. Connection IDs
-which are issued by an NEW_CONNECTION_ID frame {{Section 19.15. of QUIC-TRANSPORT}}
+which are issued by a NEW_CONNECTION_ID frame {{Section 19.15. of QUIC-TRANSPORT}}
 respectively are asociated with Path ID 0. Also, the Path ID for
 the connection ID specified in the "preferred address" transport parameter is 0.
 Use of the "preferred address" is considered as a migration event
 that does not change the Path ID.
 
 This extension uses multiple packet number spaces, one for each active path.
-As such each path has two associated packet number spaces for each direction.
+As such, each path maintains distinct packet number states for sending and receiving packets, as in {{QUIC-TRANSPORT}}.
 Using multiple packet number spaces enables direct use of the
 loss recovery and congestion control mechanisms defined in
-{{QUIC-RECOVERY}}.
+{{QUIC-RECOVERY}} on a per-path basis.
 
 Using multiple packet number spaces requires changes in the way AEAD is
 applied for packet protection, as explained in {{multipath-aead}},
 and tighter constraints for key updates, as explained in {{multipath-key-update}}.
 The Path Identifier associated with the Destination Connection ID is used to
-construct the packet protection nonce defined in {#multipath-aead}.
+construct the packet protection nonce defined in {{multipath-aead}}.
 
 This specification
 requires the sender to use a non-zero connection ID when opening an
@@ -232,8 +232,8 @@ of a path, but it can validate a new path created by a client.
 A new path can only be used once the associated 4-tuple has been validated
 by ensuring that the peer is able to receive packets at that address
 (see {{Section 8 of QUIC-TRANSPORT}}).
-The connection ID of a packet is used to associate a packet to a path identifier.
-And the path identifier is used to associate the packet to a packet number space.
+The connection ID of a packet binds the packet to a path identifier, and therefore
+to a packet number space.
 Further, the path identifier is used as numerical identifier
 in control frames. E.g. an endpoint sends a PATH_ABANDON frame to request its peer to
 abandon a certain path.
@@ -319,10 +319,11 @@ frame to provide new connection IDs and, respectively, the MP_RETIRE_CONNECTION_
 retire connection IDs after a successful handshake indicating multipath support by both endpoints.
 
 Endpoints MUST NOT issue connection IDs with path identifiers larger than
-the path limitation declared by the initial_max_paths transport parameter
-or MAX_PATHS frames.
+the path limitation advertised by the peer, corresponding to the maximum value
+between the peer's initial_max_paths transport parameter and received
+MAX_PATHS frames.
 
-To open a new path, an endpoint MUST use a connection ID assciated with
+To open a new path, an endpoint MUST use a connection ID associated with
 a new, unused Path ID.
 Still, the receiver may observe a connection ID associated with a used Path ID
 on different 4-tuples due to, e.g., NAT rebinding. In such case, the receiver reacts
@@ -344,7 +345,7 @@ Connection IDs cannot be reused, thus opening a new path requires the
 use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}}).
 Instead of NEW_CONNECTION_ID frame as specified in {{QUIC-TRANSPORT}},
 each endpoint uses MP_NEW_CONNECTION_ID frames
-to issue usable connections IDs to reach it. As such to open
+to issue usable Path ID-specific connections IDs to reach it. As such to open
 a new path by initiating path validation, both sides need at least
 one connection ID (see {{Section 5.1.1 of QUIC-TRANSPORT}}), which is associated
 with an unused Path ID.
@@ -676,7 +677,7 @@ in {{Section 8.2 of QUIC-TRANSPORT}}.
 The endhost can use all the paths in the "Active" state, provided
 that the congestion control and flow control currently allow sending
 of new data on a path. Note that if a path became idle due to a timeout,
-the endpoint SHOULD send an PATH_ABANDON frame before closing the path.
+the endpoint SHOULD send a PATH_ABANDON frame before closing the path.
 
 In the "Closing" state, the endhost SHOULD NOT send packets on this
 path anymore, as there is no guarantee that the peer can still map
@@ -1208,7 +1209,8 @@ PATH_ABANDON frames are ack-eliciting. If a packet containing
 a PATH_ABANDON frame is considered lost, the peer SHOULD repeat it.
 
 Once a Path ID is retired by PATH_ABANDON,
-it MUST NOT be reused on any other path.
+endpoints MUST NOT send packets associated with that Path ID anymore,
+even on other network paths.
 
 ## PATH_STANDBY frame {#path-standby-frame}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -341,7 +341,7 @@ All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
 ## Path Initiation {#path-initiation}
 
-Connection IDs cannot be reused, thus opening a new path requires the
+Opening a new path requires the
 use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}}).
 Instead of NEW_CONNECTION_ID frame as specified in {{QUIC-TRANSPORT}},
 each endpoint uses MP_NEW_CONNECTION_ID frames
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -153,7 +153,7 @@ loss recovery and congestion control mechanisms defined in
 {{QUIC-RECOVERY}} on a per-path basis.
 
 Using multiple packet number spaces requires changes in the way AEAD is
-applied for packet protection, as explained in {{multipath-aead}}. 
+applied for packet protection, as explained in {{multipath-aead}}.
 More concretely, the path identifier is used to construct the
 packet protection nonce in addition to the packet number
 in order to enable use of the same packet number on different paths.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -127,7 +127,7 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
 requires negotiation between the two endpoints using a new transport
 parameter, as specified in {{nego}}.
 
-This extension specifies a new Path Identifier (Path ID), which is 
+This extension specifies a new Path Identifier (Path ID), which is
 integer between 0 and 2^32 - 1 (inclusive). Path identifies are generated
 monotonically increasing and cannot be reused.
 
@@ -135,7 +135,7 @@ The Path ID is used to
 address a path the new multipath control frames,
 such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
 PATH_AVAILABLE {{path-available-frame}} as well as ACK_MP {{ack-mp-frame}}.
-Further, Connection IDs are issued per Path ID. 
+Further, Connection IDs are issued per Path ID.
 Each connection ID is associated with one path identifier
 but multiple connection IDs can be associated with the same path identifier.
 
@@ -281,7 +281,7 @@ This extension does not change the definition of any transport parameter
 defined in {{Section 18.2. of QUIC-TRANSPORT}}.
 
 The initial_max_paths transport parameter limits the initial maximum number of active paths
-that can be used during a connection. 
+that can be used during a connection.
 The transport parameter "active_connection_id_limit"
 {{QUIC-TRANSPORT}} limits the maximum number of active connection IDs per path when the
 initial_max_paths parameter is negotiated successfully.
@@ -323,8 +323,8 @@ the path limitation declared by the initial_max_paths transport parameter
 or MAX_PATHS frames.
 
 To open a new path, an endpoint MUST use a connection ID assciated with
-a new, unused Path ID. 
-Still, the receiver may observe a connection ID associated with a used Path ID 
+a new, unused Path ID.
+Still, the receiver may observe a connection ID associated with a used Path ID
 on different 4-tuples due to, e.g., NAT rebinding. In such case, the receiver reacts
 as specified in {{Section 9.3 of QUIC-TRANSPORT}} by initiating path validation
 and using a new connection ID for the same Path ID.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -153,10 +153,11 @@ loss recovery and congestion control mechanisms defined in
 {{QUIC-RECOVERY}} on a per-path basis.
 
 Using multiple packet number spaces requires changes in the way AEAD is
-applied for packet protection, as explained in {{multipath-aead}},
-and tighter constraints for key updates, as explained in {{multipath-key-update}}.
-The Path Identifier associated with the Destination Connection ID is used to
-construct the packet protection nonce defined in {{multipath-aead}}.
+applied for packet protection, as explained in {{multipath-aead}}. 
+More concretely, the path identifier is used to construct the
+packet protection nonce in addition to the packet number
+in order to enable use of the same packet number on different paths.
+Further, tighter constraints for key updates are needed, as explained in {{multipath-key-update}}.
 
 This specification
 requires the sender to use a non-zero connection ID when opening an
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -517,7 +517,7 @@ MUST send a PATH_ABANDON frame to retire the Path ID.
 
 ### Allocating, Consuming and Retiring Connection IDs {#consume-retire-cid}
 
-Each endpoints pre-allocate a Path Identifier for each new connection ID.
+Each connection ID is associated with a Path ID.
 The Path Identifier 0 indicates the initial path of the connection.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -127,12 +127,37 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
 requires negotiation between the two endpoints using a new transport
 parameter, as specified in {{nego}}.
 
-This extension uses multiple packet number spaces.
-When multipath is negotiated, each separate packet number space is linked to a path ID.
+This extension specifies a new Path Identifier (Path ID), which is
```
mirjak, line 130:
```suggestion
This extension specifies a new Path Identifier (Path ID), which is an
```

huitema (approved):
Looks good. Minor comments only.

draft-ietf-quic-multipath.md
```diff
@@ -127,12 +127,37 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+
+The Path ID of the initial path is 0. Connection IDs
+which are issued by an NEW_CONNECTION_ID frame {{Section 19.15. of QUIC-TRANSPORT}}
+respectively are asociated with Path ID 0. Also, the Path ID for
+the connection ID specified in the "preferred address" transport parameter is 0.
+Use of the "preferred address" is considered as a migration event
+that does not change the Path ID.
+
+This extension uses multiple packet number spaces, one for each active path.
+As such each path has two associated packet number spaces for each direction.
```
qdeconinck, line 150: I think I understand the meaning here, but the definition of "packet number space" in the first sentence (one per path) is not coherent with the second one (one per direction over a path). Here is a first try to better describe this, but I think we could do better.

```suggestion
This extension uses multiple packet number spaces, one for each active path.
As such, each path maintains distinct packet number states for sending and receiving packets, as in {{QUIC-TRANSPORT}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -127,12 +127,37 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
 Using multiple packet number spaces enables direct use of the
 loss recovery and congestion control mechanisms defined in
 {{QUIC-RECOVERY}}.
 
+Using multiple packet number spaces requires changes in the way AEAD is
+applied for packet protection, as explained in {{multipath-aead}},
+and tighter constraints for key updates, as explained in {{multipath-key-update}}.
+The Path Identifier associated with the Destination Connection ID is used to
+construct the packet protection nonce defined in {#multipath-aead}.
```
qdeconinck, line 159:
```suggestion
construct the packet protection nonce defined in {{multipath-aead}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -204,38 +220,29 @@ selects only one path to exchange such frames.
...
 of a path, but it can validate a new path created by a client.
 A new path can only be used once the associated 4-tuple has been validated
 by ensuring that the peer is able to receive packets at that address
 (see {{Section 8 of QUIC-TRANSPORT}}).
-The Path Identifier communicated when advertising a
-Destination Connection ID is used to associate a packet to a packet number space
-that is used on a valid path. Further, the
-Path Identifier associated with Destination Connection ID is used as numerical identifier
+The connection ID of a packet is used to associate a packet to a path identifier.
+And the path identifier is used to associate the packet to a packet number space.
```
qdeconinck, line 236: Agree with @huitema that it does not read well. What about this?

```suggestion
The connection ID of a packet binds the packet to a path identifier, and therefore
to a packet number space.
```

draft-ietf-quic-multipath.md
```diff
@@ -351,31 +314,40 @@ does not discuss when a client decides to initiate a new path. We
 delegate such discussion in separate documents.
 
 To let the peer open a new path, an endpoint needs to provide its peer with connection IDs
-and associated Path Identifiers for the new path.
+for at least one unused path identifier. Endpoints SHOULD use the MP_NEW_CONNECTION_ID
+frame to provide new connection IDs and, respectively, the MP_RETIRE_CONNECTION_ID frame to
+retire connection IDs after a successful handshake indicating multipath support by both endpoints.
```
qdeconinck, line 319: I think we should clarify that we discuss Path ID 0 here. By design, proposing new connection IDs on other Path IDs is only possible using MP_NEW_CONNECTION_ID.

draft-ietf-quic-multipath.md
```diff
@@ -351,31 +314,40 @@ does not discuss when a client decides to initiate a new path. We
...
+
+Endpoints MUST NOT issue connection IDs with path identifiers larger than
+the path limitation declared by the initial_max_paths transport parameter
+or MAX_PATHS frames.
```
qdeconinck, line 323: I think this is a bit ambiguous here, as the current text does not specify whether it refers to the value advertised by the sender, the receiver, or the minimum of both. To me, we should clarify this is relative to the value advertised by the peer.

```suggestion
Endpoints MUST NOT issue connection IDs with path identifiers larger than
the path limitation advertised by the peer, corresponding to the maximum value
between the peer's initial_max_paths transport parameter and received
MAX_PATHS frames.
```

Also, I would expect discussing here how the receiver should react if it receives Connection IDs for non-authorised paths IDs. Closing the connection with some MP_PROTOCOL_VIOLATION error?

draft-ietf-quic-multipath.md
```diff
@@ -351,31 +314,40 @@ does not discuss when a client decides to initiate a new path. We
...
 
-To open a new path, an endpoint SHALL use different Connection IDs on different paths.
-Still, the receiver may observe the same Connection ID used on different
-4-tuples due to, e.g., NAT rebinding. In such case, the receiver reacts
+To open a new path, an endpoint MUST use a connection ID assciated with
```
qdeconinck, line 325: A typo. Also, given that we restrict now to only client-initiated paths, we could replace "endpoint" by "client". Then, we could specify that the server can reply to such path opening by sending packets with the newly seen Path ID on the perceived network path.

```suggestion
To open a new path, an endpoint MUST use a connection ID associated with
```

draft-ietf-quic-multipath.md
```diff
@@ -705,7 +676,7 @@ in {{Section 8.2 of QUIC-TRANSPORT}}.
 The endhost can use all the paths in the "Active" state, provided
 that the congestion control and flow control currently allow sending
 of new data on a path. Note that if a path became idle due to a timeout,
-endpoints SHOULD send PATH_ABANDON frame before closing the path.
+the endpoint SHOULD send an PATH_ABANDON frame before closing the path.
```
qdeconinck, line 679:
```suggestion
the endpoint SHOULD send a PATH_ABANDON frame before closing the path.
```

draft-ietf-quic-multipath.md
```diff
@@ -1234,6 +1207,9 @@ Reason Phrase:
 PATH_ABANDON frames are ack-eliciting. If a packet containing
 a PATH_ABANDON frame is considered lost, the peer SHOULD repeat it.
 
+Once a Path ID is retired by PATH_ABANDON,
+it MUST NOT be reused on any other path.
```
qdeconinck, line 1211: The current formulation is a bit implicit about what is meant by "reused". What about this? In my opinion, we could also drop the "even on other network paths" part.
```suggestion
endpoints MUST NOT send packets associated with that Path ID anymore,
even on other network paths.
```

qdeconinck (commented):
This makes the draft a bit more readable. Some suggestions for typos + clarification purposes.

draft-ietf-quic-multipath.md
```diff
@@ -351,31 +314,41 @@ does not discuss when a client decides to initiate a new path. We
...
 
 To let the peer open a new path, an endpoint needs to provide its peer with connection IDs
-and associated Path Identifiers for the new path.
-
-To open a new path, an endpoint SHALL use different Connection IDs on different paths.
-Still, the receiver may observe the same Connection ID used on different
-4-tuples due to, e.g., NAT rebinding. In such case, the receiver reacts
+for at least one unused path identifier. Endpoints SHOULD use the MP_NEW_CONNECTION_ID
+frame to provide new connection IDs and, respectively, the MP_RETIRE_CONNECTION_ID frame to
+retire connection IDs after a successful handshake indicating multipath support by both endpoints.
```
mirjak, line 319:
```suggestion
for at least one unused path identifier. Connection IDs for paths with Path IDs other than 0, can only
be announced and retired with the new MP_NEW_CONNECTION_ID and MP_RETIRE_CONNECTION_ID frames.
After a successful handshake indicating multipath support by both endpoints, endpoints SHOULD also use the MP_NEW_CONNECTION_ID and MP_RETIRE_CONNECTION_ID frames to provide new connection IDs or, respectively, retire them for the initial path with Path ID 0.
```

draft-ietf-quic-multipath.md
mirjak: see below. Okay?

draft-ietf-quic-multipath.md
```diff
@@ -183,15 +208,6 @@ capitals, as shown here.
 We assume that the reader is familiar with the terminology used in
 {{QUIC-TRANSPORT}}. When this document uses the term "path", it refers
 to the notion of "network path" used in {{QUIC-TRANSPORT}}.
```
Yanmei-Liu, line 210: The path ID is a new concept which differs from RFC9000. I think it's better to update the definition of path and path id, but not to delete all this definition of the new concept.

draft-ietf-quic-multipath.md
```diff
@@ -127,11 +127,36 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
+PATH_AVAILABLE {{path-available-frame}} as well as ACK_MP {{ack-mp-frame}}.
+Further, Connection IDs are issued per Path ID.
+Each connection ID is associated with one path identifier
+but multiple connection IDs can be associated with the same path identifier.
+
+The Path ID of the initial path is 0. Connection IDs
+which are issued by a NEW_CONNECTION_ID frame {{Section 19.15. of QUIC-TRANSPORT}}
+respectively are asociated with Path ID 0. Also, the Path ID for
+the connection ID specified in the "preferred address" transport parameter is 0.
```
Yanmei-Liu, line 145:
```suggestion
the connection ID specified in the "preferred address" transport parameter is 2.
```

Yanmei-Liu:
Thanks a lot for all these work! Some minor suggestions.

draft-ietf-quic-multipath.md
mirjak: it's already covered in the intro now.

draft-ietf-quic-multipath.md
mirjak: I think this change needs discussion. See issue #317

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -351,31 +314,41 @@ does not discuss when a client decides to initiate a new path. We
...
 - PATH_STANDBY and PATH_AVAILABLE frames to express a preference
-in path usage (see {{path-standby-frame}} and {{path-available-frame}}
+in path usage (see {{path-standby-frame}} and {{path-available-frame}})
 
 All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
-## Path Initiation
+## Path Initiation {#path-initiation}
 
 Connection IDs cannot be reused, thus opening a new path requires the
```
marten-seemann, line 344: The logic here doesn't really work out anymore, since CIDs are now bound to a path.

draft-ietf-quic-multipath.md
```diff
@@ -543,22 +516,21 @@ MUST send a PATH_ABANDON frame to retire the Path ID.
 
 ### Allocating, Consuming and Retiring Connection IDs {#consume-retire-cid}
 
-Each endpoints pre-allocate a Path Identifier for each new Connection ID.
+Each endpoints pre-allocate a Path Identifier for each new connection ID.
```
marten-seemann, line 519: That's incorrect. Multiple CIDs are associated with the same path ID.

draft-ietf-quic-multipath.md
```diff
@@ -578,39 +550,39 @@ Path Identifiers.
 ### Effect of MP_RETIRE_CONNECTION_ID Frame {#retire-cid-close}
 
 Receiving a MP_RETIRE_CONNECTION_ID frame causes an endpoint to discard
-the resources associated with that Connection ID. Note that retirement of
-Connection IDs will not effect the use of Path Identifier for the specific path.
-The list of received packets used to send acknowledgements is also remain
+the resources associated with that connection ID. Note that retirement of
+connection IDs will not effect the use of Path Identifier for the specific path.
```
marten-seemann, line 554:
```suggestion
connection IDs will not retire the Path Identifier for the specific path.
```

draft-ietf-quic-multipath.md
```diff
@@ -578,39 +550,39 @@ Path Identifiers.
...
+The list of received packets used to send acknowledgements also remains
 uneffected as the packet number space is associated with a path.
```
marten-seemann, line 556: I find this confusing, and I'm surprised one would even consider this. This is just plain RFC 9000 behavior.

draft-ietf-quic-multipath.md
```diff
@@ -578,39 +550,39 @@ Path Identifiers.
...
 
-The peer, that sent RETIRE_CONNECTION_ID frame, can keep sending data using
+The peer, that sends the MP_RETIRE_CONNECTION_ID frame, can keep sending data using
 the same IP addresses and UDP ports previously associated with
-that Connection ID, but MUST use a different connection ID when doing so.
+that connection ID, but MUST use a different connection ID when doing so.
```
marten-seemann, line 560: Same here, this is just RFC 9000. I don't think we need to call out anything specifically here.

draft-ietf-quic-multipath.md
```diff
@@ -705,7 +677,7 @@ in {{Section 8.2 of QUIC-TRANSPORT}}.
 The endhost can use all the paths in the "Active" state, provided
 that the congestion control and flow control currently allow sending
```
marten-seemann, line 678: You don't necessarily send stream data.
```suggestion
that the congestion control currently allow sending
```

draft-ietf-quic-multipath.md
```diff
@@ -127,11 +127,36 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
-When multipath is negotiated, each separate packet number space is linked to a path ID.
+This extension specifies a new Path Identifier (Path ID), which is an
+integer between 0 and 2^32 - 1 (inclusive). Path identifies are generated
+monotonically increasing and cannot be reused.
+
+The Path ID is used to
+address a path in the new multipath control frames,
+such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
+PATH_AVAILABLE {{path-available-frame}} as well as ACK_MP {{ack-mp-frame}}.
+Further, Connection IDs are issued per Path ID.
```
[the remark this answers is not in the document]
huitema: We may discuss that later, but connection IDs are issued for the connection, because they must be unique withing the scope of the connection, and attached to a path-id.

draft-ietf-quic-multipath.md
```diff
@@ -127,11 +127,36 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
 Using multiple packet number spaces enables direct use of the
 loss recovery and congestion control mechanisms defined in
-{{QUIC-RECOVERY}}.
+{{QUIC-RECOVERY}} on a per-path basis.
+
+Using multiple packet number spaces requires changes in the way AEAD is
+applied for packet protection, as explained in {{multipath-aead}},
+and tighter constraints for key updates, as explained in {{multipath-key-update}}.
+The Path Identifier associated with the Destination Connection ID is used to
+construct the packet protection nonce defined in {{multipath-aead}}.
```
mirjak, line 159:
```suggestion
Using multiple packet number spaces requires changes in the way AEAD is
applied for packet protection, as explained in {{multipath-aead}}. 
More concretely, the path identifier is used to construct the
packet protection nonce in addition to the packet number
in order to enable use of the same packet number on different paths.
Further, tighter constraints for key updates are needed, as explained in {{multipath-key-update}}.
```

draft-ietf-quic-multipath.md
mirjak: True. However, this is meant to be an editorial PR. Given the SHOULD is already in the current version, maybe it cleaner to open and issue and then remove that eventually with a separate PR?

draft-ietf-quic-multipath.md
mirjak: This is still correct. CID need still to be unique per connection. However, I agree that this half sentence is probably simply unnecessary.

draft-ietf-quic-multipath.md
mirjak: This is correct because it doesn't say that it can be the same path ID. However, I agree it's confusing.

mirjak: Not fully sure what to do here. Maybe just remove the sentence entirely?
```suggestion
```

draft-ietf-quic-multipath.md
mirjak: True, we should at least not use normative language here as this is already normatively specified in RFC9000.
```suggestion
that connection ID, but has to use a different connection ID when doing so.
```

draft-ietf-quic-multipath.md
mirjak: Yes, if you don't send stream data flow control does not apply. However, when it applies you have to consider it. I think the sentence as is, is still more correct.

draft-ietf-quic-multipath.md
Yanmei-Liu: I'd suggest change the sentence into: 
Each endpoints _assign_ a Path Identifier for each new connection ID.

marten-seemann: I'd argue it's the wrong way around anyway. **First** you create the path ID, **then** you issue CIDs for this path.

Yanmei-Liu, line 519:
```suggestion
Endpoints first allocate unused Path Identifiers, then they issue Connection IDs for each Path Identifier.
```

mirjak: I don't think that is a useful sentence. I think we can just remove it. Or what do you think is missing if we remove it, @Yanmei-Liu ?

draft-ietf-quic-multipath.md
mirjak: @marten-seemann would that proposal work?

draft-ietf-quic-multipath.md
Yanmei-Liu: This sentence is certainly necessary, we should clarify the relationship between CID and PATH ID at the very beginning of this section.

draft-ietf-quic-multipath.md
Yanmei-Liu: The original statement is more accurate.

draft-ietf-quic-multipath.md
huitema: in any case, "uneffected" --> "unaffected".

draft-ietf-quic-multipath.md
huitema: I agree that the first version is better. If a frame is not subject to flow control, the flow control always allows it.

draft-ietf-quic-multipath.md
huitema: I think the first version is better, "Each endpoints pre-allocate a Path Identifier for each new connection ID." But the real process involves:

* Endpoints decide for which paths they will allocate connection identifiers, subject to both the limits imposed by "max paths" and the limits of local resource. 
* Endpoint advertise connection identifiers for these paths.
* Connection identifier values must be unique for the scope of the connection, regardless of path-id.
* The number of connection-id active for a given path must not exceed the active_connection_id_limit.

It should be clear that at any given time there will be paths for which no connection ID is advertised yet, because the peer did not yet have the time or resource to advertise them. This is not an error.

draft-ietf-quic-multipath.md
qdeconinck: Let's track this in #337 then.

qdeconinck (approved):
A few suggestions, but I would suggest merging this and addressing remaining concerns in sub issues.

draft-ietf-quic-multipath.md
```diff
@@ -543,22 +517,21 @@ MUST send a PATH_ABANDON frame to retire the Path ID.
 
 ### Allocating, Consuming and Retiring Connection IDs {#consume-retire-cid}
 
-Each endpoints pre-allocate a Path Identifier for each new Connection ID.
+Each endpoints pre-allocate a Path Identifier for each new connection ID.
```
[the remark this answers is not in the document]
mirjak: Maybe even simpler...?
```suggestion
Each connection ID is associated with a Path ID.
```

draft-ietf-quic-multipath.md
mirjak: We should rework this section in a separate PR.

mirjak: I created issue #338

draft-ietf-quic-multipath.md
```diff
@@ -127,11 +127,37 @@ path management in {{Section 9 of QUIC-TRANSPORT}} and therefore
...
+
+This extension uses multiple packet number spaces, one for each active path.
+As such, each path maintains distinct packet number states for sending and receiving packets, as in {{QUIC-TRANSPORT}}.
 Using multiple packet number spaces enables direct use of the
 loss recovery and congestion control mechanisms defined in
-{{QUIC-RECOVERY}}.
+{{QUIC-RECOVERY}} on a per-path basis.
+
+Using multiple packet number spaces requires changes in the way AEAD is
+applied for packet protection, as explained in {{multipath-aead}}.
```
mirjak, line 156:
```suggestion
applied for packet protection, as explained in {{multipath-aead}}.
```

Issue #340, opened 2024-05-24
What should an endpoint do if it receives a PATH_CHALLENGE for an path ID that it doesn't have CIDs for?


mirjak:

mirjak:
This is the discussion from PR #315:

*[mirjak](https://github.com/mirjak) [2 weeks ago](https://github.com/quicwg/multipath/issues/340#discussion_r1594249689)*
I think the client should only try to open a path if it had send CIDs for that path ID (and the reception has been ack'ed). Then if the server received a path challenges for a path ID it doesn't have a CID for, it should either be an error or it should send a path abandon on another path.

*@[qdeconinck](https://github.com/qdeconinck) qdeconinck [last week](https://github.com/quicwg/multipath/issues/340#discussion_r1604426437)*
> I think the client should only try to open a path if it had send CIDs for that path ID (and the reception has been ack'ed).

We could be in the situation where the server has sent MP_NEW_CONNECTION_ID frames to the client, and the client wants to open a new path by sending a PATH_CHALLENGE along with a MP_NEW_CONNECTION_ID frame with the associated Path ID.

*@[mirjak](https://github.com/mirjak) mirjak [last week](https://github.com/quicwg/multipath/issues/340#discussion_r1604651489)*
Yes sending it in the same packet is fine as well because than it cannot be lost or reordered separately.

*@[mirjak](https://github.com/mirjak) mirjak [yesterday](https://github.com/quicwg/multipath/issues/340#discussion_r1611853509)*
Actually this issue is not resolved because if you send it in the same packet, it's "just" a matter of processing it the right order. However, otherwise I don't think it's a good idea to save the path challenge for later because then you also have to specify how long to wait.

mirjak:
addressed by PR #341

PR #315, opened 2024-03-17
[+] Explicit Path ID: Clarify that endpoints use the same Path ID in both directions.

Yanmei-Liu:
Solving issue #294

PR #341, opened 2024-05-24
ignore or wait if no CID for path ID is available on receipt of PATH_…

mirjak:
…CHALLENGE

This PR is text from PR #315 that needs further discussion before merge. See issue #340 for discussion!

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -368,10 +368,11 @@ with the same, unused Path ID. If the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same path ID for sending the PATH_RESPONSE.
 
 If the server receives a PATH_CHALLENGE before receiving
-a MP_NEW_CONNECTION_ID for the specific path, it MAY choose
-to ignore the PATH_CHALLENGE, or it can choose to send the
-PATH_RESPONSE frame upon reception of a MP_NEW_CONNECTION_ID
-frame containing the corresponding Path ID.
+a MP_NEW_CONNECTION_ID for the specific path, it SHOULD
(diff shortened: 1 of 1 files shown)


yfmascgy (commented):
I wonder if we should design a mechanism to avoid such scenarios as much as possible. For instance, when a peer sends an MP_NEW_CONNECTION_ID for an unused Path ID to the receiving endpoint, it could prompt the receiving endpoint to send at least one CID for that Path ID if it has not done so before. This way, if no CID for that Path ID is available, we can simply ignore the PATH_CHALLENGE.

draft-ietf-quic-multipath.md
```diff
@@ -367,6 +367,12 @@ one connection ID (see {{Section 5.1.1 of QUIC-TRANSPORT}}), which is associated
 with the same, unused Path ID. If the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same path ID for sending the PATH_RESPONSE.
 
+If the server receives a PATH_CHALLENGE before receiving
+a MP_NEW_CONNECTION_ID for the specific path, it MAY choose
+to ignore the PATH_CHALLENGE, or it can choose to send the
+PATH_RESPONSE frame upon reception of a MP_NEW_CONNECTION_ID
+frame containing the corresponding Path ID.
```
mirjak, line 374:
```suggestion
If the server receives a PATH_CHALLENGE before receiving
a MP_NEW_CONNECTION_ID for the specific path, it SHOULD
to ignore the PATH_CHALLENGE. Note that the
MP_NEW_CONNECTION_ID might be sent in the same
packet and in this case the PATH_CHALLENGE SHOULD
be processed.
```

Issue #338, opened 2024-05-23
Clarify how new CIDs are allocated


mirjak:
The process involves:

- Endpoints decide for which paths they will allocate connection identifiers, subject to both the limits imposed by "max paths" and the limits of local resource.
- Endpoint advertise connection identifiers for these paths.
- Connection identifier values must be unique for the scope of the connection, regardless of path-id.
- The number of connection-id active for a given path must not exceed the active_connection_id_limit.

It should be clear that at any given time there will be paths for which no connection ID is advertised yet, because the peer did not yet have the time or resource to advertise them. This is not an error.

huitema:
We should perhaps make recommendations on the CID allocation strategy. Suppose for example that both endpoints advertise MAX_PATH=3, but one only provides CID for path_id=1, and the other for path_id=2. Looks obvious, but there would be no way to open either path_id=1 or path_id=2. or is there?

mirjak:
Maybe we should specify that path IDs has to be used in-order? This is actually issue #318

yfmascgy:
I feel the allocation strategy is more than specifying path IDs has to be used in-order. There are definitely scenarios where we would like to initiate multiple new paths concurrently but find no available CIDs for a high order path ID to use. See #370

mirjak:
True. Assuming that endpoint usually announce a small max paths value, we could say that they should provide on CID for all paths...?

LPardue:
Discussed during the interim: recommendations seem useful but that is the most we can do, @qdeconinck to write a PR

Issue #325, opened 2024-03-21
CID allocation strategy


Yanmei-Liu:
We need to specify CID allocation strategy in case endpoints doesn't have available CIDs with same unused Path ID to initialize a new path.

Also need to take security into consideration.

Comments from PR#292:

[marten-seemann](https://github.com/marten-seemann) [4 days ago](https://github.com/quicwg/multipath/pull/292/files#r1527502104)
> This seems backwards. Connection IDs are issues for paths, not the other way around.
Also, if I allow you to open 1 million paths, you're probably wise to not issue me 1 million CIDs.

mirjak:
What do you want to say here @Yanmei-Liu?

I think it simply the case if you don't have a CIDs for an unused path ID, you can't open a new path. However, the other end might purposefully not provide more path IDs. I think you just have to accept it.

If you have CIDs for an unused path ID from the peer but did not provide CIDs for the the same path ID to your peer, you need to do that before opening the path. #315 discusses that already.

What else do we need? or what else would we even be able to say?

mirjak:
@Yanmei-Liu can we close this issue and move the discussion to #338 instead?

mirjak:
Closing please discuss in #338. (Or reopen if that doesn't work)

PR #354, opened 2024-05-27
fix definition of MAX_PATHS frame

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -1432,26 +1414,39 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
+
 
 Loss or reordering can cause an endpoint to receive a MAX_PATHS frame with
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Path ID than was previously received. MAX_PATHS frames that
+do not announce a larger Path ID than previously received MUST be ignored.
+
+Endpoints SHOULD NOT issue new connection IDs which have path identifiers larger than
+the Path ID announced in the Maximum Path Identifier field in the
```
mirjak, line 1439:
```suggestion
the Path ID announced in the Maximum Path Identifier field in the
```

draft-ietf-quic-multipath.md
```diff
@@ -1432,26 +1414,39 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
+correspondes to value of initial_max_paths transport parameter.
+
+An endpoint MUST NOT initiate a path with a Path ID larger than the Maximum Path Identifier value.
+An endpoint MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION
+if a peer opens a path with a larger Path ID than permitted.
 
-An endpoint MUST NOT initiate a path with a Path ID higher than the Maximum Paths value.
-An endpoint MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION if a peer opens more paths than was permitted.
+When there are not enough unused path identifiers, endpoints can
+send a MAX_PATHS frame to inform the peer that new path identifiers are available.
```
huitema, line 1442: For information, I struggled with the same set of issues when writing the "even numbers" PR. The text that I came up with was:
```
Maximum Paths:
: A count of the cumulative number of paths that can be opened
over the lifetime of the connection.

The endpoint receiving this frame MAY send MP_NEW_CONNECTION_ID frames
with Path IDs strictly lower than twice the Maximum Paths.
For example, if an endpoint received a Maximum Paths value of 3,
it may send MP_NEW_CONNECTION_ID with Path ID 0, 2 and 4, but not higher.

The Maximum Paths value cannot exceed 2^31-1, as it is not
possible to encode Path IDs larger than 2^32-2. Receipt of a
Path Identifier larger than this limit MUST be treated
as a connection error of type FRAME_ENCODING_ERROR.

The Maximum Paths value MUST NOT be lower than the value
advertised in the "initial_max_paths" transport parameter. Receipt
of a value lower than the "initial_max_paths" transport parameter
MUST be treated as a connection error of type MP_PROTOCOL_ERROR.
```
Some of that text was motivated by the odd-even split, but a lot of it was about clarify how the path-ID is used.

huitema (commented):
I like the idea of moving all the recommendations on path-ID in a single space. You choose to move that to the definition of the MAX_PATHS frame and that's clearly a possibility, but RFC 9000 uses a different approach. The stream-ID structure [2.1. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-2.1)[Stream Types and Identifiers](https://www.rfc-editor.org/rfc/rfc9000.html#name-stream-types-and-identifier), and the restrictions on sending new streams are defined in [4.6. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-4.6)[Controlling Concurrency](https://www.rfc-editor.org/rfc/rfc9000.html#name-controlling-concurrency). The frame descriptions just describe the encoding, with text like _"Stream ID: A variable-length integer indicating the stream ID of the stream; see [Section 2.1](https://www.rfc-editor.org/rfc/rfc9000.html#stream-id)."_

On the text itself, I struggled with the same issues when writing PR #331, and I added a clip of what I wrote in the comments.

draft-ietf-quic-multipath.md
```diff
@@ -1424,26 +1406,39 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
 
 MAX_PATHS frames contain the following field:
 
 Maximum Path Identifier:
-: A count of the cumulative number of path that can be opened
-  over the lifetime of the connection. This value cannot exceed 2^32-1, as it is not
-  possible to encode Path IDs larger than 2^32-1. Receipt of a frame that permits
-  opening of a path with Path ID larger than this limit MUST be treated
-  as a connection error of type FRAME_ENCODING_ERROR.
+: The latest Path ID that can be used for this connection.
```
qdeconinck, line 1417: Note that the MAX_PATHS frame is a unidirectional advertisement.
```suggestion
: The largest Path ID that can be used for this connection by the receiving endpoint.
```

draft-ietf-quic-multipath.md
```diff
@@ -1424,26 +1406,39 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Path ID than was previously received. MAX_PATHS frames that
+do not announce a larger Path ID than previously received MUST be ignored.
+
+Endpoints SHOULD NOT issue new connection IDs which have path identifiers larger than
+the Path ID announced in the Maximum Path Identifier field in the
+MP_MAX_PATHS frame {{max-paths-frame}}.
+If no MAX_PATHS frame was received yet, the maximum Path ID
+correspondes to value of initial_max_paths transport parameter.
```
qdeconinck, line 1434:
```suggestion
corresponds to the value of initial_max_paths transport parameter.
```

draft-ietf-quic-multipath.md
```diff
@@ -1424,26 +1406,39 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
 
 Loss or reordering can cause an endpoint to receive a MAX_PATHS frame with
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Path ID than was previously received. MAX_PATHS frames that
+do not announce a larger Path ID than previously received MUST be ignored.
+
+Endpoints SHOULD NOT issue new connection IDs which have path identifiers larger than
+the Path ID announced in the Maximum Path Identifier field in the
+MP_MAX_PATHS frame {{max-paths-frame}}.
```
qdeconinck, line 1432: I think we should clarify that the MAX_PATHS frame is a unidirectional advertisement to the peer having the same semantic as the `initial_max_paths` TP. To that end, an endpoint MUST NOT issue new CIDs for a Path ID larger than the maximum one advertised by the peer (and not SHOULD NOT) and MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION. Allowing (without recommending) an endpoint to issue CIDs with Path ID larger than supported by the receiving host pose security concerns from a resource exhaustion viewpoint.

However, if the peer allows a maximum Path ID X that is larger than the local advertised maximum path ID Y (i.e., X > Y), then an endpoint may (MAY?) provides CIDs for Path IDs Z where Y < Z <= X.

Further, we probably mention the MAX_PATHS frame (and not MP_MAX_PATHS).

draft-ietf-quic-multipath.md
```diff
@@ -1424,26 +1406,39 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
+If no MAX_PATHS frame was received yet, the maximum Path ID
+correspondes to value of initial_max_paths transport parameter.
+
+An endpoint MUST NOT initiate a path with a Path ID larger than the Maximum Path Identifier value.
+An endpoint MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION
+if a peer opens a path with a larger Path ID than permitted.
```
qdeconinck, line 1438: To me, this whole paragraph can be removed once the corrections of the previous paragraph are integrated.

draft-ietf-quic-multipath.md
```diff
@@ -1416,34 +1398,46 @@ Sequence Number:
...
 
 Loss or reordering can cause an endpoint to receive a MAX_PATHS frame with
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Path ID than was previously received. MAX_PATHS frames that
+do not announce a larger Path ID than previously received MUST be ignored.
+
+Endpoints SHOULD NOT issue new connection IDs which have path identifiers larger than
+the Path ID announced in the Maximum Path Identifier field in the
+MP_MAX_PATHS frame {{max-paths-frame}}.
```
mirjak, line 1431:
```suggestion
MAX_PATHS frame.
```

draft-ietf-quic-multipath.md
```diff
@@ -1414,23 +1414,22 @@ MAX_PATHS Frame {
 MAX_PATHS frames contain the following field:
 
 Maximum Path Identifier:
-: The latest Path ID that can be used for this connection.
+: The largest Path ID that can be used for this connection.
   This value cannot exceed 2^32-1, as it is not
-  possible to encode Path IDs larger than 2^32-1.
+  possible to encode Path IDs larger than 2^32-1 as variable integer.
```
huitema, line 1420: Uh, no. We could encode up to 2^62 - 1. The real reason is that we cannot encode more than 32 bit of path-ID and 64 bit of packet number in the 96 bit AEAD nonce.

huitema (commented):
We really need a "flow control" section of some kind, explaining that we expect resource per active path (number space, ack list, pending packets) and per connection ID (memory), and how these resource are protected.

draft-ietf-quic-multipath.md
mirjak: That doesn't make sense. In this case we would allow 2^62 - 1 path IDs but require a key update earlier.

draft-ietf-quic-multipath.md
mirjak: We can define the max paths as either the cumulative number of path over the whole life or the largest id. The part that I was trying to fix is that the frame figure said the field is called "Maximum Paths" but the text was defining "Maximum Path Identifier". I thought the later is clearer but in the end we just have to fix it such that it is correct, one or the other way. @huitema do you prefer "Maximum Paths"?

mirjak:
> We really need a "flow control" section of some kind, explaining that we expect resource per active path (number space, ack list, pending packets) and per connection ID (memory), and how these resource are protected.

@huitema can you open an issue for that? It should probably be part of the implementation section. However, we might even remove or merge some of the text we have there as things are actually less complicated now. See also PR #358 where I already removed some text that is outdated.

mirjak:
@huitema as I also wrote in PR #355 I'm currently focusing first of all on defining things normatively only once. Further I really don#t think we should have normative requirements in the example section, so I also move most of this text now into the frame definition section. But I agree that it probably belongs in the path management section. I noticed this as well when I did another full editorial pass in #358. However, the clean up is my first priority now and then I would rather like to make another PR later to check all frame sections and move things accordingly.

draft-ietf-quic-multipath.md
huitema: I would be fine with that. It was actually the original design. But there are complexities. For example, can you have both a long-lived path number 1 and then create path ID 2^32 + 1?

mirjak: hm, okay I guess 2^32 - 1 paths are usually enough... In this case I think we need to actually use a MUST here and we should maybe remove the respective text from the intro in order to define it only once.

It's late here; I will update the PR accordingly tomorrow which gives also others a chance to comment.

draft-ietf-quic-multipath.md
mirjak: +1 to remove it

draft-ietf-quic-multipath.md
huitema: MAX_PATHS, MAX_PATH_ID, "bonnet blanc et blanc bonnet", as we say in French. I don't really care.

draft-ietf-quic-multipath.md
mirjak: But yes, we really need to clarify that the value is unidirectional. I was thinking at some point that this is a global value and thought that if the MAX_PATHS frame gets reordered, a MP_NEW_CONNECTION_ID frame using a new path id could arrive before. However, as MAX_PATHS is sent from one peer and MP_NEW_CONNECTION_ID from the other, this cannot happen. So this is should be a MUST NOT. I will propose further text to clarify tomorrow!

```suggestion
Endpoints MUST NOT issue new connection IDs which have path identifiers larger than
the Path ID announced by the peer in the Maximum Path Identifier field in the
MP_MAX_PATHS frame {{max-paths-frame}}.
```

draft-ietf-quic-multipath.md
mirjak: Also based on @huitema text below we need to add this:

```suggestion

The Maximum Path Identifier MUST NOT be lower than the value
advertised in the initial_max_paths transport parameter. Receipt
of a Path ID lower than the value of initial_max_paths transport parameter
MUST be treated as a connection error of type MP_PROTOCOL_ERROR.
```

draft-ietf-quic-multipath.md
```diff
@@ -1416,34 +1398,46 @@ Sequence Number:
...
-  possible to encode Path IDs larger than 2^32-1. Receipt of a frame that permits
-  opening of a path with Path ID larger than this limit MUST be treated
-  as a connection error of type FRAME_ENCODING_ERROR.
+: The largest Path ID that can be used for this connection.
+  This value cannot exceed 2^32-1, as it is not
+  possible to encode Path IDs larger than 2^32-1 as variable integer.
+
+Receipt of PATH_AVAILABLE, PATH_STANDBY, PATH_ABANDON or MP_ACK frames
+that uses a Path ID that is larger than the announced Path ID
+MUST be treated as a connection error of type FRAME_ENCODING_ERROR.
```
mirjak, line 1423: I don't FRAME_ENCODING_ERROR is actually the right error here. It should be MP_PROTOCOL_VIOLATION.

mirjak, line 1423: I don't FRAME_ENCODING_ERROR is actually the right error here. It should be MP_PROTOCOL_VIOLATION.

```suggestion
MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
```

draft-ietf-quic-multipath.md
```diff
@@ -1416,34 +1398,46 @@ Sequence Number:
...
 Loss or reordering can cause an endpoint to receive a MAX_PATHS frame with
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Path ID than was previously received. MAX_PATHS frames that
+do not announce a larger Path ID than previously received MUST be ignored.
+
+Endpoints SHOULD NOT issue new connection IDs which have path identifiers larger than
+the Path ID announced in the Maximum Path Identifier field in the
+MP_MAX_PATHS frame {{max-paths-frame}}.
+If no MAX_PATHS frame was received yet, the maximum Path Identifier
```
mirjak, line 1432:
```suggestion
If no MAX_PATHS frame was received yet, the Maximum Path Identifier
```

draft-ietf-quic-multipath.md
mirjak:
```suggestion
```

draft-ietf-quic-multipath.md
```diff
@@ -1416,34 +1398,46 @@ Sequence Number:
...
+corresponds to the value of initial_max_paths transport parameter.
+
+An endpoint MUST NOT initiate a path with a Path ID larger than the Maximum Path Identifier value.
+An endpoint MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION
+if a peer opens a path with a larger Path ID than permitted.
 
-An endpoint MUST NOT initiate a path with a Path ID higher than the Maximum Paths value.
-An endpoint MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION if a peer opens more paths than was permitted.
+When there are not enough unused path identifiers, endpoints can
+send a MAX_PATHS frame to inform the peer that new path identifiers are available.
```
mirjak, line 1440: I also proposed to remove this sentence for now. First it is unclear what "not enough" means and second we have two open issues to discuss this further (see #325 and #338).

```suggestion
```

draft-ietf-quic-multipath.md
mirjak: Okay, but I guess if we use the max path id field, we should also rename the frame... 

Looking at this now again I guess renaming the field in the definition to "Maximum Paths" would have been the smaller chance. I'll revise this PR again tomorrow in order to keep chances as small as possible for now (but ensure things are correct). Sorry for forth and back.

huitema:
> @huitema can you open an issue for that? It should probably be part of the implementation section. However, we might even remove or merge some of the text we have there as things are actually less complicated now. See also PR #358 where I already removed some text that is outdated.

See issue #360 .

draft-ietf-quic-multipath.md
```diff
@@ -1416,34 +1398,46 @@ Sequence Number:
...
 
 Maximum Path Identifier:
-: A count of the cumulative number of path that can be opened
-  over the lifetime of the connection. This value cannot exceed 2^32-1, as it is not
-  possible to encode Path IDs larger than 2^32-1. Receipt of a frame that permits
-  opening of a path with Path ID larger than this limit MUST be treated
-  as a connection error of type FRAME_ENCODING_ERROR.
+: The largest Path ID that can be used for this connection.
+  This value cannot exceed 2^32-1, as it is not
+  possible to encode Path IDs larger than 2^32-1 as variable integer.
```
Yanmei-Liu, line 1419:
```suggestion
  secure to use Path IDs larger than 2^32-1 with the security considerations outlined in {{multipath-aead}}.
```

Yanmei-Liu: @huitema is right. Variable integer is able to encode up to 2^62 -1. 
Using the Path ID as part of the AEAD nonce input limit the upper bound of Path ID.

draft-ietf-quic-multipath.md
Yanmei-Liu: Oh, we need to keep the mechanism for updating the Path ID limit. 
"Not enough" means all the available Path IDs issued by transport parameter and previous MAX_PATHS(_ID) frames are all used in current paths. If endpoints wants to create new paths(also need to issue new CIDs associated with these Path IDs) at this time, they need to use MAX_PATHS(_ID) frame to inform the peer that new used Path ID resources are allowed.

Yanmei-Liu (changes requested):

draft-ietf-quic-multipath.md
mirjak: Okay looking at 5.1.1 ([Issuing Connection IDs](https://datatracker.ietf.org/doc/html/rfc9000#name-issuing-connection-ids)) in RFC9000, I think we should not state these rules normatively again, we should just say these rules should now be applied per path.

draft-ietf-quic-multipath.md
```diff
@@ -1344,8 +1334,8 @@ Sequence Number:
 
 ## MAX_PATHS frames {#max-paths-frame}
 
-A MAX_PATHS frame (type=0x15228c0b) informs the peer of the cumulative number of paths
-it is permitted to open.
+A MAX_PATHS frame (type=0x15228c0b) informs the peer of largest Path ID that
+is permitted to use on any path.
```
mirjak, line 1338:
```suggestion
A MAX_PATHS frame (type=0x15228c0b) informs the peer of the cumulative number of paths
it is permitted to open.
```

draft-ietf-quic-multipath.md
```diff
@@ -1355,26 +1345,38 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
 ~~~
 MAX_PATHS Frame {
   Type (i) = 0x15228c0b,
-  Maximum Paths (i),
+  Maximum Path Identifier (i),
```
mirjak, line 1348:
```suggestion
  Maximum Paths (i),
```

draft-ietf-quic-multipath.md
```diff
@@ -1355,26 +1345,38 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
 
 Maximum Path Identifier:
-: A count of the cumulative number of path that can be opened
-  over the lifetime of the connection. This value cannot exceed 2^32-1, as it is not
-  possible to encode Path IDs larger than 2^32-1. Receipt of a frame that permits
-  opening of a path with Path ID larger than this limit MUST be treated
-  as a connection error of type FRAME_ENCODING_ERROR.
+: The largest Path ID that can be used for this connection.
+  This value cannot exceed 2^32-1, as it is not
+  secure to use Path IDs larger than 2^32-1 with the security considerations outlined in {{multipath-aead}}.
```
mirjak, line 1358:
```suggestion
: A count of the cumulative number of path that can be opened
  over the lifetime of the connection. This value cannot exceed 2^32-1, as it is not
  secure to use Path IDs larger than 2^32-1 with the security considerations outlined in {{multipath-aead}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -1355,26 +1345,38 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
+
+The Maximum Path Identifier MUST NOT be lower than the value
+advertised in the initial_max_paths transport parameter. Receipt
+of a Path ID lower than the value of initial_max_paths transport parameter
+MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
+Receipt of PATH_AVAILABLE, PATH_STANDBY, PATH_ABANDON or MP_ACK frames
+that uses a Path ID that is larger than the announced Path ID
+MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
```
mirjak, line 1366:
```suggestion
The Maximum Paths value MUST NOT be lower than the value
advertised in the initial_max_paths transport parameter. Receipt
of a Maximum Paths value lower than the value of initial_max_paths transport parameter
MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
Receipt of PATH_AVAILABLE, PATH_STANDBY, PATH_ABANDON or MP_ACK frames
that uses a Path ID that is larger than the announced Maximum Paths value
MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
```

draft-ietf-quic-multipath.md
```diff
@@ -1355,26 +1345,38 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
 
 Loss or reordering can cause an endpoint to receive a MAX_PATHS frame with
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Path ID than was previously received. MAX_PATHS frames that
```
mirjak, line 1369:
```suggestion
a smaller Maximum Paths value than was previously received. MAX_PATHS frames that
```

draft-ietf-quic-multipath.md
```diff
@@ -1355,26 +1345,38 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
+do not announce a larger Path ID than previously received MUST be ignored.
```
mirjak, line 1370:
```suggestion
do not announce a larger maximum Paths value than previously received MUST be ignored.
```

draft-ietf-quic-multipath.md
```diff
@@ -1355,26 +1345,38 @@ MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
...
+
+Endpoints MUST NOT issue new connection IDs which have path identifiers larger than
+the Path ID announced by the peer in the Maximum Path Identifier field in the MAX_PATHS frame.
```
mirjak, line 1373:
```suggestion
the Maximum Paths value announced by the peer in the MAX_PATHS frame.
```

draft-ietf-quic-multipath.md
```diff
@@ -1363,18 +1348,31 @@ MAX_PATHS Frame {
...
-  as a connection error of type FRAME_ENCODING_ERROR.
+: A count of the cumulative number of paths that can be opened
+  over the lifetime of the connection. This value MUST NOT exceed 2^32-1, as
+  Path IDs are defined with a maximum value 2^32-1 as the 32 bits of the Path ID are used
+  to calculate the nonce (see Section {{multipath-aead}}).
+
+The Maximum Paths value MUST NOT be lower than the value
+advertised in the initial_max_paths transport parameter. Receipt
+of a Maximum Paths value lower than the value of initial_max_paths transport parameter
+MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
```
mirjak, line 1359:
```suggestion
: A count of the cumulative number of paths that can be opened
  over the lifetime of the connection. This value MUST NOT exceed 2^32-1, as
  Path IDs are defined with a maximum value 2^32-1 as the 32 bits of the Path ID are used
  to calculate the nonce (see Section {{multipath-aead}}).
  The Maximum Paths value MUST NOT be lower than the value
  advertised in the initial_max_paths transport parameter. Receipt
  of an invalid Maximum Paths value MUST be treated as a
  connection error of type MP_PROTOCOL_VIOLATION.
```

draft-ietf-quic-multipath.md
```diff
@@ -562,7 +552,6 @@ When an endpoint finds it has not enough available unused path identifiers,
 it SHOULD send a MAX_PATHS frame to inform the peer that it could use larger active
 path identifiers.
```
mirjak, line 554:
```suggestion


```

draft-ietf-quic-multipath.md
```diff
@@ -1363,18 +1349,31 @@ MAX_PATHS Frame {
...
+
+Receipt of PATH_AVAILABLE, PATH_STANDBY, PATH_ABANDON or MP_ACK frames
+that uses a Path ID that is larger than the announced Maximum Paths value
+MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
 
 Loss or reordering can cause an endpoint to receive a MAX_PATHS frame with
-a lower path limit than was previously received. MAX_PATHS frames that
-do not increase the path limit MUST be ignored.
+a smaller Maximum Paths value than was previously received. MAX_PATHS frames that
+do not announce a larger maximum Paths value than previously received MUST be ignored.
```
mirjak, line 1367:
```suggestion
a smaller Maximum Paths value than was previously received. MAX_PATHS frames that
do not increase the path limit MUST be ignored.
```

draft-ietf-quic-multipath.md
```diff
@@ -1363,18 +1349,31 @@ MAX_PATHS Frame {
...
+a smaller Maximum Paths value than was previously received. MAX_PATHS frames that
 do not increase the path limit MUST be ignored.
 
-An endpoint MUST NOT initiate a path with a Path ID higher than the Maximum Paths value.
-An endpoint MUST terminate the a connection with an error of type MP_PROTOCOL_VIOLATION if a peer opens more paths than was permitted.
+Endpoints MUST NOT issue new connection IDs which have path identifiers larger than
+the Maximum Paths value announced by the peer in the MAX_PATHS frame.
+If no MAX_PATHS frame was received yet, the Maximum Paths value
+corresponds to the value of initial_max_paths transport parameter.
+
```
mirjak, line 1373:
```suggestion
```

draft-ietf-quic-multipath.md
```diff
@@ -1363,18 +1349,30 @@ MAX_PATHS Frame {
...
+When there are not enough unused path identifiers, endpoints can
```
mirjak, line 1374:
```suggestion
When there are no connection IDs available for unused path identifiers, endpoints can
```

draft-ietf-quic-multipath.md
```diff
@@ -556,11 +546,13 @@ using the corresponding connection IDs of the specified path and retire them
 with MP_RETIRE_CONNECTION_ID frames before adding the newly provided connection ID
 to the set of active connection IDs belonging to the specified path.
 
-Endpoints MUST NOT issue new connection IDs which have Path IDs larger than
-the Maximum Path Identifier field in MAX_PATHS frames {{max-paths-frame}}.
-When an endpoint finds it has not enough available unused path identifiers,
-it SHOULD send a MAX_PATHS frame to inform the peer that it could use larger active
-path identifiers.
+Endpoints MUST NOT issue new connection IDs which have Path IDs greater than
```
mirjak, line 549:
```suggestion
Endpoints MUST NOT issue new connection IDs with Path IDs greater than
```

Yanmei-Liu (approved):

huitema (approved):
Looks good. Also looks close to the limit of what we can do with "editorial" fixes. I think it is time to merge this, so we have a stable basis for further discussions.

PR #355, opened 2024-05-27
Move normative requirement away from example section

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -483,6 +483,9 @@ it SHOULD NOT use the associated Path ID in future frames, except
 in ACK_MP frames for inflight packets and
 in MP_RETIRE_CONNECTION_ID frames for connection ID retirement.
 
+After a path is abandoned, the Path ID MUST NOT be reused
```
mirjak, line 486:
```suggestion
After a path is abandoned, the Path ID MUST NOT be reused
```

draft-ietf-quic-multipath.md
```diff
@@ -481,6 +481,9 @@ it SHOULD NOT use the associated Path ID in future frames, except
...
+for new paths, as the Path ID is part of the nonce calculation {{multipath-aead}}.
+
```
huitema, line 486: We should also copy here the SHOULD requirement in the example text:
```
Endpoint SHOULD send MAX_PATHS frames {{max-paths-frame}} to raise
the limit of Path ID when endpoint finds there are not enough unused
Path ID (e.g. more than half of the available Path ID are used).
```

draft-ietf-quic-multipath.md
```diff
@@ -889,10 +892,6 @@ Client                                                      Server
 ~~~
 {: #fig-example-path-close1 title="Example of closing a path."}
 
-After a path is abandoned, the path identifier associated with the path
-is considered retired and MUST NOT be reused in new paths for security
-considerations discussed in {{multipath-aead}}.
-
```
huitema, line 895: We should also move the next paragraph.

huitema (commented):
We should move 2 paragraphs, "MUST NOT" ad "SHOULD", not just "MUST NOT".

draft-ietf-quic-multipath.md
qdeconinck, line 486: Copying as suggestion what Christian suggests. Note that I am not sure that we need to provide the example of "more than half of the available Path ID are used".

```suggestion

Endpoint SHOULD send MAX_PATHS frames {{max-paths-frame}} to raise
the limit of Path ID when endpoint finds there are not enough unused
Path ID (e.g. more than half of the available Path ID are used).
```

draft-ietf-quic-multipath.md
```diff
@@ -889,10 +892,6 @@ Client                                                      Server
...
 Endpoint SHOULD send MAX_PATHS frames {{max-paths-frame}} to raise
 the limit of Path ID when endpoint finds there are not enough unused
 Path ID (e.g. more than half of the available Path ID are used).
```
qdeconinck, line 897: From the suggestion of Christian.
```suggestion
```

qdeconinck (approved):
Ok, with the provided suggestions.

draft-ietf-quic-multipath.md
mirjak: This is moved as part of PR #354. I moved it into the MAX_PATHS section. Maybe it should rather go above into the path closure section but I think there is actually more that needs moving.

draft-ietf-quic-multipath.md
mirjak: This is moved as part of PR #354

draft-ietf-quic-multipath.md
mirjak: PR #354 basically moves this text in the MAX_PATHS frame section. It might fit better here but my main intention is right now to have it only once normatively defined and not in a section where we shouldn't have normative language. I think I will make another paths later where we can again move things around. However, right now I'm really focusing on ensuring we have it defined only once and clearly.

PR #361, opened 2024-05-28
Remove things on NEW/RETIRE_CID that have been normatively specified …

mirjak:
…in RFC9000

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -556,7 +556,7 @@ Sending a MP_RETIRE_CONNECTION_ID frame indicates that the connection ID
 will not be used anymore. If the path is still active, the peer SHOULD replace
 it with a new connection ID using a MP_NEW_CONNECTION_ID frame.
 
-Endpoints MUST NOT issue new connection IDs with Path IDs larger than
+Endpoints MUST NOT issue new connection IDs which have Path IDs larger than
 the Maximum Path Identifier field in MAX_PATHS frames {{max-paths-frame}}.
 When an endpoint finds it has not enough available unused path identifiers,
 it SHOULD send a MAX_PATHS frame to inform the peer that it could use larger active
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -529,7 +529,7 @@ is to not send a PATH_RESPONSE in response to the peer's PATH_CHALLENGE.
 ### Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
 
 With the multipath extension, each connection ID is associated with one path
-that is identified by the Path ID specified in the Path Identifier field of
+that is identified by the Path ID that is specified in the Path Identifier field of
 the MP_NEW_CONNECTION_ID frame {{mp-new-conn-id-frame}}.
 The Path ID 0 indicates the initial path of the connection.
 Respectively, the connection IDs used during the handshake belong to the initial path
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -533,7 +533,7 @@ that is idenfied by the Path ID as specified in the Path Identifier field of
 the MP_NEW_CONNECTION_ID frame {{mp-new-conn-id-frame}}.
 The Path ID 0 indicates the initial path of the connection.
 Respectively, the connection IDs used during the handshake belong to the initial path
-with Path ID 0. 
+with Path ID 0.
 The MP_NEW_CONNECTION_ID frame is used to new issue connection IDs for all paths.
 Endpoints SHOULD issue at least one connection ID for an unused path identifier
 to allow new paths to be opened.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -526,37 +526,37 @@ peer during the address validation procedure. According to
 is to not send a PATH_RESPONSE in response to the peer's PATH_CHALLENGE.
 
 
-### Allocating, Consuming and Retiring Connection IDs {#consume-retire-cid}
+### Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
 
-Each connection ID is associated with a Path ID.
-The Path Identifier 0 indicates the initial path of the connection.
-Endpoints SHOULD issue at least one unused connection ID with an unused path identifier.
+With the multipath extension, each connection ID is associated with one path
+that is idenfied by the Path ID as specified in the Path Identifier field of
+the MP_NEW_CONNECTION_ID frame {{mp-new-conn-id-frame}}.
+The Path ID 0 indicates the initial path of the connection.
+Respectively, the connection IDs used during the handshake belong to the initial path
+with Path ID 0. 
+The MP_NEW_CONNECTION_ID frame is used to new issue connection IDs for all paths.
+Endpoints SHOULD issue at least one connection ID for an unused path identifier
+to allow new paths to be opened.
 
-An endpoint maintains a set of connection IDs received from its peer for each path,
-any of which it can use when sending packets, as specified in {{Section 5.1 of QUIC-TRANSPORT}}.
-In the multipath extension, each connection ID belongs to one path specified by
-the Path Identifier field of MP_NEW_CONNECTION_ID frame in {{mp-new-conn-id-frame}}.
-The connection IDs used during the handshake belong to the initial path
-with Path ID 0. Usually, it is desired to provide at least one more connection ID for
-for all used path and at least one for an unused Path ID, to allow for migration and
-opening of new paths.
+Each endpoint maintains the set of connection IDs received from its peer for each path,
+any of which it can use when sending packets on that path; see also {{Section 5.1 of QUIC-TRANSPORT}}.
+Usually, it is desired to provide at least one additional connection ID for
+for all used paths, to allow for migration.
 
-If the client has used all the allocated connection IDs, it is supposed to retire
+{{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
+In order to identify a connection ID correctly when the mulitpath extension is used,
+endpoints have to use the MP_RETIRE_CONNECTION_ID frame instead
+of the RETIRE_CONNECTION_ID frame to indicate the respective Path ID together with the
+connection ID sequence number, at least for all paths with a Path ID other than 0.
+
+If the client has used all the allocated connection IDs for a path, it is supposed to retire
 those that are not used anymore, and the server is supposed to provide
-replacements, as specified in {{Section 5.1.1. of QUIC-TRANSPORT}}.
+replacements for that path, see {{Section 5.1.2. of QUIC-TRANSPORT}}.
 Sending a MP_RETIRE_CONNECTION_ID frame indicates that the connection ID
 will not be used anymore. If the path is still active, the peer SHOULD replace
 it with a new connection ID using a MP_NEW_CONNECTION_ID frame.
 
-Note that Sequeunce Number and Retire Prior To fields correspond to
-the path with the Path ID that is specified by the Path Identifier field.
-
-Upon receipt of an increased Retire Prior To field, the peer MUST stop
-using the corresponding connection IDs of the specified path and retire them
-with MP_RETIRE_CONNECTION_ID frames before adding the newly provided connection ID
-to the set of active connection IDs belonging to the specified path.
-
-Endpoints MUST NOT issue new connection IDs which have Path IDs larger than
+Endpoints MUST NOT issue new connection IDs with Path IDs larger than
 the Maximum Path Identifier field in MAX_PATHS frames {{max-paths-frame}}.
 When an endpoint finds it has not enough available unused path identifiers,
 it SHOULD send a MAX_PATHS frame to inform the peer that it could use larger active
@@ -1069,9 +1069,9 @@ MUST silently ignore the frame.
 ## ACK_MP Frame {#ack-mp-frame}
 
 The ACK_MP frame (types TBD-00 and TBD-01)
-is an extension of the ACK frame defined by {{QUIC-TRANSPORT}}. It is
+is an extension of the ACK frame specified in {{Section 19.3 of QUIC-TRANSPORT}}. It is
 used to acknowledge packets that were sent on different paths using
-multiple packet number spaces. If the frame type is TBD-01, ACK_MP frames
+each an own packet number space. If the frame type is TBD-01, ACK_MP frames
 also contain the sum of QUIC packets with associated ECN marks received
 on the acknowledged packet number space up to this point.
 
@@ -1241,11 +1241,12 @@ before or during path initiation.
 
 ## MP_NEW_CONNECTION_ID frames {#mp-new-conn-id-frame}
 
-An endpoint sends a MP_NEW_CONNECTION_ID frame (type=0x15228c09) instead of
-the NEW_CONNECTION_ID frame to provide its peer with alternative connection IDs for 1-RTT packets
-for a specific path. The peer can then used a different connection on the same path
-to break linkability when migrating connections;
-see {{Section 19.15 of QUIC-TRANSPORT}}.
+The MP_NEW_CONNECTION_ID frame (type=0x15228c09)
+is an extension of the NEW_CONNECTION_ID frame specified in
+{{Section 19.15 of QUIC-TRANSPORT}}.
+It is used to provide its peer with alternative connection IDs for 1-RTT packets
+for a specific path. The peer can then use a different connection on the same path
+to break linkability when migrating on that path; see also {{Section 9.5 of QUIC-TRANSPORT}}.
 
 MP_NEW_CONNECTION_ID frames are formatted as shown in {{fig-mp-connection-id-frame-format}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -526,37 +526,37 @@ peer during the address validation procedure. According to
...
-Note that Sequeunce Number and Retire Prior To fields correspond to
-the path with the Path ID that is specified by the Path Identifier field.
-
-Upon receipt of an increased Retire Prior To field, the peer MUST stop
-using the corresponding connection IDs of the specified path and retire them
-with MP_RETIRE_CONNECTION_ID frames before adding the newly provided connection ID
-to the set of active connection IDs belonging to the specified path.
-
-Endpoints MUST NOT issue new connection IDs which have Path IDs larger than
+Endpoints MUST NOT issue new connection IDs with Path IDs larger than
```
mirjak, line 559:
```suggestion
Endpoints MUST NOT issue new connection IDs which have Path IDs larger than
```

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,36 @@ peer during the address validation procedure. According to
...
 
-Note that Sequeunce Number and Retire Prior To fields correspond to
-the path with the Path ID that is specified by the Path Identifier field.
-
-Upon receipt of an increased Retire Prior To field, the peer MUST stop
-using the corresponding connection IDs of the specified path and retire them
-with MP_RETIRE_CONNECTION_ID frames before adding the newly provided connection ID
-to the set of active connection IDs belonging to the specified path.
-
 Endpoints MUST NOT issue new connection IDs which have Path IDs larger than
```
mirjak, line 559: This paragraph is revised in PR #354 instead

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,36 @@ peer during the address validation procedure. According to
...
-
-If the client has used all the allocated connection IDs, it is supposed to retire
+### Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
+
+With the multipath extension, each connection ID is associated with one path
+that is idenfied by the Path ID as specified in the Path Identifier field of
+the MP_NEW_CONNECTION_ID frame {{mp-new-conn-id-frame}}.
+The Path ID 0 indicates the initial path of the connection.
+Respectively, the connection IDs used during the handshake belong to the initial path
+with Path ID 0.
```
mirjak, line 536:
```suggestion
with Path ID 0.
```

draft-ietf-quic-multipath.md
```diff
@@ -1262,66 +1263,36 @@ MP_NEW_CONNECTION_ID Frame {
...
+The MP_RETIRE_CONNECTION_ID frame (type=0x15228c0a)
+is an extension of the RETIRE_CONNECTION_ID frame specified in
+{{Section 19.16 of QUIC-TRANSPORT}}. It is used
+to indicate that it will no longer use a connection ID for a specific path
+that was issued by its peer. To retire the connection ID used
+during the handshake on the initial path Path ID 0 is used.
 Sending a MP_RETIRE_CONNECTION_ID frame also serves as a request to the peer
-to send additional connection IDs for future use, unless the path specified
-by the Path ID has been abandoned. New connection IDs can be
+to send additional connection IDs for this pth for future use (see also {{Section 5.1 of QUIC-TRANSPORT}},
```
Yanmei-Liu:
```suggestion
to send additional connection IDs for this path (see also {{Section 5.1 of QUIC-TRANSPORT}},
```

draft-ietf-quic-multipath.md
```diff
@@ -1333,14 +1304,15 @@ MP_RETIRE_CONNECTION_ID Frame {
...
-  means the current connection ID can only be used on the corresponding path.
+Compared to the RETIRE_CONNECTION_ID frame specified in
+{{Section 19.16 of QUIC-TRANSPORT}}, the following
+field is added:
 
-Sequence Number:
-: The sequence number assigned to the connection ID by the sender for the path
-  specified by Path ID, encoded as a variable-length integer.
+Path Identifier:
+: The Path ID associated with the connection ID to retire.
```
Yanmei-Liu, line 1313: Note that the Sequence Number also has different meaning here. It's defined for per path.

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,36 @@ peer during the address validation procedure. According to
...
+Respectively, the connection IDs used during the handshake belong to the initial path
+with Path ID 0.
+The MP_NEW_CONNECTION_ID frame is used to issue new connection IDs for all paths.
+Endpoints SHOULD issue at least one connection ID for an unused path identifier
+to allow new paths to be opened.
+
+Each endpoint maintains the set of connection IDs received from its peer for each path,
+any of which it can use when sending packets on that path; see also {{Section 5.1 of QUIC-TRANSPORT}}.
+Usually, it is desired to provide at least one additional connection ID for
+for all used paths, to allow for migration.
```
qdeconinck, line 544:
```suggestion
Usually, it is desired to provide at least one additional connection ID for
all used paths, to allow for migration.
```

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,36 @@ peer during the address validation procedure. According to
...
+
+{{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
+In order to identify a connection ID correctly when the mulitpath extension is used,
```
qdeconinck, line 547:
```suggestion
In order to identify a connection ID correctly when the multipath extension is used,
```

draft-ietf-quic-multipath.md
```diff
@@ -1069,9 +1069,9 @@ MUST silently ignore the frame.
 ## ACK_MP Frame {#ack-mp-frame}
 
 The ACK_MP frame (types TBD-00 and TBD-01)
-is an extension of the ACK frame defined by {{QUIC-TRANSPORT}}. It is
+is an extension of the ACK frame specified in {{Section 19.3 of QUIC-TRANSPORT}}. It is
 used to acknowledge packets that were sent on different paths using
-multiple packet number spaces. If the frame type is TBD-01, ACK_MP frames
+each an own packet number space. If the frame type is TBD-01, ACK_MP frames
```
qdeconinck, line 1074: I cannot create a suggestion because of the deleted line, but I would suggest instead

"used to acknowledge packets that were sent on different paths, as each
path has its own packet number space. If the frame type [...]"

draft-ietf-quic-multipath.md
Yanmei-Liu: That do matters. When people are implementing the MP_CID_RETIRE frame, they could get all the important changes within this section, don't necessarily need to jump to the MP_NEW_CID frame definition section.

draft-ietf-quic-multipath.md
```diff
@@ -1262,66 +1263,36 @@ MP_NEW_CONNECTION_ID Frame {
...
-The Retire Prior To field applies to connection IDs established during
-connection setup. If the Path ID is 0 is applies to the initial path; see {{consume-retire-cid}}.
-The value in the Retire Prior To field MUST be less than or equal to the value
-in the Sequence Number field. Receiving a value in the Retire Prior To field
-that is greater than that in the Sequence Number field MUST be treated as
-a connection error of type FRAME_ENCODING_ERROR.
-
-Length, Connection ID, Stateless Reset Token fields have exactly the same
-definition in NEW_CONNECTION_ID frame {{Section 19.15 of QUIC-TRANSPORT}}.
+: The Path ID assocaited with the connection ID. This
```
qdeconinck, line 1271:
```suggestion
: The Path ID associated with the connection ID. This
```

draft-ietf-quic-multipath.md
```diff
@@ -1262,66 +1263,36 @@ MP_NEW_CONNECTION_ID Frame {
...
+  means the provided connection ID can only be used on the corresponding path.
 
-Note that connection IDs issued or retired in NEW_CONNECTION_ID frames MUST be treated as if
-their Path ID is 0 for compatibility with {{QUIC-TRANSPORT}}.
+Note that, other than for the NEW_CONNECTION_ID frame of {{Section 19.15 of QUIC-TRANSPORT}},
+the sequence number is set dependently for each path.
```
qdeconinck, line 1275:
```suggestion
the sequence number applies on a per-path context.
```

draft-ietf-quic-multipath.md
mirjak: But it's not a change. you look up the Path ID then you find the sequence number and remove the CID. So the only change is that you have to look up the Path ID first, which is described. That the sequence number is path-specific is only important when you select one in the NEW_CID frame.

draft-ietf-quic-multipath.md
```diff
@@ -1069,9 +1069,9 @@ MUST silently ignore the frame.
 ## ACK_MP Frame {#ack-mp-frame}
 
 The ACK_MP frame (types TBD-00 and TBD-01)
-is an extension of the ACK frame defined by {{QUIC-TRANSPORT}}. It is
+is an extension of the ACK frame specified in {{Section 19.3 of QUIC-TRANSPORT}}. It is
 used to acknowledge packets that were sent on different paths using
```
mirjak, line 1073:
```suggestion
used to acknowledge packets that were sent on different paths, as
```

draft-ietf-quic-multipath.md
mirjak, line 1074:
```suggestion
each path as its own packet number space. If the frame type is TBD-01, ACK_MP frames
```

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,36 @@ peer during the address validation procedure. According to
...
-The connection IDs used during the handshake belong to the initial path
-with Path ID 0. Usually, it is desired to provide at least one more connection ID for
-for all used path and at least one for an unused Path ID, to allow for migration and
-opening of new paths.
-
-If the client has used all the allocated connection IDs, it is supposed to retire
+### Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
+
+With the multipath extension, each connection ID is associated with one path
+that is identified by the Path ID as specified in the Path Identifier field of
```
Yanmei-Liu, line 532:
```suggestion
that is identified by the Path ID specified in the Path Identifier field of
```

draft-ietf-quic-multipath.md
qdeconinck: I agree with Mirja here. However, we could probably add a sentence in the above paragraph making the link with the "sequence number on a per-path context", similarly to what was done with MP_NEW_CONNECTION_ID frame, to address Yanmei's concern.

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,36 @@ peer during the address validation procedure. According to
...
+Each endpoint maintains the set of connection IDs received from its peer for each path,
+any of which it can use when sending packets on that path; see also {{Section 5.1 of QUIC-TRANSPORT}}.
+Usually, it is desired to provide at least one additional connection ID for
+all used paths, to allow for migration.
+
+{{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
+In order to identify a connection ID correctly when the multipath extension is used,
+endpoints have to use the MP_RETIRE_CONNECTION_ID frame instead
+of the RETIRE_CONNECTION_ID frame to indicate the respective Path ID together with the
+connection ID sequence number, at least for all paths with a Path ID other than 0.
```
Yanmei-Liu, line 550:
```suggestion
connection ID sequence number, at least for all paths with a Path ID other than 0.
To keep compatible with RFC9000, endpoints MUST treat RETIRE_CONNECTION_ID frame as 
used for initial Path (with Path ID 0), and it doesn't cause extra error.
```

mirjak: We don't need to use normative language here because this specification does not invalidate any RFC9000 frames. Therefore NEW/RETIRE_CID frames must be processed. The only thing we need to say is that the path ID for these CIDs is 0. That also doesn't need normative language either because that just a fact that is needed to be specified to implement this extension but not something for the endpoints to act on in a specific way.

draft-ietf-quic-multipath.md
```diff
@@ -526,36 +526,37 @@ peer during the address validation procedure. According to
...
-The connection IDs used during the handshake belong to the initial path
-with Path ID 0. Usually, it is desired to provide at least one more connection ID for
-for all used path and at least one for an unused Path ID, to allow for migration and
-opening of new paths.
-
-If the client has used all the allocated connection IDs, it is supposed to retire
+### Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
+
+With the multipath extension, each connection ID is associated with one path
+that is identified by the Path ID specified in the Path Identifier field of
```
mirjak, line 532:
```suggestion
that is identified by the Path ID that is specified in the Path Identifier field of
```

draft-ietf-quic-multipath.md
mirjak: You can make a proposal but a sentence like that make less sense here as the sequence number is selected when the NEW_CID frame is set and here only used. So saying the sequence number is "applied" on a per-path basis doesn't really make sense. Maybe the sequence number has be applied on a per path basis when the NEW_CID was sent but that information is already in the other section and not really helpful here.

draft-ietf-quic-multipath.md
```diff
@@ -517,43 +517,42 @@ peer during the address validation procedure. According to
...
+that is identified by the Path ID that is specified in the Path Identifier field of
+the MP_NEW_CONNECTION_ID frame {{mp-new-conn-id-frame}}.
+The Path ID 0 indicates the initial path of the connection.
+Respectively, the connection IDs used during the handshake belong to the initial path
+with Path ID 0.
+The MP_NEW_CONNECTION_ID frame is used to issue new connection IDs for all paths.
+In order to let the peer open new paths, it is RECOMMENDED to proactively
+issue a Connection ID for at least one unused path ID, as long as it remains
+compatible with the peer's Maximum Path ID limit.
+
```
huitema, line 532: Yes. Further discussion could be done on issue #338.

huitema (approved):
I think this is mostly good, but I am concerned that "retire prior to" is left a bit unspecified. Maybe we need to add some description of using it in the example section. But still, my priority right now is to get a reasonably stable draft that can be used as the basis for the next pass of editorial fixes.

mirjak:
@huitema "retire prio to" is specified in RFC9000:

>The Retire Prior To field applies to connection IDs established during connection setup and the preferred_address transport parameter; see [Section 5.1.2](https://datatracker.ietf.org/doc/html/rfc9000#retire-cid). The value in the Retire Prior To field MUST be less than or equal to the value in the Sequence Number field. Receiving a value in the Retire Prior To field that is greater than that in the Sequence Number field MUST be treated as a connection error of type FRAME_ENCODING_ERROR.

>Once a sender indicates a Retire Prior To value, smaller values sent in subsequent NEW_CONNECTION_ID frames have no effect. A receiver MUST ignore any Retire Prior To fields that do not increase the largest received Retire Prior To value.

>An endpoint that receives a NEW_CONNECTION_ID frame with a sequence number smaller than the Retire Prior To field of a previously received NEW_CONNECTION_ID frame MUST send a corresponding RETIRE_CONNECTION_ID frame that retires the newly received connection ID, unless it has already done so for that sequence number. 

The only thing that changes is that the MP_ frames are used und that the sequence numbers are per path. This is explained in the text. What else do you think needs to be specified?

draft-ietf-quic-multipath.md
```diff
@@ -517,43 +517,42 @@ peer during the address validation procedure. According to
...
+Each endpoint maintains the set of connection IDs received from its peer for each path,
+any of which it can use when sending packets on that path; see also {{Section 5.1 of QUIC-TRANSPORT}}.
+Usually, it is desired to provide at least one additional connection ID for
+all used paths, to allow for migration.
+
+{{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
+In order to identify a connection ID correctly when the multipath extension is used,
+endpoints have to use the MP_RETIRE_CONNECTION_ID frame instead
+of the RETIRE_CONNECTION_ID frame to indicate the respective Path ID together with the
+connection ID sequence number, at least for all paths with a Path ID other than 0.
```
mirjak, line 542:
```suggestion
connection ID sequence number, at least for all paths with a Path ID other than 0.
As endpoints can also use MP_NEW_CONNECTION_ID and MP_RETIRE_CONNECTION_ID for the initial path with Path ID 0, however, the use of NEW_CONNECTION_ID and RETIRE_CONNECTION_ID is still valid as well and endpoints need to process these frames accordingly belong to Path ID 0. 
```

draft-ietf-quic-multipath.md
```diff
@@ -1322,14 +1297,15 @@ MP_RETIRE_CONNECTION_ID Frame {
...
+field is added:
 
-Sequence Number:
-: The sequence number assigned to the connection ID by the sender for the path
-  specified by Path ID, encoded as a variable-length integer.
+Path Identifier:
+: The Path ID associated with the connection ID to retire.
 
+Note that the RETIRE_CONNECTION_ID frame can only be used to retire
+connection IDs for the initial path with Path ID 0.
```
mirjak, line 1308:
```suggestion
connection IDs for the initial path with Path ID 0.

As the MP_NEW_CONNECTION_ID frames applies the sequence number per path,
the sequence number in the MP_RETIRE_CONNECTION_ID frame
also need to be consider in the context of the the Path Identifier field.
```

draft-ietf-quic-multipath.md
```diff
@@ -517,43 +517,47 @@ peer during the address validation procedure. According to
...
+{{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
+In order to identify a connection ID correctly when the multipath extension is used,
+endpoints have to use the MP_RETIRE_CONNECTION_ID frame instead
+of the RETIRE_CONNECTION_ID frame to indicate the respective Path ID together with the
+connection ID sequence number, at least for all paths with a Path ID other than 0.
+Endpoints can also use MP_NEW_CONNECTION_ID and
+MP_RETIRE_CONNECTION_ID for the initial path with Path ID 0,
+however, the use of NEW_CONNECTION_ID and RETIRE_CONNECTION_ID
+is still valid as well and endpoints need to process these frames accordingly
+as corresponding to Path ID 0.
```
```suggestion
as corresponding to Path ID 0.
```

mirjak:
@huitema I merged this PR. Please open an issue if you think there is something more that needs to be specified about the "retire prior to" field!

huitema:
See issue #368.

Issue #295, opened 2024-03-12
Is RETIRE_PATH_ID needed?


mirjak:
This is from the discussion on PR #292:

**[yfmascgy](https://github.com/yfmascgy) [on Nov 21, 2023](https://github.com/quicwg/multipath/issues/new#r1400142192)**
```
What happens if the one endpoint doesn't how many path is needed? In the case of a deteriorating wifi, the endpoint wants to switch to 5G, but finds itself with no Path ID to use? What is the pre-allocating strategy?
```

**@[Yanmei-Liu](https://github.com/Yanmei-Liu) Yanmei-Liu [on Nov 21, 2023](https://github.com/quicwg/multipath/issues/new#r1400147984)
```
Endpoints just need to decide the max concurrent path and pre-allocate at lease one unused CID for each path Identifier. Endpoints don't need to decide which Path Identifier to use until client start a new path and choose one available CID with the unused Path Identifers.
```

**@[yfmascgy](https://github.com/yfmascgy) yfmascgy [on Nov 21, 2023](https://github.com/quicwg/multipath/issues/new#r1400179527)**
```
I am thinking a situation where you want max of three paths. You have your first path established. Try to establish another two paths on the 2nd sim and 3rd sim but failed. Then you don't have a path ID to build a path on the 4th sim even though it is available to use and can give you in total 2 paths (< max_num_path)
```

**@[kazuho](https://github.com/kazuho) kazuho [on Nov 21, 2023](https://github.com/quicwg/multipath/issues/new#r1400194859)**
```
Maybe using the scheme that we have for Connection IDs in QUIC v1 would drive the discussion.

If we do that, then

- the user of path IDs will send a TP indicating the number of paths from which it can send packets concurrently,
- the peer sends MAX_PATH_IDS frame signaling the maximum path ID the user can use (consider MAX_DATA or MAX_STREAMS as precedents),
- the user uses the path ids, and when it is done, sends a RETIRE_PATH_ID carrying the path id to be retired. We cannot use PATH_ABANDON here because it serves a different purpose and sent in the opposite direction,
- when receiving a RETIRE_PATH_ID frame, the peer may raise the limit advertised by MAX_PATH_IDS.
```

**@[Yanmei-Liu](https://github.com/Yanmei-Liu) Yanmei-Liu [on Nov 21, 2023](https://github.com/quicwg/multipath/issues/new#r1400306454)**
```
@kazuho 's idea is quite useful, I added these mechanism:

[...]

PR #365, opened 2024-05-28
Clarify path id retirement

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -478,10 +478,12 @@ The receiver of a PATH_ABANDON frame MAY also send
 a PATH_ABANDON frame to indicate its own unwillingness to receive
 any packet on this path anymore.
 
-The PATH_ABANDON frame retires the associated Path ID.
-When an endpoint receives a PATH_ABANDON frame,
+Receiption of the PATH_ABANDON frame is the first step to retire a
+path ID. However, the Path ID can only be retired after all active
+connection IDs for the Path ID have been retired.
```
Yanmei-Liu, line 483: I'm considering that there might be situations that attackers doesn't send CID retirement, on purpose to keep the path id and path resources not to be freed, and it could cause memory increase under this situation. 

Would it be a better way that we use Path Abandon like Connection_Close in RFC9000? After reception of Path Abandon, Endpoints could still send ACK_MP and MP_CID_Retirements within 3 PTO, but if endpoints doesn't send CID retirement or the packets containing CID retirements just lost, the peer could still be able to close the Path and release all the resources related after a period of time.

mirjak: This is not a problem because Maximum Paths limits the number of open paths. So if an attacker does not retire all CIDs, the path gets not closed but it can also not open more path when the Maximum Paths limit is reached.

qdeconinck: The Maximum Paths is not a (relative) counter indicating up to how many paths can be concurrently open. It indicates the maximum Path ID (or a way to derive the maximum path ID if we consider even/odd split) that can be used by the peer over the connection.

Regarding the first part about resource release, I agree that we should enable endpoints to release the resources related to the to-be-closed path if it has not seen CID retirement in the 3 PTO interval. This would make the endpoint ignores ("drops") further packets belonging to that path, which is in the end the idea of a closed path.

draft-ietf-quic-multipath.md
```diff
@@ -478,10 +479,12 @@ The receiver of a PATH_ABANDON frame MAY also send
 a PATH_ABANDON frame to indicate its own unwillingness to receive
 any packet on this path anymore.
 
-The PATH_ABANDON frame retires the associated Path ID.
-When an endpoint receives a PATH_ABANDON frame,
+Receiption of the PATH_ABANDON frame is the first step to retire a
```
qdeconinck, line 482:
```suggestion
Reception of the PATH_ABANDON frame is the first step to retire a
```

qdeconinck (changes requested):
There should be some "automatic" Path ID retirement if the path is in "Closing" state for too long, relying on the peer to have retired all its CIDs is probably wrong from a security perspective.

draft-ietf-quic-multipath.md
mirjak: Should we maybe say "release all ressources related to the Path ID" instead of "retire the Path ID"?

So what happens is, if you send an Abandon frame, you wait 3 PTOs, then you send the Retire CID frames of your connection IDs and release them. Further, if you can also release the CID that you got from your peer for this path ID, even if you didn't receive a retire CID frame from your peer, as you will anyway not use them anymore. The only risk is that the peer believes the path is not retired and therefore doesn't issue new CIDs for a potential new path if the path limit is reached. However, there is nothing you can do about this (as long as the abandon frame you've sent was acknowledged by the peer).

So we need to be more clear about this?

draft-ietf-quic-multipath.md
```diff
@@ -478,10 +479,12 @@ The receiver of a PATH_ABANDON frame MAY also send
 a PATH_ABANDON frame to indicate its own unwillingness to receive
 any packet on this path anymore.
 
-The PATH_ABANDON frame retires the associated Path ID.
-When an endpoint receives a PATH_ABANDON frame,
+Reception of the PATH_ABANDON frame is the first step to retire a
+path ID. However, the Path ID can only be retired after all active
+connection IDs for the Path ID have been retired.
```
mirjak, line 484:
```suggestion
Reception or sending of the PATH_ABANDON frame is
the first step to release all resources related to a
path ID. However, the Path ID can only be released after all active
connection IDs for the Path ID have been retired or timed-out after
the PATH_ABANDON frame was sent.
```
Not sure this is much better...

huitema (changes requested):
This is not an editorial change, and I don't think we have agreement there. I really like the idea that "ABANDON" clears all the resource associated with the path, and I dislike the idea that the endpoint abandoning the path needs to also send a series of "MP_RETIRE_CID" frames for each individual CID. How about we discuss that first in an issue before we make any change in the text?

mirjak:
I realised that this needs more discussion. However, the problem is the draft in its current form is not consistent. It was always the case that you have to send path_abandon, then wait for 3 PTOs and then send retire_cid. Actually this procedure has had quite some discussion in the past. We can discuss and change that again as things are different now with the path ID. However, I think we should fix the draft now (with this PR) to keep that procedure and then have a separate discussion to change it. Actually there is still issue #295 open which proposed a separate PATH_ID_RETIRE frame.

mirjak:
To be more concrete here: This is what section [4.3.1](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-4.3.1).  [Use PATH_ABANDON Frame to Close a Path](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#name-use-path_abandon-frame-to-c) says in the currently submitted version of the and this PR does not change it but align the other pieces with it:

> When a path is abandoned, all connection IDs allocated by both of the endpoints for the specified Path ID need to be retired. When sending or receiving a PATH_ABANDON frame, endpoints SHOULD wait for at least three times the current Probe Timeout (PTO) interval after the last packet was sent on the path, as defined in [Section 6.2](https://rfc-editor.org/rfc/rfc9002#section-6.2) of [[QUIC-RECOVERY](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-RECOVERY)], before sending MP_RETIRE_CONNECTION_ID frames. This is inline with the requirement of [Section 10.2](https://rfc-editor.org/rfc/rfc9000#section-10.2) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)]. Both endpoints SHOULD send MP_RETIRE_CONNECTION_ID frames for all connection IDs associated to the Path ID of the abandoned path to ensure that paths close cleanly and that delayed or reordered packets are properly discarded.

huitema (approved):
OK, @mirjak convinced me that the changes are in fact editorial, and that the current draft already requires that all existing CID be retired 1 by 1. I think that's a mistake, but let's have a discussion of that in issue #367, and then enter a PR once we have consensus.

PR #372, opened 2024-05-31
Cleanup MAX_PATHS

martinthomson:
There was a naming inconsistency here.

I went with one direction, which I think works best as it has a direct relation to the path identifier.  MAX_STREAMS uses a count because of the way in which streams are allocated odd/even.  That awkwardness we have for streams isn't necessary here.


Yanmei-Liu (approved):

mirjak:
I think if we want to rename, we should also rename max_initial_paths.

qdeconinck (approved):

Yanmei-Liu:
> I think if we want to rename, we should also rename max_initial_paths.

Yes, Indeed. I submit another PR #376 to rename the transport parameter and keep the semantics same.

PR #374, opened 2024-06-03
There is always a valid CID now for each path

mirjak:
This text is not needed anymore, as this issue does exists as we now manage CIDs per path with MP_NEW/RETIRE_CIDS

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -586,6 +586,12 @@ however, the use of NEW_CONNECTION_ID and RETIRE_CONNECTION_ID
 is still valid as well and endpoints need to process these frames accordingly
 as corresponding to Path ID 0.
 
+Endpoints MUST NOT issue new connection IDs with Path IDs greater than
+the Maximum Path Identifier field in MAX_PATHS frames (see Section {{max-paths-frame}}).
+When an endpoint finds it has not enough available unused path identifiers,
+it SHOULD send a MAX_PATHS frame to inform the peer that it could use new active
+path identifiers.
+
 If the client has used all the allocated connection IDs for a path, it is supposed to retire
 those that are not used anymore, and the server is supposed to provide
 replacements for that path, see {{Section 5.1.2. of QUIC-TRANSPORT}}.
@@ -593,39 +599,14 @@ Sending a MP_RETIRE_CONNECTION_ID frame indicates that the connection ID
 will not be used anymore. If the path is still active, the peer SHOULD replace
 it with a new connection ID using a MP_NEW_CONNECTION_ID frame.
 
-Endpoints MUST NOT issue new connection IDs with Path IDs greater than
-the Maximum Path Identifier field in MAX_PATHS frames (see Section {{max-paths-frame}}).
-When an endpoint finds it has not enough available unused path identifiers,
-it SHOULD send a MAX_PATHS frame to inform the peer that it could use new active
-path identifiers.
-
-
-### Effect of MP_RETIRE_CONNECTION_ID Frame {#retire-cid-close}
-
-Receiving a MP_RETIRE_CONNECTION_ID frame causes an endpoint to discard
-the resources associated with that connection ID. Note that retirement of
-connection IDs will not retire the Path ID for the specific path.
+Retirement of connection IDs will not retire the Path ID
+that correspones to the connection ID.
 The list of received packets used to send acknowledgements also remains
 unaffected as the packet number space is associated with a path.
 
 The peer that sends the MP_RETIRE_CONNECTION_ID frame can keep sending data
 on the path that the retired connection ID was used on but has
 to use a different connection ID for the same Path ID when doing so.
-If no other connection ID for the same Path ID is available, the endpoint cannot send on
-this path. This can happen if, e.g., the connection ID issuer requests retirement of a
-connection ID using the Retire Prior To field in the MP_NEW_CONNECTION_ID frame but does
-provide sufficient new connection IDs.
-
-Note that even if a peer cannot send on a path anymore because it does not have
-a valid connection ID to use, it can still acknowledge packets received on the path
-by sending ACK_MP frames on another path, if available.
-Similarly it can still send multipath control frames for that Path ID
-(such as PATH_ABANDON, PATH_STANDBY or PATH_AVAILABLE) on other available paths.
-
-If the peer cannot send on a path and no data is received on the path, the idle time-out will close
-the path. If, before the idle timer expires, a new connection ID gets issued
-by its peer, the endpoint can re-activate the path by
-sending a packet with a new connection ID on that path.
 
 ## Path States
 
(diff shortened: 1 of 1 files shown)


Yanmei-Liu (changes requested):
I think we definitely need to keep the guidance here, it's all about the corner case handle. Although we suggest that each endpoint issue at least one unused CID for each path, endpoints could still have no unused CID in real world cases:

- Scene 1: Endpoints have already used the last available CID, and the MP_NEW_CID frame containing the new CID of this Path is lost on inflight travel. When endpoint is waiting for the new CID to be retransmitted, there's a short period that no available CID can be used.
- Scene 2: Endpoints have used all the CIDs limited by _ active_connection_id_limit_ , and fail to retire the oldest CID. That could happen when the MP_RETIRE_CID is lost and waiting for retransmission. There's also a period that endpoints might not have available CID for the path. Specially, if endpoints set the limit _ active_connection_id_limit_ to 1, this situation could happen when endpoints want to trigger migration on a path.

qdeconinck:
I am not sure to understand how you could not have any CID for a path once some packets went over it, unless you abandon the path. An endpoint should not retire the last (sending) CID of a path ID if it does not have alternative.

BTW, it is invalid to have active_connection_id_limit to be 1, as RFC9000 states that it must be at least 2.

Yanmei-Liu:
> I am not sure to understand how you could not have any CID for a path once some packets went over it, unless you abandon the path. An endpoint should not retire the last (sending) CID of a path ID if it does not have alternative.
> 
> BTW, it is invalid to have active_connection_id_limit to be 1, as RFC9000 states that it must be at least 2.

Yes, there is a limitation in RFC9000 that active_connection_id_limit MUST be at least 2, but we didn't claim that in the current multi-path draft. As the semantic of active_connection_id_limit has been changed (for per path), the limit is also needed to be declared(at least 'follow the limit of RFC9000').

The scene of migration could happen like this (e.g. set active_connection_id_limit = 2) :
0. Client A and Server B have a connection negotiated active_connection_id_limit = 2, both endpoints issued 2 available CIDs for each path;
1. Client A create a new path with Server B, Server B choose to use DCID(path_id = P1, CID_seq = C1) for peer address(IP1, Port 1);
2. Client A is inside a NAT environment, after some time, the NAT port changed to(IP1, Port 2). While Server B find Client A is still using Path_id = P1, it consider the situation as NAT rebinding trigger the path migration, then it start to use DCID(path_id = P1, CID_seq = C2) for peer address(IP1, Port 2); 
3. When Client find out Server start to use DCID(path_id = P1, CID_seq = C2), it sends MP_NEW_CID(path_id = P1, CID_seq = C3, Retire Prio To = C2) containing the new CID. But this frame never touch the peer and waits to be retransmitted.  
4. When the NAT Port changes again at client side, the server have no unused available CID for the new address, it could still wait for the retransmission of MP_NEW_CID, and hold the path without any error.

mirjak:
The situation described is not specific for the multipath extension and is the same for RFC9000, so we don't have to say anything specific here. Also not that the requirement from RFC9000 that active_connection_id_limit has be at least 2 still applies without specifying this explicitly if we don't say anything else. Re-specifying something that is already specified somewhere else is best avoided because that makes it harder to change things later.

draft-ietf-quic-multipath.md
```diff
@@ -586,46 +584,27 @@ however, the use of NEW_CONNECTION_ID and RETIRE_CONNECTION_ID
...
+Sending a MP_RETIRE_CONNECTION_ID frame indicates that the connection ID
+will not be used anymore. If the path is still active, the peer SHOULD replace
+it with a new connection ID using a MP_NEW_CONNECTION_ID frame.
 
-### Effect of MP_RETIRE_CONNECTION_ID Frame {#retire-cid-close}
-
-Receiving a MP_RETIRE_CONNECTION_ID frame causes an endpoint to discard
-the resources associated with that connection ID. Note that retirement of
-connection IDs will not retire the Path ID for the specific path.
+Retirement of connection IDs will not retire the Path ID
```
Yanmei-Liu, line 600:
```suggestion
### Effect of MP_RETIRE_CONNECTION_ID Frame {#retire-cid-close}

Retirement of connection IDs will not retire the Path ID
```

draft-ietf-quic-multipath.md
```diff
@@ -586,46 +584,27 @@ however, the use of NEW_CONNECTION_ID and RETIRE_CONNECTION_ID
...
+that correspones to the connection ID.
 The list of received packets used to send acknowledgements also remains
 unaffected as the packet number space is associated with a path.
 
 The peer that sends the MP_RETIRE_CONNECTION_ID frame can keep sending data
 on the path that the retired connection ID was used on but has
 to use a different connection ID for the same Path ID when doing so.
```
Yanmei-Liu, line 607:
```suggestion
to use a different connection ID for the same Path ID when doing so.
If no other connection ID for the same Path ID is available, the endpoint cannot send on
this path. This can happen if, e.g., the connection ID issuer requests retirement of a
connection ID using the Retire Prior To field in the MP_NEW_CONNECTION_ID frame but does
provide sufficient new connection IDs.

Note that even if a peer cannot send on a path anymore because it does not have
a valid connection ID to use, it can still acknowledge packets received on the path
by sending ACK_MP frames on another path, if available.
Similarly it can still send multipath control frames for that Path ID
(such as PATH_ABANDON, PATH_STANDBY or PATH_AVAILABLE) on other available paths.

If the peer cannot send on a path and no data is received on the path, the idle time-out will close
the path. If, before the idle timer expires, a new connection ID gets issued
by its peer, the endpoint can re-activate the path by
sending a packet with a new connection ID on that path.
```

mirjak: This text is not needed anymore because this cannot happen with the new MP_NEW_CID frame.

draft-ietf-quic-multipath.md
```diff
@@ -586,46 +586,27 @@ however, the use of NEW_CONNECTION_ID and RETIRE_CONNECTION_ID
...
 When an endpoint finds it has not enough available unused path identifiers,
 it SHOULD send a MAX_PATHS frame to inform the peer that it could use new active
 path identifiers.
 
+If the client has used all the allocated connection IDs for a path, it is supposed to retire
+those that are not used anymore, and the server is supposed to provide
+replacements for that path, see {{Section 5.1.2. of QUIC-TRANSPORT}}.
+Sending a MP_RETIRE_CONNECTION_ID frame indicates that the connection ID
+will not be used anymore. If the path is still active, the peer SHOULD replace
+it with a new connection ID using a MP_NEW_CONNECTION_ID frame.
```
yfmascgy, line 598: Should we use "it" (singular) or "them" (plural) in the last sentence?

qdeconinck (commented):
Probably need to clarify if retiring all the DCIDs of a path without abandoning it is a valid behaviour.

Issue #318, opened 2024-03-17
Should path numbers be continuous?


huitema:
We observed an interop failure between picoquic and the Quiche implementation modified by Quentin. The picoquic client is waiting for a CID with pathID=1 before opening the second path with path number 1, but Quentin's implementation will only send paths with even numbers.

The immediate interop issue would be fixed if Picoquic used number 2 for its first path creation, but that is not a long term solution. If one implementation expects to see all number used and the other only uses one of them, we will shortly see an issue -- some CID do not get used but still contribute to the "MAX_PATHS" limits, or to the implicit number of path limit at the client. Creation of new paths will then stop, even if the number of paths in actual use remains below the limit.

Should we have a rule about not skipping path numbers?

marten-seemann:
Path validation on a particular path might fail, without the server receiving any packet for that path. When the client then validates another path, this will look like a skipped path to the server.

So to answer the question: Path IDs SHOULD be used sequentially, but you can't enforce this.

huitema:
Yes, we do have the case in with the client probes path[3] and the server never gets the probe. The client should send an Abandon[3] to clear the state.

marten-seemann:
That's true, but you don't know when the PATH_ABANDON arrives. It might be lost / reordered. So you'll need to deal with what looks like non-continuous use of Path IDs anyway.

huitema:
Yes. My real concern was one of the nodes issuing just some CID -- like Quentin's code only creating CID for even numbers, 0, 2, etc. We will have two limits:

1) the "max paths" value, which determines the maximum but does not force the peers to send that much.
2) the lower "local maximum" value, based on available memory and other local resource.

My real concern is that if nodes are not going to publish CID for all possible paths, they should start with the lowest numbers to maximize chances of finding numbers supported by both sides.

marten-seemann:
I would consider this a misbehaviour of the peer. If you refuse to use path IDs available to you, the protocol might get stuck.

huitema:
Yes. But the current draft does not say "dont do that"

mirjak:
[...]

PR #375, opened 2024-06-04
Recommendations regarding CID allocation and Path ID continuity

### LINKED ISSUE #338: Clarify how new CIDs are allocated

qdeconinck:
Fix #338, #318.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -582,7 +582,9 @@ in the usage of Path ID through their connection ID advertisements. For instance
 if the maximum Path ID limit is 2 and the endpoint wants to provide connection IDs
 for only one Path ID inside range [1, 2], it should select Path ID 1 (and not Path
 ID 2). Similarly, endpoints SHOULD consume Path IDs in a continuous way, i.e., when
-creating paths.
+creating paths. However, endpoints cannot expect to receive path
+creation attempts with in order use of Path IDs
+due to out-of-order delivery or path validation failure.
 
 {{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
 In order to identify a connection ID correctly when the multipath extension is used,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -575,11 +575,11 @@ any of which it can use when sending packets on that path; see also {{Section 5.
 Usually, it is desired to provide at least one additional connection ID for
 all used paths, to allow for migration.
 
-Over a given path, both endpoints use Connection IDs associated to a given Path
-ID. To initiate a path, each endpoint needs to advertise at least one Connection ID
+Over a given path, both endpoints use connection IDs associated to a given Path
+ID. To initiate a path, each endpoint needs to advertise at least one connection ID
 for a given Path ID to its peer. Endpoints SHOULD NOT introduce discontinuity
-in the usage of Path ID through their Connection ID advertisements. For instance,
-if the Maximum Path ID limit is 2 and the endpoint wants to provide Connection IDs
+in the usage of Path ID through their connection ID advertisements. For instance,
+if the maximum Path ID limit is 2 and the endpoint wants to provide connection IDs
 for only one Path ID inside range [1, 2], it should select Path ID 1 (and not Path
 ID 2). Similarly, endpoints SHOULD consume Path IDs in a continuous way, i.e., when
 creating paths.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -578,10 +578,11 @@ all used paths, to allow for migration.
 Over a given path, both endpoints use Connection IDs associated to a given Path
 ID. To initiate a path, each endpoint needs to advertise at least one Connection ID
 for a given Path ID to its peer. Endpoints SHOULD NOT introduce discontinuity
-in the usage of Path ID through their Connection ID advertisements. For instance, if the Maximum Path ID
-limit is 2 and the endpoint wants to provide Connection IDs for only one Path ID inside range[1, 2],
-it should select Path ID 1 (and not Path ID 2). Similarly, endpoints SHOULD consume
-Path IDs in a continuous way, i.e., when creating paths.
+in the usage of Path ID through their Connection ID advertisements. For instance,
+if the Maximum Path ID limit is 2 and the endpoint wants to provide Connection IDs
+for only one Path ID inside range [1, 2], it should select Path ID 1 (and not Path
+ID 2). Similarly, endpoints SHOULD consume Path IDs in a continuous way, i.e., when
+creating paths.
 
 {{Section 5.1.2. of QUIC-TRANSPORT}} specifies the retirement of connection IDs.
 In order to identify a connection ID correctly when the multipath extension is used,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -575,6 +575,14 @@ any of which it can use when sending packets on that path; see also {{Section 5.
 Usually, it is desired to provide at least one additional connection ID for
 all used paths, to allow for migration.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,14 @@ any of which it can use when sending packets on that path; see also {{Section 5.
 Usually, it is desired to provide at least one additional connection ID for
 all used paths, to allow for migration.
 
+Over a given path, both endpoints use Connection IDs associated to a given Path
+ID. To initiate a path, each endpoint needs to advertise at least one Connection ID
+for a given Path ID to its peer. Endpoints SHOULD NOT create Path ID discontinuity
+through their Connection ID advertisements. For instance, if the Maximum Path ID
+limit is 2 and the endpoint wants to provide Connection IDs for only one Path ID,
+it should select Path ID 1 (and not Path ID 2). Similarly, endpoints SHOULD consume
+Path IDs in a continuous way, i.e., when creating paths.
```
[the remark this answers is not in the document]
qdeconinck: Not quite; this is not because a Path ID is valid (e.g., if max path id is 2, Path IDS 0 1 and 2 are valid) that you can use it nor "generate" it. A Path ID becomes usable once Connection IDs associated to that specific Path ID are advertised. We should ensure that the way we propose Path IDs through MP_NEW_CONNECTION_IDs do not create such discontinuity (e.g., client only provide CIDs for Path ID 0 and 2, while server only provide CIDs for Path ID 0 and 1); but I haven't found a better way to phrase this yet.

draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,14 @@ any of which it can use when sending packets on that path; see also {{Section 5.
 Usually, it is desired to provide at least one additional connection ID for
 all used paths, to allow for migration.
 
+Over a given path, both endpoints use Connection IDs associated to a given Path
+ID. To initiate a path, each endpoint needs to advertise at least one Connection ID
+for a given Path ID to its peer. Endpoints SHOULD NOT create Path ID discontinuity
+through their Connection ID advertisements. For instance, if the Maximum Path ID
+limit is 2 and the endpoint wants to provide Connection IDs for only one Path ID,
```
Yanmei-Liu, line 582:
```suggestion
limit is 2 and the endpoint wants to provide Connection IDs for only one Path ID inside range[1, 2],
```

draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,14 @@ any of which it can use when sending packets on that path; see also {{Section 5.
 Usually, it is desired to provide at least one additional connection ID for
 all used paths, to allow for migration.
 
+Over a given path, both endpoints use Connection IDs associated to a given Path
+ID. To initiate a path, each endpoint needs to advertise at least one Connection ID
+for a given Path ID to its peer. Endpoints SHOULD NOT create Path ID discontinuity
```
Yanmei-Liu, line 580:
```suggestion
for a given Path ID to its peer. Endpoints SHOULD NOT use Path ID discontinuity
```

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
Yanmei-Liu: I agree with your point. 
Some minor fix. About the example: We jumped out of Path ID 0 because it's the initial Path ID, but it's still an available choice for endpoints.

draft-ietf-quic-multipath.md
qdeconinck: Maybe "Endpoints SHOULD NOT introduce discontinuity in the usage of Path ID through their Connection ID advertisements"?

draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,14 @@ any of which it can use when sending packets on that path; see also {{Section 5.
...
+through their Connection ID advertisements. For instance, if the Maximum Path ID
```
Yanmei-Liu, line 581:
```suggestion
 in the usage of Path ID through their Connection ID advertisements. For instance, if the Maximum Path ID
```

Yanmei-Liu (approved):

mirjak:
I thought we also wanted to change this recommendation:

> In order to let the peer open new paths, it is RECOMMENDED to proactively 
> issue a Connection ID for at least one unused path ID, as long as it remains
> compatible with the peer's Maximum Path ID limit.

We should say that it is recommended to provide CIDs for the minimum of the max path ids of both ends, instead of just one unused.

draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,15 @@ any of which it can use when sending packets on that path; see also {{Section 5.
 Usually, it is desired to provide at least one additional connection ID for
 all used paths, to allow for migration.
```
mirjak, line 576: I think there are some additional points here that could be clarified:
```suggestion
all used paths, to allow for migration. An endpoint can provide up
to active_connection_ids connection IDs per Path ID
as discussed in {{Section 5.1.1 of QUIC- TRANSPORT}}.
As further specified in {{Section 5.1 of QUIC- TRANSPORT}} connection IDs
cannot be issued more than once on the same connection
and therefore are unique for the scope of the connection,
regardless of the associated Path ID.
```

qdeconinck: Do we not already state the CID limit per path ID this before? Also, the TP is active_connection_id_limit and it seems there is a space in QUIC- TRANSPORT.
I am not sure we need the last sentence (but we can include it if it really helps).

draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,15 @@ any of which it can use when sending packets on that path; see also {{Section 5.
...
 
+Over a given path, both endpoints use connection IDs associated to a given Path
+ID. To initiate a path, each endpoint needs to advertise at least one connection ID
+for a given Path ID to its peer. Endpoints SHOULD NOT introduce discontinuity
+in the usage of Path ID through their connection ID advertisements. For instance,
+if the maximum Path ID limit is 2 and the endpoint wants to provide connection IDs
+for only one Path ID inside range [1, 2], it should select Path ID 1 (and not Path
+ID 2). Similarly, endpoints SHOULD consume Path IDs in a continuous way, i.e., when
+creating paths.
```
mirjak, line 585:
```suggestion
creating paths. However, endpoints cannot expect to receive path
creation attempts with in order use of Path IDs
due to out-of-order delivery or path validation failure.
```

draft-ietf-quic-multipath.md
```diff
@@ -575,6 +575,18 @@ any of which it can use when sending packets on that path; see also {{Section 5.
...
+ID. To initiate a path, each endpoint needs to advertise at least one connection ID
+for a given Path ID to its peer. Endpoints SHOULD NOT introduce discontinuity
+in the issuing of Path IDs through their connection ID advertisements as path creation
+requires available connection IDs for the same Path ID on both sides. For instance,
+if the maximum Path ID limit is 2 and the endpoint wants to provide connection IDs
+for only one Path ID inside range [1, 2], it should select Path ID 1 (and not Path
+ID 2). Similarly, endpoints SHOULD consume Path IDs in a continuous way, i.e., when
+creating paths. However, endpoints cannot expect to receive path
+creation attempts with in order use of Path IDs
+due to out-of-order delivery or path validation failure.
```
mirjak, line 588:
```suggestion
creating paths. However, endpoints cannot expect to receive new connection IDs
or path creation attempts with in order use of Path IDs
due to out-of-order delivery or path validation failure.
```

Issue #324, opened 2024-03-21
Rename "initial_max_paths / MAX_PATHS" to "intial_max_path_id / MAX_PATH_ID" ?


Yanmei-Liu:
Record from comments:

[marten-seemann](https://github.com/marten-seemann) [4 days ago](https://github.com/quicwg/multipath/pull/292#discussion_r1527500625)
> Probably better: initial_max_path_id. This would mirror the meaning of the frame as suggested in https://github.com/quicwg/multipath/pull/292/files#r1527500161.

[michael-eriksson](https://github.com/michael-eriksson) [4 days ago](https://github.com/quicwg/multipath/pull/292#discussion_r1527532226)
> I think it should remain initial_max_paths to be symmetric with initial_max_streams_*.

> The explanation can be copied from [RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.22.1):

>> Setting this parameter is equivalent to sending a MAX_PATHS of the corresponding type with the same value.

[marten-seemann](https://github.com/marten-seemann) [4 days ago](https://github.com/quicwg/multipath/pull/292#discussion_r1527658157)
> My 2c: There's only so much value you get from consistency with RFC 9000. This has created a fair amount of confusion in the review of this PR, and it seems like calling it initial_max_path_id (and the corresponding renaming of the frame), would make it 100% clear what is meant, without the need for long explanations.

[michael-eriksson](https://github.com/michael-eriksson) [3 days ago](https://github.com/quicwg/multipath/pull/292#discussion_r1527988506)
> Well, then I think that the frame type should be MAX_PATH_ID to further reduce the confusion and increase the multipath specification consistency.

mirjak:
the current draft (-07) uses MAX_PATH to announce the max number of paths and not the largest path ID that can be used. So the name matches the meaning, I would say.

Is this issue to discuss to use the max path ID instead (or was this just about aligning the naming)?

mirjak:
This is an entire mess right now in the draft:

initial_max_path is defined like this (as the max number of path):

```
[...]

PR #376, opened 2024-06-04
Rename transport parameter initial_max_paths to initial_max_path_id

Yanmei-Liu:
This PR try to solve the issue #324 and remove the outdated text.

As @martinthomson pointed out in PR #372, we don't have to use a count. This would help remove the ambiguity in the old draft version: 
- When endpoints negotiate a transport parameter value with N, but initial path(Path ID = 0) is also count into the number, the max available path id is actually N-1. It's weird and doesn't match the description of Max Path ID field inside MAX_PATH frame.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -254,17 +254,17 @@ defined as follows:
 
 - initial_max_path_id (current version uses 0x0f739bbc1b666d09): the
   initial_max_path_id transport parameter is included if the endpoint supports
-  the multipath extension as defined in this document. This is 
+  the multipath extension as defined in this document. This is
   a variable-length integer specifying the maximum path identifier
   an endpoint is willing to maintain at connection initiation.
   E.g. if initial_max_path_id is set to 1, only connection IDs
   associated to Path IDs 0 and 1 should be issued by the peer.
-  If an endpoint receives transport parameter initial_max_path_id with value 0, 
-  it MUST NOT cause any error, and notice that the peer want to 
-  enable the multipath extension without allowing extra paths immediately 
+  If an endpoint receives transport parameter initial_max_path_id with value 0,
+  it MUST NOT cause any error, and notice that the peer want to
+  enable the multipath extension without allowing extra paths immediately
   after the handshake is done. If initial_max_path_id is successfully negotiated
-  (including negotiated as 0), endpoints can still use MAX_PATHS frame to raise up 
-  the limitation for maximum path identifier of the connection in the future. 
+  (including negotiated as 0), endpoints can still use MAX_PATHS frame to raise up
+  the limitation for maximum path identifier of the connection in the future.
 
 Setting initial_max_path_id parameter is equivalent to sending a
 MAX_PATHS frame ({{max-paths-frame}}) with the same value.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1335,10 +1335,9 @@ MAX_PATH_ID Frame {
 MAX_PATH_ID frames contain the following field:
 
 Maximum Path Identifier:
-: A count of the cumulative number of paths that can be opened
-  over the lifetime of the connection. This value MUST NOT exceed 2^32-1, as
-  Path IDs are defined with a maximum value 2^32-1 as the 32 bits of the Path ID are used
-  to calculate the nonce (see {{multipath-aead}}).
+: The maximum path identifier that the sending endpoint is willing to accept. 
+  This value MUST NOT exceed 2^32-1, as Path IDs are defined with a maximum value 2^32-1 
+  as the 32 bits of the Path ID are used to calculate the nonce (see {{multipath-aead}}).
   The Maximum Path Identifier value MUST NOT be lower than the value
   advertised in the initial_max_path_id transport parameter. Receipt
   of an invalid Maximum Path Identifier value MUST be treated as a
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -263,11 +263,11 @@ defined as follows:
   it MUST NOT cause any error, and notice that the peer want to
   enable the multipath extension without allowing extra paths immediately
   after the handshake is done. If initial_max_path_id is successfully negotiated
-  (including negotiated as 0), endpoints can still use MAX_PATHS frame to raise up
+  (including negotiated as 0), endpoints can still use MAX_PATH_ID frame to raise up
   the limitation for maximum path identifier of the connection in the future.
 
 Setting initial_max_path_id parameter is equivalent to sending a
-MAX_PATHS frame ({{max-paths-frame}}) with the same value.
+MAX_PATH_ID frame ({{max-paths-frame}}) with the same value.
 
 If either of the endpoints does not advertise the initial_max_path_id transport
 parameter, then the endpoints MUST NOT use any frame or
@@ -339,7 +339,7 @@ This proposal adds four multipath control frames for path management:
 (see {{path-abandon-frame}}),
 - PATH_STANDBY and PATH_AVAILABLE frames to express a preference
 in path usage (see {{path-standby-frame}} and {{path-available-frame}}), and
-- MAX_PATHS frame for increasing the limit of active paths.
+- MAX_PATH_ID frame for increasing the limit of active paths.
 
 All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
@@ -600,9 +600,9 @@ will not be used anymore. If the path is still active, the peer SHOULD replace
 it with a new connection ID using a MP_NEW_CONNECTION_ID frame.
 
 Endpoints MUST NOT issue new connection IDs with Path IDs greater than
-the Maximum Path Identifier field in MAX_PATHS frames (see Section {{max-paths-frame}}).
+the Maximum Path Identifier field in MAX_PATH_ID frames (see Section {{max-paths-frame}}).
 When an endpoint finds it has not enough available unused path identifiers,
-it SHOULD send a MAX_PATHS frame to inform the peer that it could use new active
+it SHOULD send a MAX_PATH_ID frame to inform the peer that it could use new active
 path identifiers.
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -252,41 +252,39 @@ the use of the multipath extension during the connection handshake,
...
-The initial_max_paths parameter MUST NOT be remembered
+The initial_max_path_id parameter MUST NOT be remembered
 ({{Section 7.4.1 of QUIC-TRANSPORT}}).
 New paths can only be used after handshake completion.
 
 This extension does not change the definition of any transport parameter
 defined in {{Section 18.2. of QUIC-TRANSPORT}}.
 
-The initial_max_paths transport parameter limits the initial maximum number of active paths
-that can be used during a connection.
```
mirjak, line 286: Why is this removed? This is still true.

draft-ietf-quic-multipath.md
```diff
@@ -252,41 +252,39 @@ the use of the multipath extension during the connection handshake,
...
+- initial_max_path_id (current version uses 0x0f739bbc1b666d09): the
+  initial_max_path_id transport parameter is included if the endpoint supports
   the multipath extension as defined in this document. This is
-  a variable-length integer value specifying the maximum number of
-  active concurrent paths an endpoint is willing to maintain.
-  The value of the initial_max_paths parameter MUST be at least 2.
+  a variable-length integer specifying the initial maximum value of
+  available path identifier an endpoint is willing to support.
+  The value of the initial_max_path_id parameter MUST be at least 2.
   An endpoint that receives a value less than 2 MUST close
```
mirjak, line 261:
```suggestion
   a variable-length integer specifying the maximum path identifier
  an endpoint is willing to maintain at connection initiation.
  E.g. if initial_max_path_id is set to 1, only connection IDs associated to Path IDs 0 and 1
  should be issued by the peer.
  The value of the initial_max_path_id parameter MUST be at least 1.
  An endpoint that receives a value less than 1 MUST close
```

draft-ietf-quic-multipath.md
```diff
@@ -252,41 +252,39 @@ the use of the multipath extension during the connection handshake,
...
   the connection with an error of type TRANSPORT_PARAMETER_ERROR.
```
mirjak, line 262: However, I'm not sure if we really need to close the connection if initial_max_path_id=0. I think this would simply mean that you don't allow for any additional path yet.

Yanmei-Liu: If endpoints don't want to allow additional paths, endpoints could choose not to send the transport parameter. 
There's only one extra scenario: Endpoint doesn't want to allow any additional paths when handshake is finished, but keep the possibility to allow for new paths with MAX_PATH_ID frame when necessary. That could be a possible strategy.

Yanmei-Liu: I record this issue in #381 . If people want to add this mechanism, we could add it in a separate PR.

Yanmei-Liu:
I'm convinced by people in issue #381 that we could allow the initial_max_path_id to be set to 0. Updated the PR. 
Thanks to @mirjak @marten-seemann and @huitema.

draft-ietf-quic-multipath.md
```diff
@@ -1368,7 +1366,7 @@ the "QUIC Transport Parameters" registry under the "QUIC Protocol" heading.
 
 Value                                         | Parameter Name.   | Specification
 ----------------------------------------------|-------------------|-----------------
-TBD (current version uses 0x0f739bbc1b666d07) | initial_max_paths | {{nego}}
+TBD (current version uses 0x0f739bbc1b666d07) | initial_max_path_id | {{nego}}
 {: #transport-parameters title="Addition to QUIC Transport Parameters Entries"}
```
huitema, line 1378: I think we should change the parameter ID. The previous value meant "max path = 1 means path_id < 1"; we are changing that to "max path = 1 means path_id <= 1". That's an incompatible change, thus we need a new parameter ID value.

draft-ietf-quic-multipath.md
```diff
@@ -1047,7 +1045,7 @@ it MUST close the connection with an error of type FRAME_ENCODING_ERROR.
 
 Receipt of multipath-specific frames
 that use a Path ID that is greater than the announced Maximum Paths value
-in the MAX_PATHS frame or in the initial_max_paths transport parameter,
+in the MAX_PATHS frame or in the initial_max_path_id transport parameter,
 if no MAX_PATHS frame was received yet,
 MUST be treated as a connection error of type MP_PROTOCOL_VIOLATION.
```
huitema, line 1058: To be consistent, we should we also change the name of that frame to MAX_PATH_ID and call the variable the "Maximum Path ID".

huitema (commented):
If we change the transport parameter name and behavior, we need to change the parameter ID as well. We probably also need to rename the MAX_PATHS frame to MAX_PATH_ID, and change the text accordingly. And get a new frame ID for the MAX_PATH_ID frame.

draft-ietf-quic-multipath.md
Yanmei-Liu: It's true. I've changed the parameter ID from 0x0f739bbc1b666d07 to 0x0f739bbc1b666d09.

draft-ietf-quic-multipath.md
Yanmei-Liu: That's true. I updated the Frame Name and Type from MAX_PATHS to MAX_PATH_ID, please review.

Yanmei-Liu: Note that we only need to change Frame Type and Frame Name in this PR, because Martin has already changed the definition of Frame field "Maximum Path ID" in PR #372.

draft-ietf-quic-multipath.md
```diff
@@ -1308,38 +1314,38 @@ As the MP_NEW_CONNECTION_ID frames applies the sequence number per path,
...
 
-MAX_PATHS frames are formatted as shown in {{fig-max-paths-frame-format}}.
+MAX_PATH_ID frames are formatted as shown in {{fig-max-paths-frame-format}}.
 
 ~~~
-MAX_PATHS Frame {
-  Type (i) = 0x15228c0b,
+MAX_PATH_ID Frame {
+  Type (i) = 0x15228c0c,
   Maximum Paths (i),
```
martinthomson, line 1330: This probably needs to be "Maximum Path ID" (and the same throughout)

draft-ietf-quic-multipath.md
```diff
@@ -1308,42 +1314,43 @@ As the MP_NEW_CONNECTION_ID frames applies the sequence number per path,
...
 ~~~
-{: #fig-max-paths-frame-format title="MAX_PATHS Frame Format"}
+{: #fig-max-paths-frame-format title="MAX_PATH_ID Frame Format"}
 
-MAX_PATHS frames contain the following field:
+MAX_PATH_ID frames contain the following field:
 
 Maximum Path Identifier:
 : A count of the cumulative number of paths that can be opened
   over the lifetime of the connection. This value MUST NOT exceed 2^32-1, as
```
martinthomson, line 1339: This definition might be better stated as:

> The maximum path identifier that the sending endpoint is willing to accept.

draft-ietf-quic-multipath.md
```diff
@@ -1308,42 +1314,43 @@ As the MP_NEW_CONNECTION_ID frames applies the sequence number per path,
...
   Path IDs are defined with a maximum value 2^32-1 as the 32 bits of the Path ID are used
-  to calculate the nonce (see Section {{multipath-aead}}).
+  to calculate the nonce (see {{multipath-aead}}).
   The Maximum Paths value MUST NOT be lower than the value
```
martinthomson, line 1342:
```suggestion
  The Maximum Path Identifier value MUST NOT be lower than the value
```

draft-ietf-quic-multipath.md
```diff
@@ -1308,42 +1314,43 @@ As the MP_NEW_CONNECTION_ID frames applies the sequence number per path,
...
-  advertised in the initial_max_paths transport parameter. Receipt
-  of an invalid Maximum Paths value MUST be treated as a
+  advertised in the initial_max_path_id transport parameter. Receipt
```
martinthomson, line 1343: ... or what?  Is it a frame error to receive a MAX_PATH_ID with a value less than (or equal to...?) the transport parameter?

martinthomson (approved):
This seems good.

draft-ietf-quic-multipath.md
Yanmei-Liu: That's true. I've updated the definition of maximum path identifier field.

draft-ietf-quic-multipath.md
```diff
@@ -252,41 +252,47 @@ the use of the multipath extension during the connection handshake,
...
-Setting initial_max_paths parameter is equivalent to sending a
-MAX_PATHS frame ({{max-paths-frame}}) with the same value.
-
-If either of the endpoints does not advertise the initial_max_paths transport
+  a variable-length integer specifying the maximum path identifier
+  an endpoint is willing to maintain at connection initiation.
+  For example, if initial_max_path_id is set to 1, only connection IDs
+  associated with Path IDs 0 and 1 should be issued by the peer.
+  If an endpoint receives an initial_max_path_id transport parameter with value 0,
+  it MUST NOT cause any error, and notice that the peer want to
```
mirjak, line 263:
```suggestion
the peer want to
```

draft-ietf-quic-multipath.md
```diff
@@ -252,41 +252,47 @@ the use of the multipath extension during the connection handshake,
...
+  enable the multipath extension without allowing extra paths immediately
+  after the handshake is done. If initial_max_path_id is successfully negotiated
+  (including negotiated as 0), endpoints can use MAX_PATH_ID frame to increase
+  the maximum path identifier and enable the use of new paths.
```
mirjak, line 267:
```suggestion
the peer aims to  enable the multipath extension without allowing extra paths immediately.
To allow for the use of more paths later,
endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
```

huitema (approved):
I think this is ready.

draft-ietf-quic-multipath.md
```diff
@@ -252,41 +252,45 @@ the use of the multipath extension during the connection handshake,
...
-
-If either of the endpoints does not advertise the initial_max_paths transport
+  a variable-length integer specifying the maximum path identifier
+  an endpoint is willing to maintain at connection initiation.
+  For example, if initial_max_path_id is set to 1, only connection IDs
+  associated with Path IDs 0 and 1 should be issued by the peer.
+  If an endpoint receives an initial_max_path_id transport parameter with value 0,
+the peer aims to  enable the multipath extension without allowing extra paths immediately.
+To allow for the use of more paths later,
+endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
```
qdeconinck, line 265:
```suggestion
  the peer aims to  enable the multipath extension without allowing extra paths immediately.
  To allow for the use of more paths later,
  endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
```

qdeconinck (approved):
Modulo a formatting issue, this PR is ready to merge.

Issue #313, opened 2024-03-16
When to CID expire if using Unique Path ID?


huitema:
In the Unique Path ID model, the creation of a CID is always linked to the creation of a potential path. CIDs are managed using two frames:

* MP_NEW_CONNECTION_ID with parameters CID value, path ID, sequence, and retire prior.
* MP_RETIRE_CID with parameters path ID and sequence.

Path IDs are not supposed to live forever. If a path is abandoned, the path ID will not be reused, because of potential ambiguity: reusing path id and sequence numbers could break encryption; when packets arrive for an abandoned path ID and the path state has been delete, there is no way to distinguish new packets from copies of old packets. Yet, as long as there is a valid CID available, nothing prevents a node from using it and sending packets with that corresponding path_id.

The sender of the MP_NEW_CONNECTION_ID can only force retirement of old path IDs by sending an MP_NEW_CONNECTION_ID frame and setting the "retire prior" value, but doing so creates a new CID for the old path and does not actually "retire all CID for the path".

In the current specification, these "old path CID" can only be retired by the peer, who could send series of MP_RETIRE_CID frames. The local node knows that the path is abandoned, so it could refrain from providing new CID to replace them. But nothing forces the peer to do that.

We miss a way to say "retire all CID for that path". One way would be to state that as a side effect of "PATH_ABANDON": the peer should immediately retire all CIDs tied to that path. We should also specify that a node MUST retain knowledge of all valid CID until they have been retired -- otherwise we get issue like a peer sending packets with old CID values, resulting in a stateless reset. Instead, the node that receive a packet with an old CID for an abandoned path should probably just ignore it. 

Maybe there are simpler solutions, but the current behavior is under-specified.

gloinul:
To my understanding is that how it is currently specified is that each endpoint need to close the path in the direction towards  itself by first sending the path_abandon then wait the 3*PTO, then send the retire CID for all the CIDs used by that path. And if the peer don't send a path_abandon back to you when you have sent one, this results in a path that is open in only one direction. That is probably not what we want the situation to be in.

gloinul:
I would suggest that we ensure that we have when it comes to the path close scenarion, that one says first stop sending, i.e. path_abandon. Then a path_close that retires the path in this direction and all related CIDs. This could actually be a frame that says path_closed which references a path Index. 

[...]

PR #377, opened 2024-06-04
Rewrite the path close section to specify symmetric Abandon

huitema:
As discussed during the interim. This solves issue #313.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -449,17 +449,23 @@ of local resources. It is also possible to abandon a path for which no
 packet has been sent (see {{abandon-early}}).
 
 When an endpoint receives a PATH_ABANDON frame, it MUST send a corresponding
-PATH_ABANDON frame if it has not already done so, and it MUST treat all
+PATH_ABANDON frame if it has not already done so. It MUST stop sending
+any new packet on the abandoned path, and it MUST treat all
 connection identifiers received from the peer for that path as immediately
-retired. However, the endpoint should retain some knowledge of the
-connection identifiers sent to the peer for a short period,
-as specified in {{spurious-stateless-reset}}.
+retired. However, knowledge of the
+connection identifiers received from the peer and of the state
+of the number space associated to the path SHOULD be retained while
+packets from the peer might still be in transit, i.e., for a delay of
+3 PTO after the PATH_ABANDON frame has been received from the peer,
+both to avoid generating spurious stateless packets as specified in
+{{spurious-stateless-reset}} and to be able to acknowledge the
+last packets received from the peer as specified in {{ack-after-abandon}}).
 
 After receiving or sending a PATH_ABANDON frame, the endpoints SHOULD
-promptly send ACK_MP frames to acknowledge all packets received on
+promptly send MP_ACK frames to acknowledge all packets received on
 the path and not yet acknowledged, as specified in {{ack-after-abandon}}).
 When an endpoint finally deletes all resource associated with the path,
-the packets sent over the path and not yet acknowledged MUST be considered lost.
+the packets sent over the path and not yet acknowledged MUST be considered lost. 
 
 After a path is abandoned, the Path ID MUST NOT be reused
 for new paths, as the Path ID is part of the nonce calculation {{multipath-aead}}.
@@ -493,7 +499,8 @@ well arrive after the PATH_ABANDON frames have been sent or received.
 If these packets arrive after the connection identifiers sent to the peer
 have been retired, they will not be recognizes as bound for the local
 connection and could trigger the peer to send a Stateless Reset
-packet. The rule to "retain knowledge of connection ID for a short period"
+packet. The rule to "retain knowledge of connection ID for 3 PTO
+after receiving a PATH_ABANDON"
 is intended to reduce the risk of sending such spurious stateless
 packets, but it cannot completely avoid that risk.
 
@@ -501,19 +508,19 @@ The immediate retirement of connection identifiers received for the
 path guarantees that spurious stateless reset packets
 sent by the peer will not cause the closure of the QUIC connection.
 
-### Handling ACK_MP for abandoned paths {#ack-after-abandon}
+### Handling MP_ACK for abandoned paths {#ack-after-abandon}
 
 When an endpoint decides to send a PATH_ABANDON frame, there may
 still be some unacknowledged packets. Some other packets may well
 be in transit, and could be received shortly after sending the
 PATH_ABANDON frame. As specified above, the endpoints SHOULD
-send ACK_MP frames promptly, to avoid unnecessary data
+send MP_ACK frames promptly, to avoid unnecessary data
 retransmission after the peer deletes path resources.
 
-These ACK_MP frames SHOULD be send on a different path than the
+These MP_ACK frames SHOULD be send on a different path than the
 path being abandoned.
 
-ACK_MP frames received after the endpoint has entirely deleted
+MP_ACK frames received after the endpoint has entirely deleted
 a path MUST be silently discarded.
 
 ### Idle Timeout {#idle-time-close}
@@ -549,11 +556,11 @@ MUST do so explicitly by sending a PATH_ABANDON frame, as defined in
 
 ### Early Abandon {#abandon-early}
 
-There are scenarios in which an endpoint will receive a PATH_ABANDON frame
+The are scenarios in which an endpoint will receive a PATH_ABANDON frame
 before receiving or sending any traffic on a path. For example, if the client
-tries to initiate a path and the path cannot be established, it will send a
+tries to initiate a path and the path cannot be establish, it will send a
 PATH_ABANDON frame (see {{path-initiation}}). An endpoint may also decide
-to abandon a path for any reason, for example, removing a hole from
+to abandon a path for any reason, such as for example removing a hole from
 the sequence of path IDs in use. This is not an error. The endpoint that
 receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
 
@@ -563,7 +570,7 @@ An endpoint may deny the establishment of a new path initiated by its
 peer during the address validation procedure. According to {{QUIC-TRANSPORT}},
 the standard way to deny the establishment of a path is to not send a
 PATH_RESPONSE in response to the peer's PATH_CHALLENGE.
-
+ 
 ## Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
 
 With the multipath extension, each connection ID is associated with one path
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -662,43 +662,6 @@ on the path that the retired connection ID was used on but has
 to use a different connection ID for the same Path ID when doing so.
 
 
-# Multipath Operation with Multiple Packet Number Spaces
-
-The QUIC multipath extension uses different packet number spaces for each path.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -537,25 +538,31 @@ can send ack-eliciting packets such as packets containing PING frames
...
+MUST do so explicitly by sending a PATH_ABANDON frame, as defined in
+{{path-close}}.
+
+### Early Abandon {#abandon-early}
+
+The are are scenarios in which an endpoint will receive a PATH_ABANDON frame
+before receiving or sending any traffic on a path. For example, if the client
+tries to initiate a path and the path cannot be establish, it will send a
+PATH_ABANDON frame (see {{path-initiation}}). An endpoint may also decide
+to abandon a path for any reason, such as for example removing a hole from
```
Yanmei-Liu, line 556:
```suggestion
to abandon a path for any reason, for example, removing a hole from
```

Yanmei-Liu (commented):
Thanks a lot for this PR! Some minor fixes.

huitema:
@Yanmei-Liu thanks for the review and the fixes!

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -515,6 +480,42 @@ the server MAY wait for a short, limited time such as one PTO if a path
...
+### Handling ACK_MP for abandoned paths {#ack-after-abandon}
+
+When an endpoint decides to send a PATH_ABANDON frame, there may
+still be some unacknowledged packets. Some other packets may well
+be in transit, and could be received shortly after sending the
+PATH_ABANDON frame. As specified above, the endpoints SHOULD
+send ACK_MP frames promptly, to avoid unnecessary data
+retransmission after the peer deletes path resources.
+
+These ACK_MP frames SHOULD be send on a different path than the
```
mirjak, line 513: I think this is a MUST. Or no normatively language here at all. I would rather say somewhere else that if the PATH_ABANDON frame was sent, the endpoint MUST NOT send on that path anymore.

huitema: Yes, that's probably the best way to handle it, but there is the slight issue of abandoning the last path.

draft-ietf-quic-multipath.md
```diff
@@ -438,62 +437,29 @@ connectivity or local preferences. After an endpoint abandons
...
+(see {(idle-time-close}}) or for any other reason, such as management
+of local resources. It is also possible to abandon a path for which no
+packet has been sent (see {{abandon-early}}).
+
+When an endpoint receives a PATH_ABANDON frame, it MUST send a corresponding
+PATH_ABANDON frame if it has not already done so, and it MUST treat all
+connection identifiers received from the peer for that path as immediately
+retired. However, the endpoint should retain some knowledge of the
+connection identifiers sent to the peer for a short period,
+as specified in {{spurious-stateless-reset}}.
```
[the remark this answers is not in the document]
huitema: I tried to develop the ACK_MP issue in the subsection "Handling ACK_MP for abandoned paths {#ack-after-abandon}". Maybe we should call that out here, in the same way that we call out stateless reset.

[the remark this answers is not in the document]
huitema: I also think that we have a hard rule on the data sender side: "if a path info is deleted, then all pending packets MUST be considered lost", and softer rule "don't delete the info too soon because you will end up with spurious retransmit". On the data receiver side, we could have a hard rule, "you MUST send an ACK_MP for all packets received before abandon by the peer", and a softer rule "you should wait a bit for packets to arrive after abandon so you can ACK them and avoid forcing the peer to resend them". Maybe develop that a bit more in the dedicated subsection.

draft-ietf-quic-multipath.md
mirjak: For the last path you should do connection close instead. That’s already written somewhere but it might be a SHOULD only. If so, we can maybe change to a MUST.

huitema: Not sure it matters. Suppose that we treat "abandon of last path" as a protocol error. That means the peer receiving the abandon will automatically issue "connection close" with reason protocol error, and the peer sending the abandon will send nothing, because it must not send data after abandon. But yes, sending connection close is more graceful.

draft-ietf-quic-multipath.md
```diff
@@ -515,6 +480,42 @@ the server MAY wait for a short, limited time such as one PTO if a path
...
+connection and could trigger the peer to send a Stateless Reset
+packet. The rule to "retain knowledge of connection ID for a short period"
+is intended to reduce the risk of sending such spurious stateless
+packets, but it cannot completely avoid that risk.
+
+The immediate retirement of connection identifiers received for the
+path guarantees that spurious stateless reset packets
+sent by the peer will not cause the closure of the QUIC connection.
+
+### Handling MP_ACK for abandoned paths {#ack-after-abandon}
```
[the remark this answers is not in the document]
huitema: Sorry, updates lost in a github snafu. Resetting.

draft-ietf-quic-multipath.md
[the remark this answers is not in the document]
huitema: @mirjak I think that the "tighten" commit fixes this. Please review.

draft-ietf-quic-multipath.md
huitema: Replace SHOULD by MUST.

draft-ietf-quic-multipath.md
```diff
@@ -361,11 +361,10 @@ unless it has previously validated that address.
 
 If validation succeeds, the client can continue to use the path.
 If validation fails, the client MUST NOT use the path and can
-remove any status associated to the path initation attempt.
+remove any status associated to the path initiation attempt.
 However, as the used Path ID is anyway consumed,
-and the endpoint MUST abandon the path by sending a PATH_ABANDON frame
-and retire all corresponding connection IDs by sending MP_RETIRE_CONNECTION_ID frames
-on another path to inform the peer that the Path ID cannot be used anymore.
+and the endpoint MUST explicitly close the path as specified in
```
qdeconinck, line 366:
```suggestion
the endpoint MUST explicitly close the path as specified in
```

draft-ietf-quic-multipath.md
```diff
@@ -515,6 +486,43 @@ the server MAY wait for a short, limited time such as one PTO if a path
...
+The peers that send a PATH_ABANDON frame MUST treat all connection
+identifiers received from the peer for the path ID as immediately
+retired. The Stateless Reset Tokens associated with these connection
+identifiers MUST NOT be used to identify Stateless Reset packets
+per {{Section 10.3 of QUIC-TRANSPORT}}.
+
+Due to packet losses and network delays, packets sent on the path may
+well arrive after the PATH_ABANDON frames have been sent or received.
+If these packets arrive after the connection identifiers sent to the peer
+have been retired, they will not be recognizes as bound for the local
```
qdeconinck, line 500:
```suggestion
have been retired, they will not be recognized as bound for the local
```

qdeconinck (approved):
Two typos, but looks good to me

draft-ietf-quic-multipath.md
```diff
@@ -558,25 +568,31 @@ can send ack-eliciting packets such as packets containing PING frames
...
-This means if all paths remain idle for the idle timeout, the connection
-is implicitly closed.
-
 Server implementations need to select the sub-path idle timeout as a trade-
 off between keeping resources, such as connection IDs, in use
 for an excessive time or having to promptly re-establish a path
 after a spurious estimate of path abandonment by the client.
 
+Endpoints that desire to close a path because of the idle timer rule
+MUST do so explicitly by sending a PATH_ABANDON frame, as defined in
```
mirjak, line 577:
```suggestion
MUST do so explicitly by sending a PATH_ABANDON frame on another active path, as defined in
```

draft-ietf-quic-multipath.md
```diff
@@ -645,75 +661,8 @@ The peer that sends the MP_RETIRE_CONNECTION_ID frame can keep sending data
...
-
-"Closing" state is entered after an PATH_ABANDON frame was sent or received.
-In this state, the endpoint waits for three PTOs before sending MP_RETIRE_CONNECTION_ID frames.
-This allows for graceful tear down and processing of in-flight packets.
-
-When a path reaches the "Closed" state, the endpoint releases all the
-path's associated resources, including the associated connection IDs and the path identifier.
 
-# Packet Protection {#multipath-aead}
+## Packet Protection {#multipath-aead}
```
mirjak, line 665:
```suggestion
# Packet Protection {#multipath-aead}
```

draft-ietf-quic-multipath.md
Yanmei-Liu: I think it's better to allow endpoints to send PATH_ABANDON frame either on the same path or another active path. This depends on whether the same path is a blackhole. So we don't need to specify here.

draft-ietf-quic-multipath.md
```diff
@@ -461,60 +460,35 @@ connectivity or local preferences. After an endpoint abandons
...
-Reception or sending of the PATH_ABANDON frame is
-the first step to release all resources related to a
-Path ID. However, the Path ID can only be released after all active
-connection IDs for the Path ID have been retired or timed-out after
-the PATH_ABANDON frame was sent.
-Still, when an endpoint receives an PATH_ABANDON frame,
-it SHOULD NOT use the associated Path ID in future frames, except
-in ACK_MP frames for acknowledging inflight packets and
-in MP_RETIRE_CONNECTION_ID frames for connection ID retirement.
+does not indicate path abandonment (see {{retire-cid-close}}).
```
mirjak, line 467:
```suggestion
does not indicate path abandonment (see further {{consume-retire-cid}}).
```

draft-ietf-quic-multipath.md
```diff
@@ -932,7 +881,7 @@ for example the PTO timeout: if an ACK_MP is not received after more
 than 350ms, either the data packet or its ACK_MP were probably lost.
 
 The simplest implementation is to compute smoothedRTT and RTTvar per
-{{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which MP_ACKs are
+{{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which ACK_MPs are
```
mirjak, line 884:
```suggestion
{{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which ACK_MP frames are
```

draft-ietf-quic-multipath.md
```diff
@@ -461,60 +460,35 @@ connectivity or local preferences. After an endpoint abandons
...
 connection ID retirement, the inverse is not true: retiring the associated connection IDs
 does not indicate path abandonment (see further {{consume-retire-cid}}).
-Implicit signals such as idle time or packet losses might be
-the only way for an endpoint to detect path closure (see {(idle-time-close}})
-if connectivity is broken on that path.
-
-Note that other explicit closing mechanisms of {{QUIC-TRANSPORT}} still
-apply on the whole connection. In particular, the reception of either a
-CONNECTION_CLOSE ({{Section 10.2 of QUIC-TRANSPORT}}) or a Stateless
-Reset ({{Section 10.3 of QUIC-TRANSPORT}}) closes the connection.
```
mirjak, line 476: @huitema why did you remove this paragraph?

huitema: Can't remember why. Restore it if you want.

PR #382, opened 2024-06-17
Rewrite outdated examples in Section 6

Yanmei-Liu:
According to PR #377 and PR #375, the examples in [Section 6](https://datatracker.ietf.org/doc/html/draft-ietf-quic-multipath-08#section-6) is outdated. Try to update the figures and keep examples compatible.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -829,8 +829,8 @@ Path ID 1), and server provides two connection IDs
 Before the client opens a new path by sending a packet on that path
 with a PATH_CHALLENGE frame, it has to check whether there is
 an unused connection IDs for the same unused Path ID available for each side.
-Endpoints SHOULD consume the smallest unused Path ID available
-when creating the new path as specified in {{consume-retire-cid}}.
+In this example the Path ID 1 is used which is the smallest unused Path ID available
+as recommended in {{consume-retire-cid}}.
 Respectively, the client chooses the connection ID S1
 as the Destination Connection ID of the new path.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -856,9 +856,6 @@ Client                                                      Server
 ~~~
 {: #fig-example-path-close1 title="Example of closing a path."}
 
-Note that if endpoints have other available paths, endpoints are also
-able to send the PATH_ABANDON frame on another available path. This mechanism
-can be used for closing blackhole paths.
 
 
 # Implementation Considerations
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -840,14 +840,9 @@ as the Destination Connection ID of the new path.
 {{fig-example-path-close1}} illustrates an example of path closure.
 
 In this example, the client wants to close the path with Path ID 1.
-It sends the PATH_ABANDON frame to terminate the path. After received
-the PATH_ABANDON frame containing Path ID 1, the server decide to close
-the path with sending PATH_ABANDON frame either. During this time,
-if endpoints still receive inflight packets except the acknowledgements,
-endpoints only sends PATH_ABANDON frame as reply or choose to ignore the inflight packets.
-After a short period of time (3 PTOs as specified in {#path-close}),
-both endpoints retire and free all the resources associated with
-the closed Path ID, and the closed Path ID MUST NOT be reused in the future.
+It sends the PATH_ABANDON frame to terminate the path. After receiving
+the PATH_ABANDON frame with Path ID 1, the server also send a 
+PATH_ABANDON frame with Path ID 1. 
 
 ~~~
 Client                                                      Server
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -853,8 +853,6 @@ Client                                                      Server
                     <-1-RTT[Y]: DCID=C1 PATH_ABANDON[Path ID=1],
                                            ACK_MP[PATH ID=1, PN=X]
 1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
-(After 3 PTOs, client retires the corresponding CID)
-              (After 3 PTOs, server retires the corresponding CID)
 ~~~
 {: #fig-example-path-close1 title="Example of closing a path."}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -795,9 +795,10 @@ should not cause linkability issue.
 {{fig-example-new-path}} illustrates an example of new path establishment
 using multiple packet number spaces.
 
-In the handshake negotiation of the example flow, Client and Server
-successfully negotiate the initial_max_path_id value as 2, which means
-both endpoints could use Path ID 0, 1, and 2.
+In this example it is assume that both endpoint have
+indicated a initial_max_path_id value as at least 2, which means
+both endpoints can use Path IDs 0, 1, and 2. Note that
+Path ID 0 is already used for the initial path.
 
 ~~~
    Client                                                  Server
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -857,7 +857,6 @@ Client                                                      Server
 {: #fig-example-path-close1 title="Example of closing a path."}
 
 
-
 # Implementation Considerations
 
 ## Number Spaces
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -841,8 +841,8 @@ as the Destination Connection ID of the new path.
 
 In this example, the client wants to close the path with Path ID 1.
 It sends the PATH_ABANDON frame to terminate the path. After receiving
-the PATH_ABANDON frame with Path ID 1, the server also send a 
-PATH_ABANDON frame with Path ID 1. 
+the PATH_ABANDON frame with Path ID 1, the server also send a
+PATH_ABANDON frame with Path ID 1.
 
 ~~~
 Client                                                      Server
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -852,9 +852,11 @@ Client                                                      Server
                            (server tells client to abandon a path)
                     <-1-RTT[Y]: DCID=C1 PATH_ABANDON[Path ID=1],
                                            ACK_MP[PATH ID=1, PN=X]
-1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -795,6 +795,10 @@ should not cause linkability issue.
 {{fig-example-new-path}} illustrates an example of new path establishment
 using multiple packet number spaces.
 
+In the handshake negotiation of the example flow, Client and Server
+successfully negotiate the initial_max_path_id value as 2, which means
+both endpoints could use Path ID 0, 1, and 2.
```
mirjak, line 800:
```suggestion
In this example it is assume that both endpoint have
indicated a initial_max_path_id value as at least 2, which means
both endpoints can use Path IDs 0, 1, and 2. Note that
Path ID 0 is already used for the initial path.
```

draft-ietf-quic-multipath.md
```diff
@@ -824,49 +828,44 @@ Path ID 1), and server provides two connection IDs
 Before the client opens a new path by sending a packet on that path
 with a PATH_CHALLENGE frame, it has to check whether there is
 an unused connection IDs for the same unused Path ID available for each side.
+Endpoints SHOULD consume the smallest unused Path ID available
+when creating the new path as specified in {{consume-retire-cid}}.
```
mirjak, line 832:
```suggestion
In this example the Path ID 1 is used which is the smallest unused Path ID available
as recommended in {{consume-retire-cid}}.
```
Note that we really should not use any normative language in the example section. And also note that not every "should" needs to be a normative "SHOULD". It is often practice to avoid lower cases "should" and use e.g. "can" instead to avoid confusion.

draft-ietf-quic-multipath.md
```diff
@@ -824,49 +828,44 @@ Path ID 1), and server provides two connection IDs
...
+
+In this example, the client wants to close the path with Path ID 1.
+It sends the PATH_ABANDON frame to terminate the path. After received
+the PATH_ABANDON frame containing Path ID 1, the server decide to close
+the path with sending PATH_ABANDON frame either. During this time,
+if endpoints still receive inflight packets except the acknowledgements,
+endpoints only sends PATH_ABANDON frame as reply or choose to ignore the inflight packets.
+After a short period of time (3 PTOs as specified in {#path-close}),
+both endpoints retire and free all the resources associated with
+the closed Path ID, and the closed Path ID MUST NOT be reused in the future.
```
mirjak, line 849:
```suggestion
In this example, the client wants to close the path with Path ID 1.
It sends the PATH_ABANDON frame to terminate the path. After receiving
the PATH_ABANDON frame with Path ID 1, the server also send a 
PATH_ABANDON frame with Path ID 1. 
```
Do not repeat protocol requirements here. Only explain the example.

draft-ietf-quic-multipath.md
```diff
@@ -824,49 +828,44 @@ Path ID 1), and server provides two connection IDs
...
+                                           ACK_MP[PATH ID=1, PN=X]
+1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
+(After 3 PTOs, client retires the corresponding CID)
+              (After 3 PTOs, server retires the corresponding CID)
 ~~~
 {: #fig-example-path-close1 title="Example of closing a path."}
 
+Note that if endpoints have other available paths, endpoints are also
+able to send the PATH_ABANDON frame on another available path. This mechanism
+can be used for closing blackhole paths.
```
mirjak, line 867:
```suggestion
```
Again don't repeat the protocol mechanism here. Only example the example. As this is not shown in the example this is not needed. We could add another example with sends the abandon frame on a different path if we want to explain that part as well.

draft-ietf-quic-multipath.md
```diff
@@ -824,49 +828,44 @@ Path ID 1), and server provides two connection IDs
...
-1-RTT[U]: DCID=S2 MP_RETIRE_CONNECTION_ID[Path ID=1, Seq=0],
-ACK_MP[Path ID=1, PN=Y] ->
-                            (server retires the corresponding CID)
-           <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[Path ID1, Seq=0],
-                                                ACK_MP[Path ID=1, PN=U]
+                    <-1-RTT[Y]: DCID=C1 PATH_ABANDON[Path ID=1],
+                                           ACK_MP[PATH ID=1, PN=X]
+1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
+(After 3 PTOs, client retires the corresponding CID)
+              (After 3 PTOs, server retires the corresponding CID)
```
mirjak, line 861:
```suggestion
```
As this is not connected to any protocol actions, I don't think this is needed here anymore.

draft-ietf-quic-multipath.md
```diff
@@ -824,49 +828,44 @@ Path ID 1), and server provides two connection IDs
...
+After a short period of time (3 PTOs as specified in {#path-close}),
+both endpoints retire and free all the resources associated with
+the closed Path ID, and the closed Path ID MUST NOT be reused in the future.
 
 ~~~
 Client                                                      Server
 
-(client tells server to abandon a path)
-1-RTT[X]: DCID=S2 PATH_ABANDON[Path ID=1]->
+(client tells server to abandon a path with Path ID=1)
```
mirjak, line 854:
```suggestion
(client tells server to abandon a path with Path ID 1)
```

draft-ietf-quic-multipath.md
```diff
@@ -824,50 +829,35 @@ Path ID 1), and server provides two connection IDs
...
-           <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[Path ID1, Seq=0],
-                                                ACK_MP[Path ID=1, PN=U]
+                    <-1-RTT[Y]: DCID=C1 PATH_ABANDON[Path ID=1],
+                                           ACK_MP[PATH ID=1, PN=X]
+1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
 ~~~
 {: #fig-example-path-close1 title="Example of closing a path."}
 
 
+
```
mirjak:
```suggestion
```

draft-ietf-quic-multipath.md
```diff
@@ -795,6 +795,11 @@ should not cause linkability issue.
 {{fig-example-new-path}} illustrates an example of new path establishment
 using multiple packet number spaces.
 
+In this example it is assume that both endpoint have
```
qdeconinck, line 798:
```suggestion
In this example it is assumed that both endpoint have
```

draft-ietf-quic-multipath.md
```diff
@@ -795,6 +795,11 @@ should not cause linkability issue.
...
+indicated a initial_max_path_id value as at least 2, which means
```
qdeconinck, line 799:
```suggestion
indicated an initial_max_path_id value of at least 2, which means
```

draft-ietf-quic-multipath.md
```diff
@@ -824,46 +829,30 @@ Path ID 1), and server provides two connection IDs
...
+the PATH_ABANDON frame with Path ID 1, the server also send a
+PATH_ABANDON frame with Path ID 1.
 
 ~~~
 Client                                                      Server
 
-(client tells server to abandon a path)
-1-RTT[X]: DCID=S2 PATH_ABANDON[Path ID=1]->
+(client tells server to abandon a path with Path ID 1)
+1-RTT[X]: DCID=S1 PATH_ABANDON[Path ID=1]->
```
qdeconinck, line 851: But from the above CID negotiation, S1 belongs to Path ID 1. Should we advertise the PATH_ABANDON on the path we want to close? I do not think this is invalid, but... (see below)

draft-ietf-quic-multipath.md
```diff
@@ -824,46 +829,30 @@ Path ID 1), and server provides two connection IDs
...
-                                               ACK_MP[PATH ID=1, PN=X]
-(client retires the corresponding CID)
-1-RTT[U]: DCID=S2 MP_RETIRE_CONNECTION_ID[Path ID=1, Seq=0],
-ACK_MP[Path ID=1, PN=Y] ->
-                            (server retires the corresponding CID)
-           <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[Path ID1, Seq=0],
-                                                ACK_MP[Path ID=1, PN=U]
+                    <-1-RTT[Y]: DCID=C1 PATH_ABANDON[Path ID=1],
+                                           ACK_MP[PATH ID=1, PN=X]
+1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
```
qdeconinck, line 855: While the server sends the PATH_ABANDON on C1 (which belongs to Path ID 1), which may be fine, I am a bit more skeptical about the ACK_MP sent on S1, while the client previously abandoned the Path ID 1...

draft-ietf-quic-multipath.md
```diff
@@ -807,10 +812,10 @@ using multiple packet number spaces.
    1-RTT[0]: DCID=S1, PATH_CHALLENGE[X] -->
```
qdeconinck, line 812: I can't add a suggestion a few line above, but I would state that there are DCIDs C0 and S0, both belonging to Path ID 0, and showing them (like the packets containing MP_NEW_CONNECTION_ID are probably sent using these DCIDs).

qdeconinck (changes requested):
I am not sure the example is 100% clear, nor the best example neither, as we abandon Path ID 1 using the path with ID 1, and continue using the Path ID 1 to acknowledge packets. While I do not think this is fully invalid, the example would probably benefit from showing that we could send such messages on Path ID 0, for instance.

draft-ietf-quic-multipath.md
```diff
@@ -824,46 +829,30 @@ Path ID 1), and server provides two connection IDs
...
-1-RTT[U]: DCID=S2 MP_RETIRE_CONNECTION_ID[Path ID=1, Seq=0],
-ACK_MP[Path ID=1, PN=Y] ->
-                            (server retires the corresponding CID)
-           <- 1-RTT[V]: DCID=C2 RETIRE_CONNECTION_ID[Path ID1, Seq=0],
-                                                ACK_MP[Path ID=1, PN=U]
+                    <-1-RTT[Y]: DCID=C1 PATH_ABANDON[Path ID=1],
+                                           ACK_MP[PATH ID=1, PN=X]
+1-RTT[U]: DCID=S1 ACK_MP[Path ID=1, PN=Y] ->
 ~~~
 {: #fig-example-path-close1 title="Example of closing a path."}
```
mirjak, line 857:
```suggestion
1-RTT[U]: DCID=S2 ACK_MP[Path ID=1, PN=Y] ->
~~~
{: #fig-example-path-close1 title="Example of closing a path."}
```

Note that the last acknowledgement needs to be send on a different path. This examples assumes another path which uses connection ID S2 exists.

PR #385, opened 2024-06-28
Rename ACK_MP frame to MP_ACK

### LINKED ISSUE #359: Naming consistency of frames
We should maybe review the names of the frames, e.g.:

1) Should we rename ACK_MP to MP_ACK? Or PATH_ACK?

2) Is it MP_NEW_CONNECTION_ID or maybe rather PATH_NEW_CONNECTION_ID or NEW_PATH_CONNECTION_ID or go shorter to PATH_NEW_CID or something?

3) Depending on the final definition is could be MAX_PATHS or MAX_PATH_ID

qdeconinck:
Fix #359.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -195,7 +195,7 @@ to a packet number space.
 The same Path ID is used in both directions to
 address a path in the new multipath control frames,
 such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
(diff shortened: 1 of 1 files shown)


mirjak (approved):

Yanmei-Liu (approved):

huitema (commented):
Why exactly are we doing that? ACK_MP, MP_ACK, there is no reason to pick one over the other. If we must change, I would propose PATH_ACK, to align with other frames like PATH_ABANDON or PATH_AVAILABLE.

huitema (approved):

PR #388, opened 2024-07-01
Clarify denying a path and move text

### LINKED ISSUE #362: Clarify that "refusing a path" means "closing a path"
We have section on "[Refusing a New Path](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#name-refusing-a-new-path)" and it says this:

> An endpoint may deny the establishment of a new path initiated by its
peer during the address validation procedure. According to
{{QUIC-TRANSPORT}}, the standard way to deny the establishment of a path
is to not send a PATH_RESPONSE in response to the peer's PATH_CHALLENGE.

However, I don't think RFC9000 talks about refusing a path anywhere. Or can someone point me at the right section?

mirjak:
fixes #362

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -381,10 +381,12 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
-An endpoint MAY deny the establishment of a new path initiated by its
-peer by not replying with a PATH_RESPONSE frame but sending a
+A server can also decide to not perform path validation if it wants to deny the new path,
+or not perform it immediately if, e.g., no connection ID is yet available. 
+If an endpoint decides to deny the establishment of a new path,
+this should be considered as a path closure and therefore requires sending a
 PATH_ABANDON frame on another path for the associated Path ID used
-for the PATH_CHALLENGE instead.
+by the client as specified in section {{path-close}}.
 
 MP_ACK frames (defined in {{mp-ack-frame}}) can be returned on any path.
 If the MP_ACK is preferred to be sent on the same path as the acknowledged
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -375,6 +375,10 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+An endpoint MAY deny the establishment of a new path initiated by its
+peer by not replying with a PATH_RESPONSE frame but sending a
+PATH_ABANDON frame on another path for the associated Path ID used
+for the PATH_CHALLENGE instead.
 
 ACK_MP frames (defined in {{ack-mp-frame}}) can be returned on any path.
 If the ACK_MP is preferred to be sent on the same path as the acknowledged
@@ -595,15 +599,8 @@ before receiving or sending any traffic on a path. For example, if the client
 tries to initiate a path and the path cannot be established, it will send a
 PATH_ABANDON frame (see {{path-initiation}}). An endpoint may also decide
 to abandon a path for any reason, for example, removing a hole from
-the sequence of path IDs in use. This is not an error. The endpoint that
-receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
-
-## Refusing a New Path
-
-An endpoint may deny the establishment of a new path initiated by its
-peer during the address validation procedure. According to {{QUIC-TRANSPORT}},
-the standard way to deny the establishment of a path is to not send a
-PATH_RESPONSE in response to the peer's PATH_CHALLENGE.
+the sequence of path IDs in use. This is not an error. An endpoint that
+receives such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
 
 ## Allocating, Consuming, and Retiring Connection IDs {#consume-retire-cid}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -382,7 +382,7 @@ client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
 A server can also decide to not perform path validation if it wants to deny the new path,
-or not perform it immediately if, e.g., no connection ID is yet available. 
+or not perform it immediately if, e.g., no connection ID is yet available.
 If an endpoint decides to deny the establishment of a new path,
 this should be considered as a path closure and therefore requires sending a
 PATH_ABANDON frame on another path for the associated Path ID used
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -381,12 +381,12 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
-A server can also decide to not perform path validation if it wants to deny the new path,
+A server can also decide not to perform path validation if it wants to deny the new path,
 or not perform it immediately if, e.g., no connection ID is yet available.
 If an endpoint decides to deny the establishment of a new path,
 this should be considered as a path closure and therefore requires sending a
 PATH_ABANDON frame on another path for the associated Path ID used
-by the client as specified in section {{path-close}}.
+by the client as specified in {{path-close}}.
 
 MP_ACK frames (defined in {{mp-ack-frame}}) can be returned on any path.
 If the MP_ACK is preferred to be sent on the same path as the acknowledged
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -384,7 +384,7 @@ client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
 An endpoint that receives a PATH_CHALLENGE and does not want to establish
-this path should immediately close the path by sending a PATH_ABANDON
+this path is expected to close the path by sending a PATH_ABANDON
 on another path, as specified in section {{path-close}}.
 
 MP_ACK frames (defined in {{mp-ack-frame}}) can be returned on any path.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -377,16 +377,15 @@ When the multipath extension is negotiated, a client that wants to use an
 additional path MUST first initiate the Address Validation procedure
 with PATH_CHALLENGE and PATH_RESPONSE frames as described in
 {{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
-that address. After receiving packets from the
+that address.
+
+After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
-A server can also decide not to perform path validation if it wants to deny the new path,
-or not perform it immediately if, e.g., no connection ID is yet available.
-If an endpoint decides to deny the establishment of a new path,
-this should be considered as a path closure and therefore requires sending a
-PATH_ABANDON frame on another path for the associated Path ID used
-by the client as specified in {{path-close}}.
+An endpoint that receives a PATH_CHALLENGE and does not want to establish
+this path should immediately close the path by sending a PATH_ABANDON
+on another path, as specified in section {{path-close}}.
 
 MP_ACK frames (defined in {{mp-ack-frame}}) can be returned on any path.
 If the MP_ACK is preferred to be sent on the same path as the acknowledged


draft-ietf-quic-multipath.md
```diff
@@ -375,6 +375,10 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+An endpoint MAY deny the establishment of a new path initiated by its
+peer by not replying with a PATH_RESPONSE frame but sending a
+PATH_ABANDON frame on another path for the associated Path ID used
+for the PATH_CHALLENGE instead.
```
mirjak, line 381: Do we actually need to change this to a MUST: "If an endpoints decides to deny the establishment of a path, it MUST send an PATH_ABANDON frame..."?

Yanmei-Liu: I tend to support use MUST here, as we have insisted that PATH_ABANDON frame is a necessary explicit signal for path closure. If endpoint doesn't send PATH_ABANDON frame here, the peer may not know what happened with the path, and also not sure whether the path id resource should be consumed or not.

huitema: I would not use normative language. In fact, we have  potential race conditions: arrival of a new connection ID, availability of other resource, etc. If the endpoint ignores the path challenge, the challenge will be repeated a bit later, and the race conditions will be resolved. I think it is generally better to ignore the path challenge, and let the peer abandon the path when it abandons its own challenge -- unless there is a strong reason, such as "please don't use that interface."

huitema (approved):
Please leave it as is, without adding any "MUST abandon" rule!

draft-ietf-quic-multipath.md
mirjak: Is that better?
```suggestion
If an endpoint wants to deny the establishment of a new path initiated by its
peer by not replying with a PATH_RESPONSE frame, this is considered a path
closure and requires sending a
PATH_ABANDON frame on another path for the associated Path ID used
for the PATH_CHALLENGE as specified in section {{path-close}}.
```

mirjak: @huitema what's the problem of sending an path_abandon immediately rather then waiting for the peer to give up?

huitema: Race conditions. There are cases when the endpoint will not be ready to accept the PATH CHALLENGE, but will be ready a moment later, when a new PATH CHALLENGE is repeated on the same path.

In normal behavior, the endpoint just ignores the path challenge. The peer may try to repeat it, or may eventually abandon the path, triggering the reciprocal abandon by the challenged endpoint.

mirjak: But that's fine because the texts say "If an endpoint wants to deny the establishment". In you described case it does not want to deny it but it might delay it a bit. So ignoring the path challenges is always okay because there is also nothing you can do to enforce any other behavior. However, if the endpoint already notes that it doesn't want to have the connection then it should perform a path close.

huitema: Maybe, but I am afraid that this is a bit too subtle, and that implementers will just go for "queue an Abandon" instead of doing nothing. Maybe give a little warning, such as, cannot agree because of missing resource.

mirjak:
```suggestion
A server can also decide to not perform path validation if it wants to deny the new path,
or not perform it immediately if, e.g., no connection ID is yet available. 
If an endpoint decides to deny the establishment of a new path,
this should be considered as a path closure and therefore requires sending a
PATH_ABANDON frame on another path for the associated Path ID used
by the client as specified in section {{path-close}}.
```

mirjak:
@huitema I updated the test. Please review.

However, I propose to not merge this PR now but double-check with the group and then hopefully merge after the meeting!

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+A server can also decide to not perform path validation if it wants to deny the new path,
+or not perform it immediately if, e.g., no connection ID is yet available.
```
mirjak, line 385:
```suggestion
or not perform it immediately if, e.g., no connection ID is yet available.
```

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
...
+If an endpoint decides to deny the establishment of a new path,
+this should be considered as a path closure and therefore requires sending a
+PATH_ABANDON frame on another path for the associated Path ID used
+by the client as specified in section {{path-close}}.
```
huitema, line 390: I think that we are close to agreement. My suggestion would be to write:

An endpoint that cannot accept a path immediately has two choices. It can
decide to simply wait, waiting for example for an appropriate connection
ID; or, it can decide to deny the establishment of the new path. 
If the endpoint decides to deny the establishment of the new path,
this should be considered as a path closure and therefore requires sending a
PATH_ABANDON frame on another path for the associated Path ID used
by the client as specified in section {{path-close}}.

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
...
```
MikeBishop, line 389:
```suggestion
by the client as specified in {{path-close}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+A server can also decide to not perform path validation if it wants to deny the new path,
```
MikeBishop, line 384:
```suggestion
A server can also decide not to perform path validation if it wants to deny the new path,
```

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
...
+or not perform it immediately if, e.g., no connection ID is yet available.
```
MikeBishop, line 385: Should this also call out the impact of delaying path validation? For example, are anti-amplification limits in place until the path is validated?

mirjak: This likely needs some editorial work anyway based on the discussion in issue #390. I proposed to merge this PR first as it is and then work on another separate PR for issue #390.

However, @MikeBishop what do you mean by "impact on delaying path validation"? Yes, you can't send before the path is validated, or at least not more the amplification limit. But I guess that's the intention. I guess the only thing we can mention is that if you delay longer than the time-out, the other end will assume validation failed and close the path. But that should be fine as well because it will send an explicit abandon.

draft-ietf-quic-multipath.md
mirjak: There are actually two different cases here (and I agree it would be better to note them explicitly rather than mixing everything together into one sentence):

1. An endpoint receives a path validation on a new path and for whatever reason it simply doesn't want to open a new path or at least on that path. In this case, there is no point in sending the path_response but you rather send a path_abandon on another active path immediately.
2. An endpoint receives a path validation but it can't  reply to it because it doesn't have a CID for that path ID. In this case there are two option: either immediately send the path_abandon (and the other end could try again with a different path ID), or wait for a while to see if a CID for that path ID arrives. Further down in the text we actually don't recommend this second option except if the new CID frame was sent n the same packet but processed in the wrong order.

I think your proposed text doesn't cover case 1 accordingly and actually we might only really need to cover case 1 as case 2 is discussed further below. But again, I agree we need to adapt the wording.

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+A server can also decide not to perform path validation if it wants to deny the new path,
+or not perform it immediately if, e.g., no connection ID is yet available.
+If an endpoint decides to deny the establishment of a new path,
+this should be considered as a path closure and therefore requires sending a
+PATH_ABANDON frame on another path for the associated Path ID used
+by the client as specified in {{path-close}}.
```
mirjak, line 389:
```suggestion
If an endpoint decides to deny the establishment of a new path,
it may not response to the path challenge. However,
this should be considered as a path closure and therefore requires sending a
PATH_ABANDON frame on another path for the associated Path ID used
by the client as specified in {{path-close}}.
```

draft-ietf-quic-multipath.md
```diff
@@ -381,6 +381,12 @@ that address. After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+A server can also decide not to perform path validation if it wants to deny the new path,
+or not perform it immediately if, e.g., no connection ID is yet available.
```
mirjak, line 385:
```suggestion
```

draft-ietf-quic-multipath.md
mirjak: Okay, what I'm proposing here now is to remove the first sentence (see edit further above) and extend the second sentence to also say that you don't have to reply to the path challenge if you want to close. Does that work @huitema ?

What to do if you don't have a CID for this path is discussed further below. However, the text says this 

> If the server receives a PATH_CHALLENGE before receiving
> a MP_NEW_CONNECTION_ID for the specific path, it SHOULD
> ignore the PATH_CHALLENGE. Note that the
> MP_NEW_CONNECTION_ID might be sent in the same
> packet and in this case the PATH_CHALLENGE SHOULD
> be processed.

This is based on the discussion we have in issue #340 and added by PR #341. However, this seems wrong now as it should say that you can either close immediately by sending an abandon frame or wait for a bit. However, it should not be just ignored! Should we try to fix this part in this PR as well? I think I would prefer to leave it as it, merge this PR and create another issue for a new PR. Okay?

huitema: You want to add something about not sending a PATH_RESPONSE frame. That's fine, but maybe we need first to agree on what we want to say. There are two parts:

* sending a PATH_RESPONSE or not. 
* sending PATH_ABANDON.

We had a discussion in the WG, and there is some plausible variation on either statement.

First, I think that the way to not accept a path is indeed to not send a path response. We agree on that, but RFC 9000 section 8.2.2 says: _On receiving a PATH_CHALLENGE frame, an endpoint MUST respond by echoing the data contained in the PATH_CHALLENGE frame in a PATH_RESPONSE frame. An endpoint MUST NOT delay transmission of a packet containing a PATH_RESPONSE frame unless constrained by congestion control._ That means we make a reference to the 8.2.2, and specify that if the receiver does not want to accept the path, it must not send the PATH_RESPONSE. (Of course, an endpoint could also send a PATH_RESPONSE and then send a PATH_ABANDON later -- there is not rule against that.)

Second, we had a discussion of whether the PATH_ABANDON shall be sent immediately. I think it depends on the reason why the endpoint cannot accept the new path. It may be a definitive reason, but it may also be a temporary reason, such as "I have not received a PATH NEW CID for that PATH yet." In the temporary case, it makes sense to not send the PATH_ABANDON immediately, just wait until the client repeats the PATH_CHALLENGE and acknowledge it then.

If for whatever reason the denying endpoint does not send the PATH_ABANDON immediately, the peer will trying repeating the challenge, and eventually will give up and PATH_ABANDON. At that point, the endpoint MUST send its own PATH_ABANDON.

Our own challenge is to convey all that in a couple of short paragraphs.

draft-ietf-quic-multipath.md
```diff
@@ -602,15 +608,8 @@ before receiving or sending any traffic on a path. For example, if the client
...
-receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
-
-## Refusing a New Path
-
-An endpoint may deny the establishment of a new path initiated by its
-peer during the address validation procedure. According to {{QUIC-TRANSPORT}},
-the standard way to deny the establishment of a path is to not send a
-PATH_RESPONSE in response to the peer's PATH_CHALLENGE.
+the sequence of path IDs in use. This is not an error. An endpoint that
+receives such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
```
huitema, line 613: I agree on this part: removing the "refusing" scenario from this section.

draft-ietf-quic-multipath.md
huitema, line 389: An endpoint that receives a PATH_CHALLENGE and does not want to establish this path should immediately close the path by sending a PATH_ABANDON on another path, as specified in section {{path-close}}.

An endpoint that lacks resource to immediately configure a new path MAY delay sending the PATH_RESPONSE until sufficient resource are available. Long delays may cause the peer to repeat the PATH_CHALLENGE and eventually send a PATH_ABANDON, in which case the procedures specified in  section {{path-close}} apply.


If an endpoint decides to deny the establishment of a new path,
it may not response to the path challenge. However,
this should be considered as a path closure and therefore requires sending a
PATH_ABANDON frame on another path for the associated Path ID used
by the client as specified in {{path-close}}.

huitema (commented):
See proposal for alternate text.

draft-ietf-quic-multipath.md
mirjak: I have committed your first sentence. I will create a separate PR for you section sentence. The first part has been discussed and can be directly merged. So the second we might want to open an issue and way a bit for further feedback just to ensure transparency.

huitema: Two nits: 

1) the "should" in the first sentence SHOULD be capitalized.
2) the second semtence is the reason it I wrote "SHOULD" and not "MUST", so the two sentences complement each other.

mirjak: I don't think the second sentence is the reason for a SHOULD. In the first sentence the server doesn't want to open the path. In the second sentence the server wants to open the path but can't. So I think if the server does NOT want to open the path, it MUST send the path_abandon.

However, I don't think we should use normative language in this sentence at all because the normative part is the closing section which says:
> An endpoint that wants to close a path MUST explicitly
> terminate the path by sending a PATH_ABANDON frame.

This applies to all situation where an endpoint wants to close a path. We don't need to specify the same thing normatively twice.

draft-ietf-quic-multipath.md
```diff
@@ -377,10 +377,15 @@ When the multipath extension is negotiated, a client that wants to use an
...
 {{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
-that address. After receiving packets from the
+that address.
+
+After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+An endpoint that receives a PATH_CHALLENGE and does not want to establish
+this path should immediately close the path by sending a PATH_ABANDON
```
qdeconinck, line 387: Shouldn't be this "should" normative?

qdeconinck (approved):
Small non-blocking comment

draft-ietf-quic-multipath.md
mirjak: @huitema and I are still discussing this. Please see discussion in the conversations. I will relabel as not not-ready-to-merge.

huitema: @mirjak the conversations are marked as "resolved", so it is hard for @qdeconinck and others to get the arguments. (This is a side effect of carrying discussions in PR instead of in a specialized issue.)

@qdeconinck I think the text is fine, but we have two items of disagreement.

1. Should we use normative language (e.g., should -> SHOULD) in this specific section? 

2. If we do use normative language, should we say SHOULD or MUST?

The argument against normative language that we are going to make a reference to {{close}}, which itself contains normative language, so we should not duplicate it. I do not think this is a strong argument, because the paragraph just before this says _"When the multipath extension is negotiated, a client that wants to use an
additional path MUST first initiate the Address Validation procedure
with PATH_CHALLENGE and PATH_RESPONSE frames as described in
{{Section 8.2 of QUIC-TRANSPORT}}..."_ We also know from experience that using lower case "should" tends to confuse the readers, who will wonder whether this is a hard requirement or not.

The argument for MUST is the classic issue: if you say "SHOULD" instead of "MUST", in which cases shall the implementation not follow your advice? If you cannot spell out a case, just say MUST. In this case, there are plausible arguments for waiting a bit in some cases, e.g., waiting until an appropriate CID becomes available. That CID may well be just following the PATH_CHALLENGE in the same packet, or it may be arriving on another path. My original proposal had text to that effect, but we split it out.

So, my preference would be for SHOULD (upper case), and some additional text explaining the alternative. I think the additional text should be as "matter of fact" as possible. Something like: "The endpoint MAY delay sending a PATH_RESPONSE or a PATH_ABANDON, knowing that if neither or those is sent in a reasonable time the peer will eventually close the path as specified in section {{close}}."

mirjak: Just on your first point: I don't understand your argument. This section has normative language; that's fine because it's the normative language on what you need to do when you open a path. However, the normative language on closing a path should be in the section on path closure. Also what I wanted to say is that the normative requirement is already there in that section on path closure and broad enough that is already covers this case ("An endpoint that wants to close a path MUST explicitly terminate the path by sending a PATH_ABANDON frame."). It's not only about saying the same things twice normatively in the document in different places (which I think is also bad) but also to make clear that this case here is actually not special. It's always the same: if you want to close a path you have to send a PATH_ABANDON no matter what.

However, I think the point you are trying to make about the sentence here is actually slightly different because it says "should immediately close". So the question with the word "immediately" in there is: do we want to give any (normative) recommendation when exactly to close the path? First, we can't really enforce that an endpoint sends anything immediately or not anyway. However, I also don't understand your point above why there should be any benefit in waiting...? If you don't send the PATH_RESPONSE immediately because you are hoping to get more ressources, then you actually don't want to close the path (and respectively should not send an PATH_ABANDON). However, if you have decided that you don't want to establish the path (and that's what the sentence says), I don't see any reason to delay the sending of a PATH_ABANDON.

In summary I don't see a reason to delay the PATH_ABANDON and I don't think we should talk about the timing at all. So I still don't think we should use normative language here. However, I agree that lower case "should"s are confusing and we could consider using a different phrasing that avoids that!

huitema: Maybe the simplest solution is to strike "immediately" and just say SHOULD. And yes, you are correct about the gray area between "decide it does not want a path" and "decide it wants it but cannot accept just yet" and "accept the path". Maybe that's something best explained in the implementation section, not here.

mirjak, line 387:
```suggestion
An endpoint that receives a PATH_CHALLENGE and does not want to establish
this path is expected to close the path by sending a PATH_ABANDON
```

qdeconinck (approved):

PR #393, opened 2024-07-01
More guidance on "standby"

### LINKED ISSUE #293: Switch to standby path more tricky than expected
I did a big batch of debugging the "unique path-id" code in picoquic, porting all the tests that were designed for the previous multipath version, and I found an interesting issue in the "standup" test. The test starts by setting a client to server connection with two paths, one available and on in standby. It runs for a while, then simulates cutting the "available" path off. Expectation is that the connection will continue with the "standby" path. The test was initially failing.

The simulated traffic is from server to client. The server quickly detects that the available path is down, and starts sending data packets on the "standby" path. But the client only sends ACK, and does not react quickly if ACK packets are not acknowledged. So the client keeps sending ACKs on the "available path". Of course, since the path is cut, they are dropped. The server does not see ACKs for the packets sent on the "standby" path, so it quickly concludes that this path is down. Pretty soon, the connection breaks.

There are two potential fixes. One would be to somehow force ACKs on the standby path, if packets are received on that path. The other, which I feel is more robust, is for the server to send mark the standby path as "available" if the available path is "broken". I did that, "promote" the standby path to available, and the "standup" test now succeeds.

This issue might deserve some discussion in the multipath draft.

mirjak:
fixes #293

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -453,7 +453,7 @@ An enpoints that detects a path breakage can also explicitly close the path
 by sending a PATH_ABANDON frame (see section {{path-close}}) in order to avoid
 that its peer keeps using it and enable faster switch over to a standby path.
 If the endpoints do not want to close the path immediately, as connectivity
-could be re-established, it is RECOMMENDED to send PING frames to quickly detect
+could be re-established, PING frames can potentially be used to quickly detect
 connectivity changes and switch back in a timely way.
 
 If no frame indicating a path usage preference was received for a certain path,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
 Note that an endpoint might not follow the peer’s advertisements,
 but these frames are still a clear signal of the peer's preference of path usage.
 Each peer indicates its preference of path usage independently of the other peer.
-It means that peers may have different usage preferences for the same path.
-Depending on the sender's decisions, this may lead to usage of paths that have been
+That means that peers may have different usage preferences for the same path.
+Depending on the data sender's decisions, this may lead to usage of paths that have been
 indicated as "standby" by the peer or non-usage of some locally available paths.
 
 PATH_AVAILABLE indicates that a path is "available", i.e., it suggests to
 the peer to use its own logic to split traffic among available paths.
+
 PATH_STANDBY marks a path as "standby", i.e., it suggests that no traffic
 should be sent on that path if another path is available.
+If all active paths are marked as "standby", no guidance is provided about
+which path should be used.
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to updates it own path state signaling such that the peer avoids using the broken path.
+An enpoints that detects a path breakage can also explicitly close the path
+by sending a PATH_ABANDON frame (see section {{path-close}}) in order to avoid
+that its peer keeps using it and enable faster switch over to a standby path.
+If the endpoints does not want to close the path immediately, as connectivity
+could be re-established, it is RECOMMENDED to send PING frames to quickly detect
+connecivtity changes and switch back in a timely way.
+
 If no frame indicating a path usage preference was received for a certain path,
 the preference of the peer is unknown and the sender needs to decide based on it
 own local logic if the path should be used.
@@ -453,9 +466,6 @@ changed. Note that both frames can be sent via a different path
 and therefore might arrive in different orders.
 The PATH_AVAILABLE and PATH_STANDBY frames share a common sequence number space
 to detect and ignore outdated information.
-
-If all active paths are marked as "standby", no guidance is provided about
-which path should be used.
 
 ## Path Close {#path-close}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -446,9 +446,10 @@ PATH_STANDBY marks a path as "standby", i.e., it suggests that no traffic
 should be sent on that path if another path is available.
 If all active paths are marked as "standby", no guidance is provided about
 which path should be used.
+
 If an endpoints starts using a path that was marked as "standby" by its peer
 because it has detected issues on the paths marked as "available", it is RECOMMENDED
-to updates it own path state signaling such that the peer avoids using the broken path.
+to update it own path state signaling such that the peer avoids using the broken path.
 An enpoints that detects a path breakage can also explicitly close the path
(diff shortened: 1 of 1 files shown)


huitema (commented):
Looks good, but I would get rid of the last sentence in the proposed paragraph. As I said in the comment, this is either too little or too much...

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
+which path should be used.
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to updates it own path state signaling such that the peer avoids using the broken path.
+An enpoints that detects a path breakage can also explicitly close the path
+by sending a PATH_ABANDON frame (see section {{path-close}}) in order to avoid
+that its peer keeps using it and enable faster switch over to a standby path.
+If the endpoints does not want to close the path immediately, as connectivity
+could be re-established, it is RECOMMENDED to send PING frames to quickly detect
+connecivtity changes and switch back in a timely way.
```
Yanmei-Liu, line 457:
```suggestion
connectivity changes and switch back in a timely way.
```

Yanmei-Liu (commented):
Basically I support the guidance concept here. 
About the PING frames used for detect connectivity, frequency guidance might be useful for implementations, but I tend to keep the flexibility as implementations could choose different interval for different scenes.  PING frames sent frequently might be harmless for PC and laptop, but it could cause non-negligible power consumption for mobile devices and we could choose to only use it in necessary scenes.

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
 should be sent on that path if another path is available.
+If all active paths are marked as "standby", no guidance is provided about
+which path should be used.
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to updates it own path state signaling such that the peer avoids using the broken path.
+An enpoints that detects a path breakage can also explicitly close the path
+by sending a PATH_ABANDON frame (see section {{path-close}}) in order to avoid
+that its peer keeps using it and enable faster switch over to a standby path.
+If the endpoints does not want to close the path immediately, as connectivity
```
mirjak, line 455:
```suggestion
If the endpoints do not want to close the path immediately, as connectivity
```

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
+could be re-established, it is RECOMMENDED to send PING frames to quickly detect
+connecivtity changes and switch back in a timely way.
+
```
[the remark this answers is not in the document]
mirjak: I agree that normative language is probably not useful here. Above, I propose a change to "PING frames can potentially be used". However, I think I would like to keep this sentence in some form to at least say that a broken path doesn't need to be closed immediately because it might make sense to see if works again. However, then the questions is how to know that it works again...?

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
+which path should be used.
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to updates it own path state signaling such that the peer avoids using the broken path.
+An enpoints that detects a path breakage can also explicitly close the path
+by sending a PATH_ABANDON frame (see section {{path-close}}) in order to avoid
+that its peer keeps using it and enable faster switch over to a standby path.
+If the endpoints do not want to close the path immediately, as connectivity
+could be re-established, it is RECOMMENDED to send PING frames to quickly detect
+connectivity changes and switch back in a timely way.
```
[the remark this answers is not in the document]
huitema: Works for me.

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to updates it own path state signaling such that the peer avoids using the broken path.
+An enpoints that detects a path breakage can also explicitly close the path
+by sending a PATH_ABANDON frame (see section {{path-close}}) in order to avoid
+that its peer keeps using it and enable faster switch over to a standby path.
+If the endpoints do not want to close the path immediately, as connectivity
+could be re-established, PING frames can potentially be used to quickly detect
+connectivity changes and switch back in a timely way.
+
```
huitema, line 458: Nits: you may want to add a blank line after line 448, to delineate a paragraph. Also, no s for update in "to update"

huitema (approved):

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
+Depending on the data sender's decisions, this may lead to usage of paths that have been
 indicated as "standby" by the peer or non-usage of some locally available paths.
 
 PATH_AVAILABLE indicates that a path is "available", i.e., it suggests to
 the peer to use its own logic to split traffic among available paths.
+
 PATH_STANDBY marks a path as "standby", i.e., it suggests that no traffic
 should be sent on that path if another path is available.
+If all active paths are marked as "standby", no guidance is provided about
+which path should be used.
```
mirjak, line 448:
```suggestion
which path should be used.

```

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,27 @@ send packets with the preference expressed by these frames.
...
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to updates it own path state signaling such that the peer avoids using the broken path.
```
mirjak, line 451:
```suggestion
to update it own path state signaling such that the peer avoids using the broken path.
```

draft-ietf-quic-multipath.md
```diff
@@ -435,14 +435,28 @@ send packets with the preference expressed by these frames.
...
 the peer to use its own logic to split traffic among available paths.
+
 PATH_STANDBY marks a path as "standby", i.e., it suggests that no traffic
 should be sent on that path if another path is available.
+If all active paths are marked as "standby", no guidance is provided about
+which path should be used.
+
+If an endpoints starts using a path that was marked as "standby" by its peer
+because it has detected issues on the paths marked as "available", it is RECOMMENDED
+to update it own path state signaling such that the peer avoids using the broken path.
```
mirjak, line 452:
```suggestion
to update its own path state signaling such that the peer avoids using the broken path.
```

PR #394, opened 2024-07-01
limit initial_max_path to 2^32-1 and more explanation

### LINKED ISSUE #307: Clarify requirement to avoid wrapping CID seqenunce number
Sectionm 5.2:
"Section 19 of [QUIC-TRANSPORT] encodes the Connection ID Sequence Number as a variable-length integer, allowing values up to 2^62-1; in this specification, a range of less than 2^32-1 values MUST be used before updating the packet protection key"

I think this is unlikely to cause a real issue, but is sloppy writing. At no point in time may a sequence number n and its next wrapped value ( n+2^32) be protected by the same packet protection key. This is unlikely to ever happen due to AEAD limits of at least current ciphers. But I think it would be good to be formally correct here.

I will note that there is duplication of the imprecise requirement in the last paragraph of the seciton too: "Due to the way the nonce is constructed, endpoints MUST NOT use more than 2^32 Connection IDs without a key update."

mirjak:
fixes #307

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1269,6 +1269,7 @@ Maximum Path Identifier:
   as the 32 bits of the Path ID are used to calculate the nonce (see {{multipath-aead}}).
   The Maximum Path Identifier value MUST NOT be lower than the value
   advertised in the initial_max_path_id transport parameter.
+  
 Receipt of an invalid Maximum Path Identifier value MUST be treated as a
 connection error of type MP_PROTOCOL_VIOLATION.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -249,20 +249,22 @@ the use of the multipath extension during the connection handshake,
 as specified in {{QUIC-TRANSPORT}}. The new transport parameter is
 defined as follows:
 
-- initial_max_path_id (current version uses 0x0f739bbc1b666d09): the
-  initial_max_path_id transport parameter is included if the endpoint supports
-  the multipath extension as defined in this document. This is
-  a variable-length integer specifying the maximum path identifier
+- initial_max_path_id (current version uses 0x0f739bbc1b666d09): a
+  variable-length integer specifying the maximum path identifier
   an endpoint is willing to maintain at connection initiation.
-  For example, if initial_max_path_id is set to 1, only connection IDs
-  associated with Path IDs 0 and 1 should be issued by the peer.
-  If an endpoint receives an initial_max_path_id transport parameter with value 0,
-  the peer aims to  enable the multipath extension without allowing extra paths immediately.
-  To allow for the use of more paths later,
-  endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
+  This value MUST NOT exceed 2^32-1, as Path IDs are defined
+  with a maximum value 2^32-1 as the 32 bits of the Path ID
+  are used to calculate the nonce (see {{multipath-aead}}).
+
+For example, if initial_max_path_id is set to 1, only connection IDs
+associated with Path IDs 0 and 1 should be issued by the peer.
+If an endpoint receives an initial_max_path_id transport parameter with value 0,
+the peer aims to  enable the multipath extension without allowing extra paths immediately.
 
 Setting initial_max_path_id parameter is equivalent to sending a
 MAX_PATH_ID frame ({{max-paths-frame}}) with the same value.
+As such to allow for the use of more paths later,
+endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
 
 If either of the endpoints does not advertise the initial_max_path_id transport
 parameter, then the endpoints MUST NOT use any frame or
@@ -690,10 +692,10 @@ for 1-RTT packets requires changes in AEAD usage.
 the use of a nonce, N, formed by combining the packet protection IV
 with the packet number. When multiple packet number spaces are used,
 the packet number alone would not guarantee the uniqueness of the nonce.
-
-In order to guarantee the uniqueness of the nonce, the nonce N is
-calculated by combining the packet protection IV with the packet number
-and with the least significant 32 bits of the Path ID.
+Therefore, the nonce N is calculated by combining the packet protection
+IV with the packet number and with the least significant 32 bits of the
+Path ID. In order to guarantee the uniqueness of the nonce, the Path ID
+is limited to a max value of 2^30-1.
 
 To calculate the nonce, a 96-bit path-and-packet-number is composed of the least
 significant 32 bits of the Path ID in network byte order,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1266,9 +1268,10 @@ Maximum Path Identifier:
   This value MUST NOT exceed 2^32-1, as Path IDs are defined with a maximum value 2^32-1
   as the 32 bits of the Path ID are used to calculate the nonce (see {{multipath-aead}}).
   The Maximum Path Identifier value MUST NOT be lower than the value
-  advertised in the initial_max_path_id transport parameter. Receipt
-  of an invalid Maximum Path Identifier value MUST be treated as a
-  connection error of type MP_PROTOCOL_VIOLATION.
+  advertised in the initial_max_path_id transport parameter.
+
```
mirjak, line 1272:
```suggestion

```

draft-ietf-quic-multipath.md
```diff
@@ -1266,9 +1268,9 @@ Maximum Path Identifier:
   This value MUST NOT exceed 2^32-1, as Path IDs are defined with a maximum value 2^32-1
   as the 32 bits of the Path ID are used to calculate the nonce (see {{multipath-aead}}).
   The Maximum Path Identifier value MUST NOT be lower than the value
-  advertised in the initial_max_path_id transport parameter. Receipt
-  of an invalid Maximum Path Identifier value MUST be treated as a
-  connection error of type MP_PROTOCOL_VIOLATION.
+  advertised in the initial_max_path_id transport parameter.
```
mirjak, line 1271:
```suggestion
  advertised in the initial_max_path_id transport parameter.
  
```

draft-ietf-quic-multipath.md
```diff
@@ -249,20 +249,22 @@ the use of the multipath extension during the connection handshake,
...
   an endpoint is willing to maintain at connection initiation.
-  For example, if initial_max_path_id is set to 1, only connection IDs
-  associated with Path IDs 0 and 1 should be issued by the peer.
-  If an endpoint receives an initial_max_path_id transport parameter with value 0,
-  the peer aims to  enable the multipath extension without allowing extra paths immediately.
-  To allow for the use of more paths later,
-  endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
+  This value MUST NOT exceed 2^32-1, as Path IDs are defined
+  with a maximum value 2^32-1 as the 32 bits of the Path ID
+  are used to calculate the nonce (see {{multipath-aead}}).
```
mirjak, line 257: Note, this is a normative change to explicitly state the MUST here. This is the same text as used for the MAX_PATHS frame value but I think we have to specify it here as well.

draft-ietf-quic-multipath.md
mirjak, line 1272:
```suggestion

```

draft-ietf-quic-multipath.md
```diff
@@ -1266,9 +1268,9 @@ Maximum Path Identifier:
   This value MUST NOT exceed 2^32-1, as Path IDs are defined with a maximum value 2^32-1
   as the 32 bits of the Path ID are used to calculate the nonce (see {{multipath-aead}}).
   The Maximum Path Identifier value MUST NOT be lower than the value
-  advertised in the initial_max_path_id transport parameter. Receipt
-  of an invalid Maximum Path Identifier value MUST be treated as a
-  connection error of type MP_PROTOCOL_VIOLATION.
+  advertised in the initial_max_path_id transport parameter.
+Receipt of an invalid Maximum Path Identifier value MUST be treated as a
```
mirjak, line 1272:
```suggestion

Receipt of an invalid Maximum Path Identifier value MUST be treated as a
```

draft-ietf-quic-multipath.md
```diff
@@ -249,20 +249,22 @@ the use of the multipath extension during the connection handshake,
...
+- initial_max_path_id (current version uses 0x0f739bbc1b666d09): a
+  variable-length integer specifying the maximum path identifier
   an endpoint is willing to maintain at connection initiation.
-  For example, if initial_max_path_id is set to 1, only connection IDs
-  associated with Path IDs 0 and 1 should be issued by the peer.
-  If an endpoint receives an initial_max_path_id transport parameter with value 0,
-  the peer aims to  enable the multipath extension without allowing extra paths immediately.
-  To allow for the use of more paths later,
-  endpoints can send the MAX_PATH_ID frame to increase the maximum allowed path identifier.
+  This value MUST NOT exceed 2^32-1, as Path IDs are defined
```
Yanmei-Liu, line 255:
```suggestion
  This value MUST NOT exceed 2^30-1, as Path IDs are defined
```

draft-ietf-quic-multipath.md
```diff
@@ -249,20 +249,22 @@ the use of the multipath extension during the connection handshake,
...
+  with a maximum value 2^32-1 as the 32 bits of the Path ID
```
Yanmei-Liu, line 256:
```suggestion
  with a maximum value 2^30-1 as the 32 bits of the Path ID
```

Yanmei-Liu (commented):
Thanks for this PR. 
Basically I agree that declare the limitation in transport parameter definition would help, but I think it would be more clear to use the same limitation 2^30-1, as in AEAD usage section we described the reason is to guarantee the uniqueness of the nonce.

mirjak:
Sorry I missed up the title of the PR, however, the limit is 2^32-1 everywhere as the nonce uses 32 bits. In issue #307 a limit of 2^30-1 was proposed (to save some bits) but I don't think that's useful. However, feel free to further comment in issue #307 !

draft-ietf-quic-multipath.md
```diff
@@ -690,10 +692,10 @@ for 1-RTT packets requires changes in AEAD usage.
...
 with the packet number. When multiple packet number spaces are used,
 the packet number alone would not guarantee the uniqueness of the nonce.
-
-In order to guarantee the uniqueness of the nonce, the nonce N is
-calculated by combining the packet protection IV with the packet number
-and with the least significant 32 bits of the Path ID.
+Therefore, the nonce N is calculated by combining the packet protection
+IV with the packet number and with the least significant 32 bits of the
+Path ID. In order to guarantee the uniqueness of the nonce, the Path ID
+is limited to a max value of 2^30-1.
```
huitema, line 699: Should be "is limited to a max value of 2^32-1." We want 32 bits, not 30 bits. But this is redundant with what we say in the section "Introduction of an Explicit Path Identifier." We should not have two places stating the limit.

draft-ietf-quic-multipath.md
huitema: Do we need to restate the AEAD reason here? I would rather only have that in the ".Introduction of an Explicit Path Identifier." The text should be:

This value MUST NOT exceed 2^32-1, the maximum value of the Path ID (see {{introduction-of-an-explicit-path-identifier}}).
The Maximum Path Identifier value MUST NOT be lower than the value
advertised in the initial_max_path_id transport parameter.

huitema (commented):
We are defining the 2^32-1 limit at multiple places (and getting it wrong at one of these places). I think it should be fully defined in "Introduction of an Explicit Path Identifier", and then referenced in the other places.

draft-ietf-quic-multipath.md
mirjak: ups, yes I definitely messed this up here. I'll fix it to the right value now and then make another proposal to repeat this less often.

draft-ietf-quic-multipath.md
```diff
@@ -690,10 +692,10 @@ for 1-RTT packets requires changes in AEAD usage.
...
```
mirjak, line 698:
```suggestion
is limited to a max value of 2^32-1.
```

draft-ietf-quic-multipath.md
mirjak: How about this (referencing the nonce section directly rather than the intro but not repeating the value multiple times)?
```suggestion
  This value MUST NOT exceed 2^32-1, the maximum allowed value for the Path ID due to
  restrictions on the nonce calculation (see {{multipath-aead}}).
```

huitema: Yes, that's good too.

Yanmei-Liu (approved):
Looks good.

PR #395, opened 2024-07-01
stream boundaries of retransmissions

### LINKED ISSUE #309: Clarify refragmenting of stream data when MTU is different and retransmission happens on other path.
Section 7.6
Is refragmentation of data in DATA frames allowed to cope with PMTUD. Maybe should be mentioned explicitly to ensure that this is not resulting in implementation issues. Sender responsibility?

mirjak:
fixes #309

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -932,6 +932,12 @@ recommended for general purpose use due to the network
 overhead). While this document does not preclude a specific
 strategy, more detailed specification is out of scope.
 
+As noted in {{Section 2.2 of QUIC-TRANSPORT}}, STREAM frame boundaries are not
+expected to be preserved when data is retransmitted. Especially when STREAM
+frames have to be retransmitted over a different path with a smaller MTU limit,
+new smaller STREAM frames might need to be sent instead.
+
+
 ## Handling different PMTU sizes
 
 An implementation should take care to handle different PMTU sizes across


huitema (approved):

draft-ietf-quic-multipath.md
```diff
@@ -932,6 +932,12 @@ recommended for general purpose use due to the network
 overhead). While this document does not preclude a specific
 strategy, more detailed specification is out of scope.
 
+As noted in {{Section 2.2 of QUIC-TRANSPORT}}, STREAM frame boundaries are not
+expected to be preserved when data is retransmitted. Especially when STREAM
+frames have to be retransmitted over a different path with a smaller MTU limit,
+new smaller STREAM frames might need to be sent instead.
```
LPardue, line 938: Conversely, they might need to be tra omitted on a larger path, with the expectation the boundary grows to suit the MTU. 

In other words, I think the text is too specific to one case and might benefit from more generalisation, or else people might try to optimise to only the one case.

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
mirjak: I think the case noted is the important part here, because if you suddenly only have paths with smaller MTUs you *have to* re-segment. If you switch to a path with a larger MTU, you could potentially re-segment but actually you can only do the if you transmit multiple consecutive frames which is probably not that often the case.

mirjak: @LPardue note that this sentence is specific a note about retransmissions. The next section actually gives more generally advice about paths with different MTUs. Do you think we should add more guidance there? If so, can you make a proposal or open and issue?

mirjak: @LPardue will merge now. If you think more guidance is needed. Please open a new issue!

Issue #378, opened 2024-06-05
Do we need further guidance if multiple paths over the same 4-tuple are used?


mirjak:
Currently the draft says that only one path per 4-tuple is allowed. This was needed previously to better distinguish between migration and a new path. With use of the explicit path ID this is not needed anymore. Should we therefore allow for multiple paths over the same 4-tuple (using different path IDs)?

Note that from a congestion control point of view that could give you a benefit as each path has an own congestion control. However, we could recommend to use coupled congestion control or even simply divide the cwnd by the number of paths. Or we do nothing here because you can really enforce a certain congestion control behavior and effectively anybody could just multiple the cwnd by some factor even with only one path.

Also note that if we want to only allow for one path per 4-tuple, we would probably need to describe a way to enforce that by either ignoring an path validation attempt or actively closing one of the two paths.

huitema:
I think it suffices to point out that endpoints MAY abandon one path if another path uses the same 4 tuple -- because of course they MAY abandon any path at any time. The main problem I see is the possible collision. Path 1 and 2 share the same 4 tuple, client abandons 1 and at the same time server abandons 2, leave the connection with no path. Maybe we need to say something like "in these conditions, always abandon the one with the lowest path-ID".

tzaeschke:
> [...] The main problem I see is the possible collision. Path 1 and 2 share the same 4 tuple, client abandons 1 and at the same time server abandons 2, leave the connection with no path. Maybe we need to say something like "in these conditions, always abandon the one with the lowest path-ID".

Just to clarify, if I understand correctly, the problem of both sides closing paths is independent of whether we allow multiple paths over the same 4-tuple, or not? It would be the same if path 1 and 2 used different 4-tuples?

mirjak:
We merged #400 that removed the restriction to only have one path per 4-tuple. However, I'm leaving this issue open for now to potentially discuss if further guidance is needed.

gloinul:
I think one aspect of multiple paths over the same 4-tuple is that they may actually not be equal. For example one could have two paths, where the packets for one path is using a different DSCP than the other, or enable L4S. These two packets can thus see significant different forwarding behaviours. So it is not necessary true that they are equivalent.

mirjak:
At IETF-120 @huitema agreed to take a first try on more guidance and use case consideration as also indicated by @gloinul above.

PR #389, opened 2024-07-01
remove probing concept and text on migration

mirjak:
fixes #347 and some left-over on migration that do not apply anymore

PR #400, opened 2024-07-02
Multiple paths may use the same 4 tuple.

huitema:
This is a simple PR designed to fix issue #378. It is mostly editorial, simply stating that multiple path sharing the same 4-tuple is just fine, and checking the draft to make sure we never say that a path is defined by a 4-tuple.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -102,7 +102,7 @@ several changes to that mechanism:
 paths.
   *  Continuous use of an existing path even if non-probing packets have
 been received on another path.
-  *  Introduction of an path idendifier to manage connection IDs and
+  *  Introduction of an path identifier to manage connection IDs and
      packet number spaces per path.
   *  Removal of paths that have been abandoned.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -633,7 +640,7 @@ for a given Path ID to its peer. Endpoints SHOULD NOT introduce discontinuity
 in the issuing of Path IDs through their connection ID advertisements as path initiation
 requires available connection IDs for the same Path ID on both sides. For instance,
 if the maximum Path ID limit is 2 and the endpoint wants to provide connection IDs
-for only one Path ID inside range [1, 2], it should select Path ID 1 (and not Path
+for only one Path ID inside range (1, 2), it should select Path ID 1 (and not Path
```
Yanmei-Liu, line 643: The original

Yanmei-Liu (approved):
Thanks a lot for the PR.

draft-ietf-quic-multipath.md
```diff
@@ -957,8 +964,8 @@ at any time during the connection. As such a sole change of the Connection
 ID without any change in the address does not indicate a path change and
 the endpoint can keep the same congestion control and RTT measurement state.
 
-While endpoints assign a connection ID to a specific sending 4-tuple,
-networks events such as NAT rebinding may make the packet's receiver
+While endpoints send a packet on a specific path with the associated
+4-tuple, networks events such as NAT rebinding may make the packet's receiver
```
mirjak, line 968: I proposed to remove this text in PR #389 as there is nothing special here anymore. It just describing RFC9000 behavior.

draft-ietf-quic-multipath.md
```diff
@@ -175,11 +175,17 @@ which usually also requires per-path RTT measurements
...
+  * Connection IDs are associated with a path ID. The path initiation
+associates that path ID with a 4-tuple of source and destination IP
+address as well as source and destination port.
+  * Migration is detected without ambiguity
+when a packet arrives with a connection ID
+pointing to an existing path ID, but the connection ID and/or the
+4-tuple are different from the value associated with that path
+(see {{migration}}).
+  * Paths can be closed at any time, as specified in {{path-close}}.
+  * It is not impossible to create multiple paths sharing the same 4-tuple.
```
mirjak, line 187:
```suggestion
  * It is possible to create multiple paths sharing the same 4-tuple.
```

mirjak (approved):

draft-ietf-quic-multipath.md
huitema: Yes, we may have a merge conflict with PR #389. But if we delete it now, we will have a different conflict. What do you suggest? Keep the old text for now, knowing that it will be removed?

draft-ietf-quic-multipath.md
huitema: The problem is kramdown. The angle brackets are confused with a library reference, and we cannot compile the text. It might be possible to escape kramdown by using something like "\[1, 2\]", but I have not tried that.

draft-ietf-quic-multipath.md
huitema: OK, reverting that particular change to avoid a merge conflict with #389.

draft-ietf-quic-multipath.md
huitema: Tried it now, it seems to work.

Issue #368, opened 2024-05-29
Details of MP_NEW_CONNECTION_ID frames are still missing


huitema:
The specification of MP_NEW_CONNECTION_ID and MP_RETIRE_CONNECTION_ID frames was clarified in PR #361, but some minor details are still missing:

1. In RFC 9000, there is a discussion about enforcing the connection ID limit after the combination of "retire prior" and the newly added CID. Maybe we should add a bit of text explaining that this limit is per path.

2. In RFC9000, there is a prohibition from applying "retire to" to the CID used to send the packet. I think that still stands, but we that we should clarify that the test applies to the path-ID + sequence number tuple.

3. In RFC 9000 says that "_Endpoints SHOULD NOT issue updates of the Retire Prior To field before receiving RETIRE_CONNECTION_ID frames that retire all connection IDs indicated by the previous Retire Prior To value._" We may want to clarify that this requirement still stands, but only apply to the CID with the same path-ID.

mirjak:
#401 merged

PR #401, opened 2024-07-02
Specify details of MP_RETIRE_CONNECTION_ID

huitema:
Addresses issue #368.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -664,7 +664,7 @@ will not be used anymore. In response, if the path is still active, the peer
 SHOULD provide new connection IDs using MP_NEW_CONNECTION_ID frames.
 
 Retirement of connection IDs will not retire the Path ID
-that correspones to the connection ID.
+that corresponds to the connection ID.
 The list of received packets used to send acknowledgements also remains
 unaffected as the packet number space is associated with a path.
 
@@ -1197,11 +1197,21 @@ Path Identifier:
 Note that, other than for the NEW_CONNECTION_ID frame of {{Section 19.15 of QUIC-TRANSPORT}},
 the sequence number applies on a per-path context.
 This means different connection IDs on different paths may have the same
-sequence number value. Respectively, the Retire Prior To field indicates which connection IDs
-should be retired for the path with the Path ID in the Path Identifier field.
+sequence number value. However, the value of the Connection ID MUST be unique
+across all paths.
+
+The Retire Prior To field indicates which connection IDs
+should be retired among those that share the Path ID in the Path Identifier field.
+Connection IDs associated with different path IDs are not affected.
 
 Note that the NEW_CONNECTION_ID frame can only be used to issue or retire
 connection IDs for the initial path with Path ID 0.
+
+Endpoints SHOULD NOT issue updates of the Retire Prior To field for a specific path ID
+before receiving RETIRE_CONNECTION_ID frames that retire all connection IDs indicated
+by the previous Retire Prior To value for that path ID. That is, the rule stated
+in the last paragraph of {{Section 5.1.2 of QUIC-TRANSPORT}} shall be applied
+independently for each path, not for the whole connection.
 
 ## MP_RETIRE_CONNECTION_ID frames {#mp-retire-conn-id-frame}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1197,12 +1197,22 @@ Path Identifier:
 Note that, other than for the NEW_CONNECTION_ID frame of {{Section 19.15 of QUIC-TRANSPORT}},
 the sequence number applies on a per-path context.
 This means different connection IDs on different paths may have the same
-sequence number value. Respectively, the Retire Prior To field indicates which connection IDs
-should be retired for the path with the Path ID in the Path Identifier field.
+sequence number value. However, the value of the Connection ID MUST be unique
+across all paths.
```
mirjak, line 1201: I don't think we need normative language here as this is already specified in RFC9000. Let's not specify this again.

Also note that 3.5 already says this:

```
As further specified in [Section 5.1](https://rfc-editor.org/rfc/rfc9000#section-5.1) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)] connection IDs cannot be issued more than once on the same connection and therefore are unique for the scope of the connection, regardless of the associated Path ID.
```

We can repeat it here (but not normatively).

draft-ietf-quic-multipath.md
```diff
@@ -1197,12 +1197,22 @@ Path Identifier:
...
+
+The Retire Prior To field indicates which connection IDs
+should be retired among those that share the Path ID in the Path Identifier field.
+Connection IDs associated with different path IDs are not affected.
 
 Note that the NEW_CONNECTION_ID frame can only be used to issue or retire
 connection IDs for the initial path with Path ID 0.
 
+Endpoints SHOULD NOT issue updates of the Retire Prior To field for a specific path ID
```
mirjak, line 1210: Let's not use normative language here is it is already specified normatively in RFC9000.

draft-ietf-quic-multipath.md
```diff
@@ -1238,8 +1248,31 @@ Note that the RETIRE_CONNECTION_ID frame can only be used to retire
...
+in which the frame is contained. The peer MAY treat this as a connection
+error of type PROTOCOL_VIOLATION. If the multipath extension is negotiated,
+the same error is detected if a RETIRE_CONNECTION_ID frame is received,
+the path ID of the Destination Connection ID field of the packet is 0,
+and the sequence number of that connection ID matches Sequence Number
+field of the RETIRE_CONNECTION_ID frame.
+
+These last two paragraphs change on
+the processing of RETIRE_CONNECTION_ID frames specified in
+{{Section 19.17 of QUIC-TRANSPORT}}.
```
mirjak, line 1275: I don't think we should repeat all normative language from RFC9000 here. I would rather prefer something like:

```
The processing as described In {{Section 19.17 of QUIC-TRANSPORT}} applies for MP_RETIRE_CONNECTION_ID frames per path, while the processing of an RETIRE_CONNECTION_ID frame is only applied for Path ID 0.
```

mirjak (changes requested):
I don't think we should repeat normative requirements from RFC9000 here.

draft-ietf-quic-multipath.md
huitema: What changes is the processing of the plain RETIRE_CONNECTION_ID, and that is a deviation from RFC 9000. Implementations need not just implement the new MP_RETIRE_CONNECTION_ID, they need to change the way they implemented RETIRE_CONNECTION_ID.

draft-ietf-quic-multipath.md
mirjak: We can have more than one sentence but I don't think we should normatively rephrase things from RFC9000. So if you add text that ensure it is well understood what we mean without using normative language that would be my preferred option.

draft-ietf-quic-multipath.md
mirjak: Yes, I thought that exactly what my proposed sentence point out explicitly. I believe saying this is actually more clear than repeating the text from RFC9000.

draft-ietf-quic-multipath.md
huitema: OK, updating the text.

draft-ietf-quic-multipath.md
huitema: OK, updating the text.

draft-ietf-quic-multipath.md
huitema: OK, removed the last sentence of that paragraph.

Yanmei-Liu (approved):
Looks good!

draft-ietf-quic-multipath.md
```diff
@@ -1197,12 +1197,22 @@ Path Identifier:
...
+sequence number value.
+
+The Retire Prior To field indicates which connection IDs
+should be retired among those that share the Path ID in the Path Identifier field.
+Connection IDs associated with different path IDs are not affected.
 
 Note that the NEW_CONNECTION_ID frame can only be used to issue or retire
 connection IDs for the initial path with Path ID 0.
 
+The last paragraph of {{Section 5.1.2 of QUIC-TRANSPORT}} specified how to
```
mirjak, line 1209:
```suggestion
The last paragraph of {{Section 5.1.2 of QUIC-TRANSPORT}} specifies how to
```

draft-ietf-quic-multipath.md
```diff
@@ -1238,8 +1248,14 @@ Note that the RETIRE_CONNECTION_ID frame can only be used to retire
 connection IDs for the initial path with Path ID 0.
 
 As the MP_NEW_CONNECTION_ID frames applies the sequence number per path,
-the sequence number in the MP_RETIRE_CONNECTION_ID frame
-also needs to be considered in the context of the Path Identifier field.
+the sequence number in the MP_RETIRE_CONNECTION_ID frame is also per
+path. The MP_RETIRE_CONNECTION_ID frame retires the Connection ID with
+the specified path ID and sequence number.
```
mirjak, line 1253:
```suggestion
the specified Path ID and sequence number.
```

draft-ietf-quic-multipath.md
```diff
@@ -1197,12 +1197,22 @@ Path Identifier:
...
+Connection IDs associated with different path IDs are not affected.
 
 Note that the NEW_CONNECTION_ID frame can only be used to issue or retire
 connection IDs for the initial path with Path ID 0.
 
+The last paragraph of {{Section 5.1.2 of QUIC-TRANSPORT}} specifies how to
+verify the Retire Prior To field of an incoming NEW_CONNECTION_ID frame.
+The same rule
+applies for MP_RETIRE_CONNECTION_ID frames, but it applies per path. After the
+multipath extension is isolated, the rule
```
mirjak, line 1213:
```suggestion
multipath extension is negotiated successfully, the rule
```

mirjak (approved):
This is a clarification and I think we can merge it.

Issue #343, opened 2024-05-25
What if SPA migration fails?


MikeBishop:
The draft currently says, from #256, that:

> They can also start using multiple paths, unless both server preferred addresses and a disable_active_migration transport parameter were provided by the server, in which case a client is forbidden to establish new paths until "after a client has acted on a preferred_address transport parameter" ([Section 18.2.](https://rfc-editor.org/rfc/rfc9000#section-18.2.) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)]).

This is nearly correct. The `disable_active_migration` transport parameter indicates that "an endpoint ... MUST NOT use a new local address when sending to the address that the peer used during the handshake." This does not prohibit the client from connecting to any other server address that it might know about.

The targeted scenario for the RFC 9000 text is an Anycast IP, where attempting to connect from a different source may land on a different destination instance. The migration to the SPA will allow a direct connection to the server that performed the handshake regardless of source address.

Notably, if the migration fails, the current text prohibits the client from *ever* establishing any other paths, which is probably not what we want.

I would suggest that we clarify that this prohibition applies only to the handshake address, and that establishment of additional paths to other addresses (potentially including the SPA) is valid immediately.

mirjak:
With the explicit path ID PR #403 now allows to open new paths even before migration to a preferred address happened. Let's confirm this at the meeting at IETF-120 and then hopefully merge and close this issue!

mirjak:
This was discussed and confirmed at the IETF-120 session. Will merge now.

PR #403, opened 2024-07-02
Describe the effect of disable_active_migration more accurately

Yanmei-Liu:
Solving issue #343

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -319,8 +319,8 @@ multipath support. They can also start using multiple paths when both endpoints
 have issued available connection IDs for at least one unused Path ID.
 If an endpoint receives a disable_active_migration transport parameter
 provided by the peer, it is forbidden to use a new local address
-to establish new paths to the same peer address as used during handshake,
-but establishment of additional paths to other peer addresses
+to establish new paths to the peer's handshake address. However,
+establishment of additional paths from any local address to other peer addresses
 (e.g carried by peer’s preferred_address) is valid immediately.
 
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -315,11 +315,14 @@ Path ID 0.
 # Path Management {#path-management}
 
 After completing the handshake, endpoints have agreed to enable
-multipath support. They can also start using multiple paths, unless both
-server preferred addresses and a disable_active_migration transport parameter
-were provided by the server, in which case a client is forbidden to establish
-new paths until "after a client has acted on a preferred_address transport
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -315,11 +315,14 @@ Path ID 0.
...
-multipath support. They can also start using multiple paths, unless both
-server preferred addresses and a disable_active_migration transport parameter
-were provided by the server, in which case a client is forbidden to establish
-new paths until "after a client has acted on a preferred_address transport
-parameter" ({{Section 18.2. of QUIC-TRANSPORT}}).
+multipath support. They can also start using multiple paths when both endpoints
+have issued available connection IDs for at least one unused Path ID.
+If endpoint receives a disable_active_migration transport parameter
+provided by the peer, it is forbidden to use a new local address
+to establish new paths for the peer address used during handshake,
```
mirjak, line 322:
```suggestion
to establish new paths to the same peer address as used during handshake,
```

mirjak:
This PR is fine but I propose to not merge it now but have some discussion at the meeting first. The use case for this as explained in issue #343 is anycast but I would like to double-check that there are no other reasons that would limit the opening of new paths before the preferred server address is used.

draft-ietf-quic-multipath.md
```diff
@@ -315,11 +315,14 @@ Path ID 0.
...
 
 After completing the handshake, endpoints have agreed to enable
-multipath support. They can also start using multiple paths, unless both
-server preferred addresses and a disable_active_migration transport parameter
-were provided by the server, in which case a client is forbidden to establish
-new paths until "after a client has acted on a preferred_address transport
-parameter" ({{Section 18.2. of QUIC-TRANSPORT}}).
+multipath support. They can also start using multiple paths when both endpoints
+have issued available connection IDs for at least one unused Path ID.
+If endpoint receives a disable_active_migration transport parameter
```
MikeBishop, line 320:
```suggestion
If an endpoint receives a disable_active_migration transport parameter
```

draft-ietf-quic-multipath.md
```diff
@@ -315,11 +315,14 @@ Path ID 0.
...
+provided by the peer, it is forbidden to use a new local address
+to establish new paths to the same peer address as used during handshake,
+but establishment of additional paths to other peer addresses
```
MikeBishop, line 323:
```suggestion
to establish new paths to the peer's handshake address. However,
establishment of additional paths from any local address to other peer addresses
```

I think we need to state that this TP also doesn't restrict the endpoint to a single local address.

MikeBishop (changes requested):

Issue #408, opened 2024-07-10
Editorial: Wrong parenthis to create ref


gloinul:
Section 3.3:
This is true whether the decision to close the path results from implicit signals such as idle time or packet losses (see {(idle-time-close}}) 

Erronous internal ref

Yanmei-Liu:
It's true. I fixed it in PR #415

PR #415, opened 2024-07-11
Editorial: fix error internal ref

Yanmei-Liu:
Fix issue #408.


huitema (commented):
Good catch. we should have found it by running the CI tests!

PR #416, opened 2024-07-11
Editorial: add reference for MP_NEW_CONN_ID frame

### LINKED ISSUE #404: Editorial: Refere to frame type
Section 1.2:
"Further, connection IDs are issued per Path ID." 

Should this refer to MP_NEW_CONNECTION_ID to be clear that there are new frames for this.

Yanmei-Liu:
resolves #404

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -202,7 +202,7 @@ The same Path ID is used in both directions to
 address a path in the new multipath control frames,
 such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
 PATH_AVAILABLE {{path-available-frame}} as well as MP_ACK {{mp-ack-frame}}.
-Further, connection IDs are issued per Path ID.
+Further, connection IDs are issued per Path ID (specified in {{mp-new-conn-id-frame}}).
 That means each connection ID is associated with exactly one path identifier
 but multiple connection IDs are usually issued for each path identifier.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -202,7 +202,7 @@ The same Path ID is used in both directions to
 address a path in the new multipath control frames,
 such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
 PATH_AVAILABLE {{path-available-frame}} as well as MP_ACK {{mp-ack-frame}}.
-Further, connection IDs are issued per Path ID.
+Further, connection IDs are issued per Path ID (specified in {{mp-new-conn-id-frame}}).
```
gloinul, line 205: I think "using" rather "specified in" here would be more correct, or possibly "see" if one to be non-specific.

Yanmei-Liu: Indeed. I updated.

gloinul (approved):

huitema (approved):
I think this is OK. I am waiting for someone to argue that once the multipath extension is negotiated, even NEW_CONNECTION_ID frames do allocate connection ID per path -- but only for path = 0...

draft-ietf-quic-multipath.md
```diff
@@ -202,7 +202,7 @@ The same Path ID is used in both directions to
 address a path in the new multipath control frames,
 such as PATH_ABANDON {{path-abandon-frame}}, PATH_STANDBY {{path-standby-frame}}},
 PATH_AVAILABLE {{path-available-frame}} as well as MP_ACK {{mp-ack-frame}}.
-Further, connection IDs are issued per Path ID.
+Further, connection IDs are issued per Path ID (using {{mp-new-conn-id-frame}}).
```
mirjak, line 205:
```suggestion
Further, connection IDs are issued per Path ID using the new
MP_NEW_CONNECTION_ID frame (see {{mp-new-conn-id-frame}}).
```

mirjak (changes requested):
See proposal.

Issue #407, opened 2024-07-10
Is using the next available new PATH ID or a new CID that is what matters?


gloinul:
Section 3.1:
"Opening a new path requires the use of a new connection ID (see Section 9.5 of [QUIC-TRANSPORT])."

I think the mentioning of "new connection ID" here is pointing to the wrong thing. Isn't what is most important here is to use a new CID for the next path ID. And here to my understanding the next Path ID is the lowest value not yet used for any path challenge and which is not currently in use on any path challenge for another 4-tuple.

PR #418, opened 2024-07-12
Editorial: fix requirement of opening a new path

Yanmei-Liu:
Fix issue #407

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -363,7 +363,7 @@ All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
 ## Path Initiation {#path-initiation}
 
-Opening a new path requires the use of a new connection ID
+Opening a new path requires the use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}})
 for an unused Path ID (see {{explicit-path-id}}).
 Instead of NEW_CONNECTION_ID frame as specified in {{Section 19.15 of QUIC-TRANSPORT}},
 each endpoint uses the MP_NEW_CONNECTION_ID frame as specified in this extension
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -364,7 +364,7 @@ All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 ## Path Initiation {#path-initiation}
 
 Opening a new path requires the use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}})
-for an unused Path ID (see {{explicit-path-id}}).
+mapped to an unused Path ID (see {{explicit-path-id}}).
 Instead of NEW_CONNECTION_ID frame as specified in {{Section 19.15 of QUIC-TRANSPORT}},
 each endpoint uses the MP_NEW_CONNECTION_ID frame as specified in this extension
 to issue Path ID-specific connections IDs.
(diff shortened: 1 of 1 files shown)


huitema (changes requested):
I don't think this change is necessary. It is also not purely editorial. We discussed that before. We recommend that NEW_CONNECTION_ID frames be sent in order of increasing path ID, but we can never guarantee that they will be *received* in order. Suppose the server send (C1, pathID 1) and (C2, pathID 2), and that C1 is delayed, maybe by packet loss. Your new text will say that the client MUST wait until C1 is received to open a new path. That's wrong.

Also, it is better to have all the discussion of order, etc., in exactly one place. The previous text is better.

Yanmei-Liu:
> I don't think this change is necessary. It is also not purely editorial. We discussed that before. We recommend that NEW_CONNECTION_ID frames be sent in order of increasing path ID, but we can never guarantee that they will be _received_ in order. Suppose the server send (C1, pathID 1) and (C2, pathID 2), and that C1 is delayed, maybe by packet loss. Your new text will say that the client MUST wait until C1 is received to open a new path. That's wrong.
> 
> Also, it is better to have all the discussion of order, etc., in exactly one place. The previous text is better.

@huitema But in [Section 3.5](https://datatracker.ietf.org/doc/html/draft-ietf-quic-multipath-10#section-3.5) we use "Similarly, endpoints SHOULD consume Path IDs in a continuous way, i.e., when creating paths." I think in the case of retransmission, endpoints should choose to wait for C1 to create the new path, using C2 would break the "SHOULD". 

BTW, this issue and pull request is to solve the true requirement of new path initialization is the use of new Path ID. I changed the description of "the smallest unused Path ID" to "an unused Path ID". Please check again.

mirjak (approved):

draft-ietf-quic-multipath.md
```diff
@@ -363,8 +363,8 @@ All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
 ## Path Initiation {#path-initiation}
 
-Opening a new path requires the
-use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}}).
+Opening a new path requires the use of a new connection ID
+for an unused Path ID (see {{explicit-path-id}}).
```
mirjak, line 367:
```suggestion
Opening a new path requires the use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}})
for an unused Path ID (see {{explicit-path-id}}).
```
I think it is actually not correct to remove the reference to section 9.5 of RFC9000 as that section explain the privacy implications and we don't have a reference to 9.5 of RFC9000 in the section on explicit path ID. However, I think we can have both references.

mirjak (changes requested):
I think we still need the reference to sec 9.5 of RFC9000

draft-ietf-quic-multipath.md
```diff
@@ -363,8 +363,8 @@ All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
 ## Path Initiation {#path-initiation}
 
-Opening a new path requires the
-use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}}).
+Opening a new path requires the use of a new connection ID (see {{Section 9.5 of QUIC-TRANSPORT}})
+for an unused Path ID (see {{explicit-path-id}}).
```
qdeconinck, line 367:
```suggestion
mapped to an unused Path ID (see {{explicit-path-id}}).
```

qdeconinck (approved):
small suggestion

huitema (approved):

PR #419, opened 2024-07-18
SHOULD use MP frame for path ID 0

### LINKED ISSUE #405: Editorial: SHOULD for the wrong aspect on when to use MP_*_CONNECTION_ID frames
Section 3:
"Endpoints SHOULD use the MP_NEW_CONNECTION_ID frame to provide new connection IDs and, respectively, the MP_RETIRE_CONNECTION_ID frame to retire connection IDs after a successful handshake indicating multipath support by both endpoints."

The SHOULD appear to point to the wrong aspect. One is required to use MP_NEW_CONNECTION_ID frame to provide new CIDs, what is SHOULD is to do it after a successful handshake. So sentence need reformulation.

mirjak:
fixes #405

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -317,7 +317,7 @@ Path ID 0.
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use MP_ACK frames instead of ACK frames,
 also for acknowledging so far unacknowledged 0-RTT packets, using
-Path ID 0. Respectively after a successful handsaheke, endpoints SHOULD also use
+Path ID 0. Similarly after a successful handshake, endpoints SHOULD also use
 the MP_NEW_CONNECTION_ID frame to provide new connection IDs for Path ID 0 and,
 respectively, the MP_RETIRE_CONNECTION_ID frame to retire connection IDs for
 Path ID 0.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -317,6 +317,9 @@ Path ID 0.
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use MP_ACK frames instead of ACK frames,
 also for acknowledging so far unacknowledged 0-RTT packets, using
+Path ID 0. Respectively after a successful handsaheke, endpoints SHOULD also use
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -339,13 +342,9 @@ Furthermore, this document
...
 
-To let the peer open a new path, an endpoint needs to provide its peer with connection IDs
-for at least one unused path identifier. Endpoints SHOULD use the MP_NEW_CONNECTION_ID
-frame to provide new connection IDs and, respectively, the MP_RETIRE_CONNECTION_ID frame to
-retire connection IDs after a successful handshake indicating multipath support by both endpoints.
-
 To open a new path, an endpoint MUST use a connection ID associated with
-a new, unused Path ID.
+a new, unused Path ID. To let the peer open a new path, an endpoint needs
+to provide its peer with connection IDs for at least one unused path identifier.
```
mirjak, line 347:
```suggestion
to provide its peer with connection IDs for at least one unused path identifier.
```

draft-ietf-quic-multipath.md
```diff
@@ -317,6 +317,9 @@ Path ID 0.
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use MP_ACK frames instead of ACK frames,
 also for acknowledging so far unacknowledged 0-RTT packets, using
+Path ID 0. Respectively after a successful handsaheke, endpoints SHOULD also use
```
gloinul, line 320: Another strange use of respectively. I think for this one case just short the sentence: 
Endpoints SHOULD also use
the MP_NEW_CONNECTION_ID frame to provide new connection IDs and,
the MP_RETIRE_CONNECTION_ID frame to retire connection IDs, respectively for Path ID 0.

mirjak, line 320:
```suggestion
Path ID 0. Similarly after a successful handshake, endpoints SHOULD also use
```

mirjak:
@Yanmei-Liu this PR does not change the SHOULD. The SHOULD already in the current version of the draft. It only moves the text and rephrases it slightly. If we want to change the SHOULD that would be not editorial and need to go into a separate PR. However, we need more discussion as I don't think we have agreement. See also issue #220 which is related for MP_ACK.

qdeconinck (approved):

PR #427, opened 2024-09-02
Rename PATH_STANDBY to PATH_BACKUP

mirjak:
fixes issue #283


huitema (commented):
Good, but you need to fix the typos, and probably also update the internal references like path-standby-frame to path-backup-frame, and fig-path-standby-format to fig-path-backup-format

mirjak:
Thanks @huitema ! I fixed the typo and reference!

qdeconinck (approved):

PR #429, opened 2024-09-03
Do not delete both duplicate paths

### LINKED ISSUE #378: Do we need further guidance if multiple paths over the same 4-tuple are used?

huitema:
Attempt at providing further guidance and point out risks if both endpoints attempt to close one of the duplicate paths at the same time. Also mention that there will be cases where having "duplicate" path is intentional, maybe because they have different diffserv markings.

Close #378

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -620,6 +620,23 @@ to abandon a path for any reason, for example, removing a hole from
 the sequence of Path IDs in use. This is not an error. The endpoint that
 receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
 
+### Closing Duplicate Paths
+
+As noted in {{basic-design-points}}, it is possible to create paths that
+refer to the same 4-tuple. There will be cases where this is intentional,
+for example if the paths use different Differentiated Service markings.
+There may also be cases where this is not intentional, maybe as a result
+of path migration. 
+
+In the non-intentional case, the endpoints
+may want to abandon one of the paths and keep the other, but
+uncoordinated Abandon from both ends of the connection may result in deleting
+two paths instead of just one. This sort of error will be avoided if only
+the client initiates the closing of duplicate paths. Applications may override
+that behavior if they can coordinate path closures at part of the
+application protocol.
+
+
 ## Refusing a New Path
 
 An endpoint may deny the establishment of a new path initiated by its
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -620,23 +620,6 @@ to abandon a path for any reason, for example, removing a hole from
 the sequence of Path IDs in use. This is not an error. The endpoint that
 receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
 
-### Closing Duplicate Paths
-
-As noted in {{basic-design-points}}, it is possible to create paths that
-refer to the same 4-tuple. There will be cases where this is intentional,
-for example if the paths use different Differentiated Service markings.
-There may also be cases where this is not intentional, maybe as a result
-of path migration.
-
-In the non-intentional case, the endpoints
-may want to abandon one of the paths and keep the other, but
-uncoordinated Abandon from both ends of the connection may result in deleting
-two paths instead of just one. To avoid this pitfall, only the client
-SHOULD initiate the closing of duplicate paths. Applications MAY
-override this default behavior if they can coordinate path closures at part of the
-application protocol.
-
-
 ## Refusing a New Path
 
 An endpoint may deny the establishment of a new path initiated by its
@@ -1018,6 +1001,36 @@ a peer address if that address has been seen recently. However, when the
 multipath extension is used and an endpoint has multiple addresses that
 could lead to switching between different paths, it should rather maintain
 multiple open paths instead.
+
+### Using multiple paths on the same 4-tuple
+
+As noted in {{basic-design-points}}, it is possible to create paths that
+refer to the same 4-tuple. For example, the endpoints may want
+to create paths that use different Differentiated Service markings.
+This could be done in conjunction with scheduling algorithms
+that match streams to paths, so that for example data frame for
+low priority streams are sent over low priority paths.
+Since these paths use different path IDs, they can be managed
+independently to suit the needs of the application.
+
+There may be cases in which paths are created with different 4-tuples,
+but end up using the same four tuples as a consequence of path
+migrations. For example:
+
+* Client starts path 1 from address 192.0.2.1 to server address 198.51.100.1
+* Client starts path 2 from address 192.0.2.2 to server address 198.51.100.1
+* both paths are used for a while.
+* Server sends packet from address 198.51.100.1 to client address 192.0.2.1, with CID indicating path=2.
+* Client receives packet, recognizes a path migration, update source address of path 2 to 192.0.2.1.
+
+Such unintentional use of the same 4-tuple on different paths ought to
+be rare. When they happen, the two paths would be redundant, and the
+endpoint will want to close one of them. 
+Uncoordinated Abandon from both ends of the connection may result in deleting
+two paths instead of just one. To avoid this pitfall, endpoints could
+adopt a simple coordination rule, such as only letting the client
+initiate closure of duplicate paths, or perhaps relying on
+the application protocol to decide which paths should be closed.
 
 # New Frames {#frames}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -620,6 +620,23 @@ to abandon a path for any reason, for example, removing a hole from
...
+refer to the same 4-tuple. There will be cases where this is intentional,
+for example if the paths use different Differentiated Service markings.
+There may also be cases where this is not intentional, maybe as a result
+of path migration.
+
+In the non-intentional case, the endpoints
+may want to abandon one of the paths and keep the other, but
+uncoordinated Abandon from both ends of the connection may result in deleting
+two paths instead of just one. This sort of error will be avoided if only
+the client initiates the closing of duplicate paths. Applications may override
```
martinthomson, line 635: There are other ways to avoid problems.  For instance, you might institute a rule that the lowest-numbered path should be closed in the case of duplicates.  This "if only" isn't "only if", but maybe avoid using that sort of phrasing to avoid creating confusion about what you mean.

It's also unclear whether you are recommending that clients be responsible for path closure or whether this is a "for example".  That's a specific manifestation of the more general problem that we don't really know what processes drive the creation and teardown of paths.  If you lack that broader framework, the best you can do is this sort of non-specific advice.

huitema: Yes. I think that the creation of paths ought to be controlled by the application. Stacks could read what alternate IP addresses are available and just try them all, but the failure modes are easy to imagine. Plus, applications will want to specify the scheduling of packet on paths, for example setting affinity between stream and paths.

The problem with rules like "lowest number first" is that the two endpoints need to agree on the rule. In our current draft, paths are always created by the client. That does not mean that servers should never abandon paths, but they should do that mostly if the path appears broken. In the duplicate example, the two paths are both valid, so there is no obvious criteria. Letting the client chose seems the simplest solution.

huitema: But yes, the sentence could be written better.

draft-ietf-quic-multipath.md
```diff
@@ -620,6 +620,23 @@ to abandon a path for any reason, for example, removing a hole from
 the sequence of Path IDs in use. This is not an error. The endpoint that
 receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
 
+### Closing Duplicate Paths
+
+As noted in {{basic-design-points}}, it is possible to create paths that
+refer to the same 4-tuple. There will be cases where this is intentional,
+for example if the paths use different Differentiated Service markings.
+There may also be cases where this is not intentional, maybe as a result
+of path migration.
```
mirjak, line 629: I don't think this unintentional case due to path migration is possible anymore because you either use a new path ID or not. This makes it crystal clear to distinguishes between path migration and opening a new path.

draft-ietf-quic-multipath.md
```diff
@@ -620,6 +620,23 @@ to abandon a path for any reason, for example, removing a hole from
 the sequence of Path IDs in use. This is not an error. The endpoint that
 receive such a PATH_ABANDON frame must treat it as specified in {{path-close}}.
 
+### Closing Duplicate Paths
```
mirjak, line 623: I think I would rather call this section something like "Using multiple paths on the same 4-tuple" and move it into the "Implementation Considerations" section?

draft-ietf-quic-multipath.md
```diff
@@ -620,6 +620,23 @@ to abandon a path for any reason, for example, removing a hole from
...
+There may also be cases where this is not intentional, maybe as a result
+of path migration.
+
+In the non-intentional case, the endpoints
+may want to abandon one of the paths and keep the other, but
+uncoordinated Abandon from both ends of the connection may result in deleting
+two paths instead of just one. To avoid this pitfall, only the client
+SHOULD initiate the closing of duplicate paths. Applications MAY
+override this default behavior if they can coordinate path closures at part of the
+application protocol.
```
mirjak, line 637: Not sure if this normative language makes really sense to me. If you clients decided to close one of paths on the same 4-tuple, why did it open it in the first place? Or maybe the client would for some reason like to have two paths on the same 4-tuple but the server really doesn't want to have that for some other reasons. In this case I think the server will still try to close a path.

draft-ietf-quic-multipath.md
mirjak: To add to this discussion. I also think a rule like lowest number path or actually oldest path would make more sense. If the client don't want to have the new path, it would not have opened it. however, it make open the new path and then plan to close the old one for some reason. On the other hand, if the old one is in active use and the new one not yet, that it's maybe not optional to close the old. So in summary not sure if we can really give good advise here or just simply with down these consideration (and implicitly recommend to not close any of the paths)?

draft-ietf-quic-multipath.md
huitema: Example:

* Client starts path 1 from address 1.1.1.1 to server address 5.5.5.5
* Client starts path 2 from address 2.2.2.2 to server address 5.5.5.5
* both paths are used for a while.
* Server sends packet from address 5.5.5.5 to client address 1.1.1.1, with CID indicating path=2.
* Client receives packet, recognizes a path migration, update source address of path 2 to 1.1.1.1
* Bingo, path 1 and path 2 refer to the same 4 tuple.

draft-ietf-quic-multipath.md
huitema: I agree that this might be a better place. We are going to end up with fuzzy language, so implementation considerations is more adequate.

draft-ietf-quic-multipath.md
huitema: See note above on plausible path migration.

huitema: I removed the normative language after moving the text to the implementation section.

draft-ietf-quic-multipath.md
huitema: Just moved the text to the implementation section.

draft-ietf-quic-multipath.md
huitema: Example added as part of the implementation section.

draft-ietf-quic-multipath.md
huitema: I think I fixed that when moving the text to the implementation section.

draft-ietf-quic-multipath.md
```diff
@@ -1002,6 +1002,36 @@ multipath extension is used and an endpoint has multiple addresses that
...
+* Client starts path 2 from address 192.0.2.2 to server address 198.51.100.1
+* both paths are used for a while.
+* Server sends packet from address 198.51.100.1 to client address 192.0.2.1, with CID indicating path=2.
+* Client receives packet, recognizes a path migration, update source address of path 2 to 192.0.2.1.
+
+Such unintentional use of the same 4-tuple on different paths ought to
+be rare. When they happen, the two paths would be redundant, and the
+endpoint will want to close one of them.
+Uncoordinated Abandon from both ends of the connection may result in deleting
+two paths instead of just one. To avoid this pitfall, endpoints could
```
Yanmei-Liu, line 1030: As we have Path Identifier (i) field in Path Abandon frame, endpoints could use Path Identifier (i) to find and delete the corresponding path context, instead of using 4-tuple which might cause unexpected path closure.

huitema: That's not the issue. The issue is that the two endpoint each issue a different path abandon, and end up with nothing.

qdeconinck (approved):

draft-ietf-quic-multipath.md
```diff
@@ -986,6 +986,59 @@ two or more active paths during the connection lifetime. Different applications
...
+a peer address if that address has been seen recently. However, when the
+multipath extension is used and an endpoint has multiple addresses that
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead.
+
+### Using multiple paths on the same 4-tuple
+
+As noted in {{basic-design-points}}, it is possible to create paths that
+refer to the same 4-tuple. For example, the endpoints may want
+to create paths that use different Differentiated Service markings.
```
mirjak, line 1015:
```suggestion
to create paths that use different Differentiated Service {{?RFC2475}} markings.
```

PR #430, opened 2024-09-04
Anti-amplification guidance

### LINKED ISSUE #422: Anti-amplification limits
RFC 9000 requires that:
- Datagrams carrying Initial packets be expanded to 1200+ bytes
- Servers verify the client address on handshake *and* when the address changes through migration / rebinding
- The server can send no more than 3x the received bytes until the client's address is validated, both during the handshake and following client migration / rebinding

This draft states in the Security Considerations that "the anti-amplification limits as specified in [Section 8](https://rfc-editor.org/rfc/rfc9000#section-8) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)] need to be followed to limit the amplification risk." However, there's no text that describes how this maps into multipath behavior.

My proposal is that:
- The packet on a new path SHOULD be expanded to 1200+ bytes (PADDING, etc.) for the first 1 RTT or until validation completes
- When a packet is received from an unverified address, an endpoint MUST NOT send more than 3x the received bytes until address validation completes

MikeBishop:
Fixes #422.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -379,13 +379,18 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same Path ID for sending the PATH_RESPONSE.
 
 When the multipath extension is negotiated, a client that wants to use an
-additional path MUST first initiate the Address Validation procedure
-with PATH_CHALLENGE and PATH_RESPONSE frames as described in
-{{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
-that address. After receiving packets from the
+additional path MUST first initiate address validation as described in
+{{Section 8.2 of QUIC-TRANSPORT}} unless it has previously validated
+that address. Note that datagrams that contain a PATH_CHALLENGE frame are
+expanded to at least the smallest allowed maximum datagram size of 1200 bytes;
+see {{Section 8.2.1 of QUIC-TRANSPORT}}.
+
+After receiving packets from the
 client on a new path, if the server decides to use the new path,
-the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
-unless it has previously validated that address.
+the server MUST perform address validation ({{Section 8.2 of QUIC-TRANSPORT}})
+unless it has previously validated that address. Until the client's address is
+validated, the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
+applies.
 
 MP_ACK frames (defined in {{mp-ack-frame}}) can be returned on any path.
 If the MP_ACK is preferred to be sent on the same path as the acknowledged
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -379,13 +379,18 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
...
-additional path MUST first initiate the Address Validation procedure
-with PATH_CHALLENGE and PATH_RESPONSE frames as described in
-{{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
-that address. After receiving packets from the
+additional path MUST first initiate address validation as described in
+{{Section 8.2 of QUIC-TRANSPORT}} unless it has previously validated
+that address. Note that datagrams that contain a PATH_CHALLENGE frame are
+expanded to at least the smallest allowed maximum datagram size of 1200 bytes;
+see {{Section 8.2.1 of QUIC-TRANSPORT}}.
+
```
huitema, line 387: I have mixed feelings about the "note" -- it is restating the obvious. Whenever we do that, we create ambiguity. "The multipath extension spec makes a direct reference to 8.2.1, so surely that applies, but what about 8.2.2, 3 and 4? They are not called out, so mayb they can be ignored."

draft-ietf-quic-multipath.md
```diff
@@ -1405,11 +1410,10 @@ such as e.g. for packet number handling, only after path validation has successf
...
 
-The path validation mechanism as specified in {{Section 8.2. of QUIC-TRANSPORT}} for migration is used
-unchanged for initiation of new paths in this extension. Respectively the security considerations
-on source address spoofing as outlined in {{Section 21.5.4 of QUIC-TRANSPORT}} equally apply.
-Similarly, the anti-amplification limits as specified in {{Section 8 of QUIC-TRANSPORT}} need to be
-followed to limit the amplification risk.
+The path validation mechanisms and anti-amplification limits specified in
+{{Section 8 of QUIC-TRANSPORT}} for migration also apply to initiation of new
+paths in this extension. Therefore, the security considerations on source
+address spoofing in {{Section 21.5.4 of QUIC-TRANSPORT}} equally apply.
```
huitema, line 1417: If we have a reference there, maybe we do not need to also have it in the path initiation section.

huitema (commented):
I am not a great fan of belaboring the obvious, because it can confuse as many people as it helps.

draft-ietf-quic-multipath.md
MikeBishop: As this is the one which is most obvious from RFC9000 that it would always apply, this is probably the one I'd be most willing to leave out. It's a property of anything carrying a PATH_CHALLENGE at any time.

Do you feel the same way about what's split off into the next paragraph, to call out anti-amplification explicitly? I feel like that's less obvious and more worth emphasizing.

huitema: I don't know why it should be less obvious, and in any case you are calling in out in the security considerations.

mirjak: So I think this point (and specifically this sentences) really touched right on the more fundamental discussion we had during the IETF-120 session. There we said it is a requirement to validate a new path before you can use it. How you do that (sending path challenge, padding and amplification limits) is not specific to the multipath extension and already specified sufficiently in RFC9000.

Part of the discuss was that people want to still have a MUST in this draft to be explicit on the requirement that you have to validate the path before use (also for a new path and not only migration) but we should specify less of the details. That means we have to revise this sentence in any case. I guess we could maybe still also note somehow that RFC9000 considers padding to ensure that the new paths can support size requirement and there is an limit to avoid amplification attacks but it should really more being phrased as providing pointers than actual requirements.

MikeBishop: I think that gets into what it means to "use" a path. The anti-amplification limit might be moot if you're not allowed to send real data ("use the path") until it's validated, but I don't see that in RFC 9000. Regular QUIC says that clients "MAY probe for peer reachability ... prior to migrating" and "an endpoint MAY send data to an unvalidated peer address" after a potential migration, which suggests that any requirement to complete validation before "use" is new in this document.

We've eliminated the probing/non-probing distinction, and a requirement to complete validation before using the path seems like a return to that distinction.

Regardless, my intention with these sections was to provide pointers and clarify requirements that do still apply.

draft-ietf-quic-multipath.md
MikeBishop: Perhaps, though I suspect putting it there is more likely to catch the attention of someone implementing that flow.

draft-ietf-quic-multipath.md
mirjak: yes, the requirement on validation is new in some sense and we need to clarify this and that was the outcome of the IETF-120 discussion. However, I don't think this changes the requirement on amplification limits or padding.

mirjak:
Closing this PR as we are planning to merge #442 instead. Thanks anyway @MikeBishop !

PR #434, opened 2024-09-09
TRANSPORT_PARAMETER_ERROR when initial_max_path is too large

### LINKED ISSUE #432: When initial_max_path_id exceeds 2^32-1, which error code should be used?
If we follow the conventions of RFC 9000, should we use TRANSPORT_PARAMETER_ERROR? If so, should we explicitly document this?

[7.4. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-7.4)[Transport Parameters](https://www.rfc-editor.org/rfc/rfc9000.html#name-transport-parameters)

> An endpoint MUST treat receipt of a transport parameter with an invalid value as a connection error of type TRANSPORT_PARAMETER_ERROR.

By the way, would it be more appropriate to use **TRANSPORT_PARAMETER_ERROR** instead of [**TRANSPORT_PARAMETER error**](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-2-11)?

> Cipher suites with a nonce shorter than 12 bytes cannot be used together with the multipath extension. If such a cipher suite is selected and the use of the multipath extension is negotiated, endpoints MUST abort the handshake with a **TRANSPORT_PARAMETER error**.

mirjak:
fixes #432 

This PR defines an error for the case when the initial_max_path value exceeds 2^32-1. As we changes the transport parameter meaning with the introduction of the explicit path ID, it was an oversight to not define an error here. Only remaining question is if TRANSPORT_PARAMETER_ERROR is the right error type?

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -263,7 +263,7 @@ associated with Path IDs 0 and 1 should be issued by the peer.
 If an endpoint receives an initial_max_path_id transport parameter with value 0,
 the peer aims to  enable the multipath extension without allowing extra paths immediately.
 
-If an initial_max_path_id transport parameter value that is larger than 2^32-1
+If an initial_max_path_id transport parameter value that is higher than 2^32-1
 is received, the receiver MUST close the connection with an error of type
 TRANSPORT_PARAMETER_ERROR.
 


draft-ietf-quic-multipath.md
```diff
@@ -263,6 +263,10 @@ associated with Path IDs 0 and 1 should be issued by the peer.
 If an endpoint receives an initial_max_path_id transport parameter with value 0,
 the peer aims to  enable the multipath extension without allowing extra paths immediately.
 
+If an initial_max_path_id transport parameter value that is larger than 2^32-1
```
mirjak, line 266:
```suggestion
If an initial_max_path_id transport parameter value that is higher than 2^32-1
```

huitema (approved):

Yanmei-Liu (approved):

PR #440, opened 2024-10-01
new text on idle timeouts

### LINKED ISSUE #397: What to do on path time-out?
With use of multipath you have a multiple path open but not use all of them concurrently (e.g. for stand-by). Therefore a path time out makes less sense. We probably need to consider if the path if actively used or not. Only you send on a path and all packets are finally marked as lost, that's probably a good indication that the path Is broken and you should close it. If a path is not used, you still might want to keep it open to use it later. Not sure if requirement to send pings on all non-used path is that useful.

Another related issue is that if you only send non-ack eliciting packets on a path (like ACKs) and the path is not used by the peer for sending, you might not be able to detect a path breakage. May it would make sense to require sending ack electing frames on all used paths from time to time?

mirjak:
fixes #397

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
 When only one path is available, servers MUST follow the specifications
 in {{QUIC-TRANSPORT}}.
 
-When more than one path is available, hosts shall monitor the arrival of
-packets and acknowledgements for packets sent over each
-path. Hosts MAY consider closing a path if for at least the period of the
-idle timeout as specified in {{Section 10.1. of QUIC-TRANSPORT}}
-(a) no packet was received or (b) no packet sent over this path was acknowledged.
-Endpoints that desire to close a path because of the idle timer rule
-MUST do so explicitly by sending a PATH_ABANDON frame on another path,
-as defined in {{path-close}}.
+This dcoument does not specify any specific per-path timeouts. An endpoint
+can decide to close a path at any time not matter if the path is in active
+use or not by sending a PATH_ABANDON frame. Particulary it is not required
+to send a PATH_ABANDON frame at any specific point in time if that would be
+the only frame to send.
 
-To avoid idle timeout of a path, endpoints
+If a path is not actively used for a while, it might not be usable anymore,
+e.g. due to middlebox timeouts. To avoid such path breakage, endpoints
 can send ack-eliciting packets such as packets containing PING frames
-({{Section 19.2 of QUIC-TRANSPORT}}) on that path to keep it alive.  Sending
-periodic PING frames also helps prevent middlebox timeout, as discussed in
-{{Section 10.1.2 of QUIC-TRANSPORT}}.
+({{Section 19.2 of QUIC-TRANSPORT}}) on that path to keep it alive.
+As discussed in
+{{Section 10.1.2 of QUIC-TRANSPORT}}, the keep-alive interval depends
+on the timeout in the middlebox.
 
-Server implementations need to select the sub-path idle timeout as a
-trade-off between keeping resources, such as connection IDs, in use
-for an excessive time or having to promptly re-establish a path
-after a spurious estimate of path abandonment by the client.
+If a path was not actively used for a while, an endpoint might consider
+to probe it before switching to active use if there are still other paths
+that are currently usable.
 
 ### Early Abandon {#abandon-early}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -606,8 +606,8 @@ As discussed in
 {{Section 10.1.2 of QUIC-TRANSPORT}}, the keep-alive interval depends
 on the timeout in the middlebox.
 
-If a path was not actively used for a while, an endpoint might consider
-to probe it before switching to active use if there are still other paths
+If a path was not actively used for a while, an endpoint can
+probe it before switching to active use if there are still other paths
 that are currently usable.
 
 ### Early Abandon {#abandon-early}
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -593,8 +593,8 @@ When only one path is available, servers MUST follow the specifications
 in {{QUIC-TRANSPORT}}.
 
 This dcoument does not specify any specific per-path timeouts. An endpoint
-can decide to close a path at any time not matter if the path is in active
-use or not by sending a PATH_ABANDON frame. Particulary it is not required
+can decide to close a path at any time, whether the path is in active
+use or not, by sending a PATH_ABANDON frame. Particularly, it is not required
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
...
-idle timeout as specified in {{Section 10.1. of QUIC-TRANSPORT}}
-(a) no packet was received or (b) no packet sent over this path was acknowledged.
-Endpoints that desire to close a path because of the idle timer rule
-MUST do so explicitly by sending a PATH_ABANDON frame on another path,
-as defined in {{path-close}}.
-
-To avoid idle timeout of a path, endpoints
+This dcoument does not specify any specific per-path timeouts. An endpoint
+can decide to close a path at any time not matter if the path is in active
+use or not by sending a PATH_ABANDON frame. Particulary it is not required
```
MikeBishop, line 597:
```suggestion
can decide to close a path at any time, whether the path is in active
use or not, by sending a PATH_ABANDON frame. Particularly, it is not required
```

draft-ietf-quic-multipath.md
```diff
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
...
-trade-off between keeping resources, such as connection IDs, in use
-for an excessive time or having to promptly re-establish a path
-after a spurious estimate of path abandonment by the client.
+({{Section 19.2 of QUIC-TRANSPORT}}) on that path to keep it alive.
+As discussed in
+{{Section 10.1.2 of QUIC-TRANSPORT}}, the keep-alive interval depends
+on the timeout in the middlebox.
+
+If a path was not actively used for a while, an endpoint might consider
+to probe it before switching to active use if there are still other paths
```
MikeBishop, line 610:
```suggestion
If a path was not actively used for a while, an endpoint can
probe it before switching to active use if there are still other paths
```

MikeBishop (approved):

draft-ietf-quic-multipath.md
```diff
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
...
-Endpoints that desire to close a path because of the idle timer rule
-MUST do so explicitly by sending a PATH_ABANDON frame on another path,
-as defined in {{path-close}}.
-
-To avoid idle timeout of a path, endpoints
+This dcoument does not specify any specific per-path timeouts. An endpoint
+can decide to close a path at any time, whether the path is in active
+use or not, by sending a PATH_ABANDON frame. Particularly, it is not required
+to send a PATH_ABANDON frame at any specific point in time if that would be
+the only frame to send.
```
huitema, line 599: Typo dcoument 

I think we could rewrite the second sentence a little bit. Maybe: 
"It is not required to send a PATH_ABANDON frame at any specific point in time. For example,
the endpoint may want to wait for until it is not the only frame to send."

huitema (approved):
Looks good. Minor editorial suggestion.

draft-ietf-quic-multipath.md
```diff
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
...
-packets and acknowledgements for packets sent over each
-path. Hosts MAY consider closing a path if for at least the period of the
-idle timeout as specified in {{Section 10.1. of QUIC-TRANSPORT}}
-(a) no packet was received or (b) no packet sent over this path was acknowledged.
-Endpoints that desire to close a path because of the idle timer rule
-MUST do so explicitly by sending a PATH_ABANDON frame on another path,
-as defined in {{path-close}}.
-
-To avoid idle timeout of a path, endpoints
+This dcoument does not specify any specific per-path timeouts. An endpoint
```
mirjak, line 595:
```suggestion
This document does not specify any specific per-path timeouts. An endpoint
```

draft-ietf-quic-multipath.md
mirjak, line 599:
```suggestion
use or not, by sending a PATH_ABANDON frame. It is not required
to send a PATH_ABANDON frame at any specific point in time if that would be
the only frame to send. For example, an endpoint may want to wait until
it will anyway send another frame.
```

draft-ietf-quic-multipath.md
```diff
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
...
-Server implementations need to select the sub-path idle timeout as a
-trade-off between keeping resources, such as connection IDs, in use
-for an excessive time or having to promptly re-establish a path
-after a spurious estimate of path abandonment by the client.
+({{Section 19.2 of QUIC-TRANSPORT}}) on that path to keep it alive.
+As discussed in
+{{Section 10.1.2 of QUIC-TRANSPORT}}, the keep-alive interval depends
+on the timeout in the middlebox.
+
+If a path was not actively used for a while, an endpoint can
```
mirjak, line 609: We switched here from "might consider" to "can". However, with "can" the sentence seems incorrect to me as we are not giving an advise anymore. So yes, the endpoint can do it but why are we talking about it...? I'm not a native speaker but I would think "might" would make slightly more sense...? Or do we need to re-phrase this entirely?

huitema: I think this is OK. We might say "endpoints could send", to emphasize that this is some optional behavior. We certainly don't want to make a specific recommendation like 'send every 15 seconds', because the behavior of middleboxes is unpredictable.

draft-ietf-quic-multipath.md
```diff
@@ -592,25 +592,23 @@ as "no packet received on any path for the duration of the idle timeout".
...
-Endpoints that desire to close a path because of the idle timer rule
-MUST do so explicitly by sending a PATH_ABANDON frame on another path,
-as defined in {{path-close}}.
-
-To avoid idle timeout of a path, endpoints
+This document does not specify any specific per-path timeouts. An endpoint
+can decide to close a path at any time, whether the path is in active
+use or not, by sending a PATH_ABANDON frame. It is not required
+to send a PATH_ABANDON frame at any specific point in time.
+For example, an endpoint may wait until it will anyway send another frame.
```
yfmascgy, line 599: > It is not required to send a PATH_ABANDON frame at any specific point in time.

Do you mean "An endpoint does not need to send a PATH_ABANDON frame immediately or at any particular moment"?

qdeconinck (approved):

draft-ietf-quic-multipath.md
mirjak: yes, one endpoint decides to close a path and then need to send an abandon frame, however, there is nothing in this spec that says it has to take that decision as an specific point in time, e.g. as react to a certain event. For example even if you receive a packet on a new path that you don't want to open, it doesn't necessarily mean that you have to close it immediately, because opening a path (by receiving) a packet is an independent event from closing the path, meaning as you as you have received the path it is effetely on open path and similarly as you can decide to close any path at any point of time, you simply also decide to close that path short after opening it. Sorry for using these many works but I hope that makes my mental model more clear....?

PR #442, opened 2024-10-01
packets without challenge are valid

### LINKED ISSUE #390: What if a packet with a CID belonging to a new Path ID without path challenge is received?
Now that we have an explicit Path ID, it might be easier to expect that a packet with a new path ID has to also contain a path challenge. Therefore if a packet for a non-validated path ID is received that does not contain a path challenge, we could say it should be ignore. Or we could also say that the receiving peer should send a path challenge. But I think we have to same something more.

huitema:
Close #390

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1443,6 +1443,7 @@ additional measures to limit the number of concurrent path validation processes
 e.g. by pacing them out or limiting the number of path initiation attempts
 over a certain time period.
 
+
 ## Use of Transport Layer Security and the AEAD Encryption Nonce
 
 The multipath extension as specified in this document is only enabled after a
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -112,11 +112,8 @@ non-zero connection IDs in order to identify the path and respective
 packet number space.
 
 To add a new path to an existing QUIC connection with multipath support,
-a client starts a path validation on
+a client starts a path validation on or send a packet
 the chosen path, as further described in {{path-initiation}}.
-A new path can only be used once the associated 4-tuple has been validated
-by ensuring that the peer is able to receive packets at that address
-(see {{Section 8 of QUIC-TRANSPORT}}).
 In this version of the document, a QUIC server does not initiate the creation
 of a path, but it can validate a new path created by a client.
 
@@ -381,16 +378,17 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same Path ID for sending the PATH_RESPONSE.
 
 When the multipath extension is negotiated, a client that wants to use an
-additional path MUST first initiate the Address Validation procedure
+additional path SHOULD first initiate the Address Validation procedure
 with PATH_CHALLENGE and PATH_RESPONSE frames as described in
-{{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
+{{Section 8.2 of QUIC-TRANSPORT}}. It MAY skip the validation and
+simply send packets on the paths if it has previously validated
 that address.
 
 After receiving packets from the
 client on a new path, if the server decides to use the new path,
-the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
+the server SHOULD perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
-An endpoint that receives a PATH_CHALLENGE and does not want to establish
+An endpoint that receives packets on a new path and does not want to establish
 this path is expected to close the path by sending a PATH_ABANDON
 on another path, as specified in section {{path-close}}.
 
@@ -443,18 +441,6 @@ i.e., pick the last received token. To avoid issues when clients make the "wrong
 a server should issue a token that is capable of validating
 any of the previously validated addresses. Further guidance on token usage can be
 found in {{Section 8.1.3 of QUIC-TRANSPORT}}.
-
-### Immediate path initiation
-
-The server may receive packets for a yet unused Path ID that do not
-contain a path challenge. Such packets are valid if the can be properly decrypted
-and if they contain a valid connection ID. ({{path-initiation}}
-specifies that the client may send data packets immediately if "it has previously
-validated that address".) The server SHOULD create a new path, accept the
-packets arriving on that path from the client, and acknowledge them.
-
-If the server decides to not accept the packet, it MUST close the path as
-specified in {{path-close}}.
 
 ## Path Status Management
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -382,12 +382,15 @@ additional path SHOULD first initiate the Address Validation procedure
 with PATH_CHALLENGE and PATH_RESPONSE frames as described in
 {{Section 8.2 of QUIC-TRANSPORT}}. It MAY skip the validation and
 simply send packets on the paths if it has previously validated
-that address.
+that address. 
 
 After receiving packets from the
 client on a new path, if the server decides to use the new path,
 the server SHOULD perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
 unless it has previously validated that address.
+Until the client's address is
+validated, the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
+applies.
 An endpoint that receives packets on a new path and does not want to establish
 this path is expected to close the path by sending a PATH_ABANDON
 on another path, as specified in section {{path-close}}.
@@ -1414,7 +1417,7 @@ such as e.g. for packet number handling, only after path validation has successf
 ## Request Forgery with Spoofed Address
 
 The path validation mechanism as specified in {{Section 8.2. of QUIC-TRANSPORT}} for migration is used
-unchanged for initiation of new paths in this extension. Respectively the security considerations
+unchanged for initiation of new paths in this extension. Therefore, the security considerations
 on source address spoofing as outlined in {{Section 21.5.4 of QUIC-TRANSPORT}} equally apply.
 Similarly, the anti-amplification limits as specified in {{Section 8 of QUIC-TRANSPORT}} need to be
 followed to limit the amplification risk.
@@ -1422,13 +1425,12 @@ followed to limit the amplification risk.
 However, while {{QUIC-TRANSPORT}} only allows the use of one path simultaneously
 and therefore only one path migration at the time should be validated,
 this extension allows for multiple open paths, that could in theory be migrated
-all at the same time, and it allows for multiple paths that could be initialised
+all at the same time, and it allows for multiple paths that could be initialized
 simultaneously. Therefore, each path could be used to further amplify an attack.
-Respectively endpoints needs limit the number of maximum paths and might consider
+Endpoints needs limit the number of maximum paths and might consider
 additional measures to limit the number of concurrent path validation processes
 e.g. by pacing them out or limiting the number of path initiation attempts
 over a certain time period.
-
 
 ## Use of Transport Layer Security and the AEAD Encryption Nonce
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -114,6 +114,9 @@ packet number space.
 To add a new path to an existing QUIC connection with multipath support,
 a client starts a path validation or sends a packet on
 the chosen path, as further described in {{path-initiation}}.
+A new path can only be used once the associated 4-tuple has been validated
+by ensuring that the peer is able to receive packets at that address
+(see {{Section 8 of QUIC-TRANSPORT}}).
 In this version of the document, a QUIC server does not initiate the creation
 of a path, but it can validate a new path created by a client.
 
@@ -378,19 +381,26 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same Path ID for sending the PATH_RESPONSE.
 
 When the multipath extension is negotiated, a client that wants to use an
-additional path SHOULD first initiate the Address Validation procedure
-with PATH_CHALLENGE and PATH_RESPONSE frames as described in
-{{Section 8.2 of QUIC-TRANSPORT}}. It MAY skip the validation and
-simply send packets on the paths if it has previously validated
-that address.
+additional path MUST validate the peer's address before sending any data packets 
+as described in ({{Section 8.2 of QUIC-TRANSPORT}}),
+unless it has previously validated the four-tuple used for that path.
 
 After receiving packets from the
 client on a new path, if the server decides to use the new path,
-the server SHOULD perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
-unless it has previously validated that address.
+the server MUST validate the peer's address before sending any data packets 
+as described in ({{Section 8.2 of QUIC-TRANSPORT}}),
+unless it has previously validated the four-tuple used for that path.
 Until the client's address is
 validated, the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
 applies.
+
+The server may receive packets for a yet unused Path ID that do not
+contain a path challenge. Such packets are valid if they can be properly decrypted
+and if they contain a valid connection ID. The server SHOULD accept the
+packets arriving on that path from the client, and acknowledge them.
+As stated above, the server will perform path validation before sending
+packets on this new path.
+
 An endpoint that receives packets on a new path and does not want to establish
 this path is expected to close the path by sending a PATH_ABANDON
 on another path, as specified in section {{path-close}}.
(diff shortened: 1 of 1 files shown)


mirjak:
Not sure this need an own section. We also need to change the following sentence in the initiation section:

> When the multipath extension is negotiated, a client that wants to
> use an additional path MUST first initiate the Address Validation procedure
> with PATH_CHALLENGE and PATH_RESPONSE frames as described in [Section 8.2](https://rfc-editor.org/rfc/rfc9000#section-8.2) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)],
>unless it has previously validated that address.

draft-ietf-quic-multipath.md
```diff
@@ -444,6 +444,18 @@ a server should issue a token that is capable of validating
 any of the previously validated addresses. Further guidance on token usage can be
 found in {{Section 8.1.3 of QUIC-TRANSPORT}}.
 
+### Immediate path initiation
+
+The server may receive packets for a yet unused path-ID that do not
```
mirjak, line 449:
```suggestion
The server may receive packets for a yet unused Path ID that do not
```

draft-ietf-quic-multipath.md
```diff
@@ -444,6 +444,18 @@ a server should issue a token that is capable of validating
...
+
+The server may receive packets for a yet unused Path ID that do not
+contain a path challenge. Such packets are valid if the can be properly decrypted
+and if they contain a valid connection ID. ({{path-initiation}}
+specifies that the client may send data packets immediately if "it has previously
+validated that address".) The server SHOULD create a new path, accept the
+packets arriving on that path from the client, and acknowledge them.
+
+If the server decides to not accept the packet, it MUST close the path as
+specified in {{path-close}}.
```
mirjak, line 457: I don't think this really makes sense or actually not even sure what you mean by accepting a packet. You receive a packet from a valid (not-yet used) Path ID; you decrypt it and as we say above this valid not matter what's in there. So either we don't need this sentence at all or maybe you want to say something like if the receiver does not want to open this path, it has to abandon it explicitly.

huitema: We have another comment about maybe not create a separate section. I can try moving the first paragraph in the main text.

Yanmei-Liu (approved):

huitema:
We may want to add content to address amplification limits. There is a special issue if the client starts a new path with a small data packet. In the absence of padding, the server will not have enough anti-amplification credit to send its own path challenge.

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +378,20 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same Path ID for sending the PATH_RESPONSE.
 
 When the multipath extension is negotiated, a client that wants to use an
-additional path MUST first initiate the Address Validation procedure
+additional path SHOULD first initiate the Address Validation procedure
 with PATH_CHALLENGE and PATH_RESPONSE frames as described in
-{{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
+{{Section 8.2 of QUIC-TRANSPORT}}. It MAY skip the validation and
+simply send packets on the paths if it has previously validated
 that address.
```
mirjak, line 385: We discussed in the meeting that we want a MUST here but don't say that path validation would require sending a path challenge but rather just refer to RFC9000 instead.

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +378,20 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
 it MUST pick a Connection ID with the same Path ID for sending the PATH_RESPONSE.
 
 When the multipath extension is negotiated, a client that wants to use an
-additional path MUST first initiate the Address Validation procedure
+additional path SHOULD first initiate the Address Validation procedure
 with PATH_CHALLENGE and PATH_RESPONSE frames as described in
```
mirjak, line 382:
```suggestion
additional path MUST validate the peer's address before sending any data packets 
as described in
```

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +378,20 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
...
-{{Section 8.2 of QUIC-TRANSPORT}}, unless it has previously validated
+{{Section 8.2 of QUIC-TRANSPORT}}. It MAY skip the validation and
+simply send packets on the paths if it has previously validated
 that address.
 
 After receiving packets from the
 client on a new path, if the server decides to use the new path,
-the server MUST perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
+the server SHOULD perform path validation ({{Section 8.2 of QUIC-TRANSPORT}})
```
mirjak, line 389: Similarly we should keep the MUST here or not use normative language because this is also covered by the MUST above.

draft-ietf-quic-multipath.md
```diff
@@ -112,11 +112,8 @@ non-zero connection IDs in order to identify the path and respective
...
 
 To add a new path to an existing QUIC connection with multipath support,
-a client starts a path validation on
+a client starts a path validation or sends a packet on
 the chosen path, as further described in {{path-initiation}}.
-A new path can only be used once the associated 4-tuple has been validated
-by ensuring that the peer is able to receive packets at that address
-(see {{Section 8 of QUIC-TRANSPORT}}).
 In this version of the document, a QUIC server does not initiate the creation
 of a path, but it can validate a new path created by a client.
```
mirjak, line 118: I think this is also confusing because there is effectively not really something that explicitly initiates the path.

draft-ietf-quic-multipath.md
```diff
@@ -112,11 +112,8 @@ non-zero connection IDs in order to identify the path and respective
 packet number space.
 
 To add a new path to an existing QUIC connection with multipath support,
-a client starts a path validation on
+a client starts a path validation or sends a packet on
 the chosen path, as further described in {{path-initiation}}.
-A new path can only be used once the associated 4-tuple has been validated
-by ensuring that the peer is able to receive packets at that address
-(see {{Section 8 of QUIC-TRANSPORT}}).
```
mirjak, line 119: Why did you remove this sentence? I think t's correct and simply part of the summary overview at the beginning. It is still true that a path/address as to be validated before using it. It's just the case that we a) can't enforce that validation by the other end and b) there can be other way to validate than sending a path challenge. However, this sentence doesn't say anything about these two points.

mirjak:
> We may want to add content to address amplification limits. There is a special issue if the client starts a new path with a small data packet. In the absence of padding, the server will not have enough anti-amplification credit to send its own path challenge.

I don't think this an issue that is specific for the multipath extension. If you can't validate the address because of the amplification limit, you simply can't sent on that path (or in case of RFC9000 you can't migrate to this path). In this case you can create the path and hope to receive more data such that you can sooner or later send the path challenge, or you might want to close the path by sending a path abandon frame on another path. If you don't want to close the path immediately but wait for more data, I guess it makes sense to start a timer. However, I don't think we need to specify anything here but can leave this to the implementation. Again, it's the same for RFC900 where you receive a packet that creates some migration state that you eventually want to remove at some point again.

draft-ietf-quic-multipath.md
huitema: I know we discussed that. But we also agreed that packets received from the peer should be processed and acknowledged. To acknowledge, the receiving peer has to create a "path context". We need to thread the needle through a fine hole...

The critical issue is that the peer address could be spoofed, as stated in section 9.2 of RFC 9000. Thus, address validation is required. It is not perfect.

I will try to rewrite that without any MUST/SHOULD/MAY, with simply a reference to RFC 9000.

draft-ietf-quic-multipath.md
huitema: I am concerned that receivers will treat the "PATH CHALLENGE" as a trigger to create a path context, and will drop packets that do not contain a path challenge. That can easily lead to interop issues.

draft-ietf-quic-multipath.md
huitema: There is: using a connection ID that reference a new path. That's unambiguous, and crypto protected.

draft-ietf-quic-multipath.md
huitema: I will try to rewrite.

huitema:
I will try to revive the previous note.

draft-ietf-quic-multipath.md
mirjak: Yes, address validation is required, however, if you have valid the address previously or in a different way, you don't need to the the QUIC path challenge/response mechanism. Therefore the agreement in the session was to have a MUST for address validation but not explicitly say that it has to be performed by using path challenge/response instead but just provide a point to RFC9000. See my proposed wording.

draft-ietf-quic-multipath.md
mirjak: The important part here is the "if a server decides to use the new path". However, we can also rephrase this and say more explicitly "before sending on a new path" or something.

But I actually thought the whole point of this PR was to explicitly clarify that even packets without a path challenge can create a new path as the other end is obligated to perform address validation but a) it can and shouldn't be enforced and b) it can be done by other means than sending a path challenge. Therefore to re-add this text to say valid packets have to be processed is probably good to ensure this is crystal clear.

draft-ietf-quic-multipath.md
huitema: @mirjak, I think the commit [Rewrite per review feedback](https://github.com/quicwg/multipath/pull/442/commits/9f74f86bb43f5509f4f49e60e61aac40a1f6679a) incorporates the text that we agreed upon. Please review!

draft-ietf-quic-multipath.md
huitema: Rewritten in
[Rewrite per review feedback.](https://github.com/quicwg/multipath/pull/442/commits/9f74f86bb43f5509f4f49e60e61aac40a1f6679a)

draft-ietf-quic-multipath.md
```diff
@@ -112,7 +112,7 @@ non-zero connection IDs in order to identify the path and respective
 packet number space.
 
 To add a new path to an existing QUIC connection with multipath support,
-a client starts a path validation on
+a client starts a path validation or sends a packet on
```
mirjak, line 115: Maybe be simply less concrete (as the next sentence already talks about address validation):
```suggestion
a client sends a packet with a connection ID belonging to a so-far unused Path ID on
```

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +381,27 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
...
+the server MUST validate the peer's address before sending any data packets
+as described in ({{Section 8.2 of QUIC-TRANSPORT}}),
+unless it has previously validated the four-tuple used for that path.
+Until the client's address is
+validated, the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
+applies.
+
+The server may receive packets for a yet unused Path ID that do not
+contain a path challenge. Such packets are valid if they can be properly decrypted
+and if they contain a valid connection ID. The server SHOULD accept the
```
mirjak, line 399: Not sure we need normative language here, or this sentence at all?

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +381,27 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
...
+packets arriving on that path from the client, and acknowledge them.
+As stated above, the server will perform path validation before sending
+packets on this new path.
```
mirjak, line 402:
```suggestion
As stated above, the server independently need to perform path validation before sending
packets on this new path.
```

draft-ietf-quic-multipath.md
huitema: Less is more. I will just remove the sentence.

mirjak:
Thanks for the update. The proposed change look good to me now.

Looking at PR #430, do we want to note anything about padding or not?

huitema:
I have been thinking about that. It is kinda implicit from the requirement that the server follows anti amplification rule. Maybe we should add some text.

huitema:
I added a short sentence about amplification limits, but I did not write normative text about padding at the client. I am concerned that doing so would open another can of worms, e.g., should the server reject a packet on a new path if that packet was not padded? Just mentioning amplification is probably sufficient: the implementers should know that not padding will prevent the server from sending data. Plus, the text in RFC 9000 already says:

_When an endpoint is unable to expand the datagram size to 1200 bytes due to the anti-amplification limit, the path MTU will not be validated. To ensure that the path MTU is large enough, the endpoint MUST perform a second path validation by sending a PATH_CHALLENGE frame in a datagram of at least 1200 bytes. This additional validation can be performed after a PATH_RESPONSE is successfully received or when enough bytes have been received on the path that sending the larger datagram will not result in exceeding the anti-amplification limit._

So I think that we are covered.

yfmascgy (approved):
LGTM

qdeconinck (approved):

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +381,27 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
...
+Until the client's address is
+validated, the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
+applies.
+
+The server may receive packets for a yet unused Path ID that do not
+contain a path challenge. Such packets are valid if they can be properly decrypted
+and if they contain a valid connection ID.
+As stated above, the server independently needs to perform path validation before sending
+data packets on this new path, and the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
+applies.
```
mirjak, line 402: Now we are stating this two points twice in two subsequence paragraphs...? I don't think this sentence is needed.

mirjak:
```suggestion
```

mirjak:
> I added a short sentence about amplification limits, but I did not write normative text about padding at the client. I am concerned that doing so would open another can of worms, e.g., should the server reject a packet on a new path if that packet was not padded? Just mentioning amplification is probably sufficient: the implementers should know that not padding will prevent the server from sending data. Plus, the text in RFC 9000 already says:
> 
> _When an endpoint is unable to expand the datagram size to 1200 bytes due to the anti-amplification limit, the path MTU will not be validated. To ensure that the path MTU is large enough, the endpoint MUST perform a second path validation by sending a PATH_CHALLENGE frame in a datagram of at least 1200 bytes. This additional validation can be performed after a PATH_RESPONSE is successfully received or when enough bytes have been received on the path that sending the larger datagram will not result in exceeding the anti-amplification limit._
> 
> So I think that we are covered.

Maybe similarly as we explicitly say that each IP address must be validated, we might also want to say that it must be validated for each path (and each direction) that a MTU of at least 1200 bytes is supported? I think this is really the same as the requirement on IP address validation, where each endpoint need to do it independently but can and should enforce it for the other end. So something like this:

`Each endpoint MUST also validate that a minimum MTU of 1200 bytes is supported on the path. This can be done during initial path validation or separately later if the amplification limit prevents it initially, as specified in Section 8.2.1 of RFC9000.`

draft-ietf-quic-multipath.md
mirjak, line 402:
```suggestion
applies.

Each endpoint MUST also validate that a minimum MTU of 1200 bytes is supported on the path. This can be done during initial path validation or separately later if the amplification limit prevents it initially, as specified in Section 8.2.1 of RFC9000.
```

draft-ietf-quic-multipath.md
```diff
@@ -381,16 +381,26 @@ with the same, unused Path ID. When the peer receives the PATH_CHALLENGE,
...
+unless it has previously validated the four-tuple used for that path.
+Until the client's address is
+validated, the anti-amplification limit from {{Section 8 of QUIC-TRANSPORT}}
+applies.
+
+The server may receive packets for a yet unused Path ID that do not
+contain a path challenge. Such packets are valid if they can be properly decrypted
+and if they contain a valid connection ID.
+
+Each endpoint MUST also validate that a minimum MTU of 1200 bytes is supported on the path. This can be done during initial path validation or separately later if the amplification limit prevents it initially, as specified in Section 8.2.1 of RFC9000.
```
mirjak, line 401:
```suggestion
Each endpoint MUST also validate that a minimum MTU of 1200 bytes is supported
on the path. This can be done during initial path validation or separately later if
the amplification limit prevents it initially, as specified in Section 8.2.1 of RFC9000.
```

draft-ietf-quic-multipath.md
```diff
@@ -1416,22 +1428,21 @@ such as e.g. for packet number handling, only after path validation has successf
...
 and therefore only one path migration at the time should be validated,
 this extension allows for multiple open paths, that could in theory be migrated
-all at the same time, and it allows for multiple paths that could be initialised
+all at the same time, and it allows for multiple paths that could be initialized
 simultaneously. Therefore, each path could be used to further amplify an attack.
-Respectively endpoints needs limit the number of maximum paths and might consider
+Endpoints needs limit the number of maximum paths and might consider
 additional measures to limit the number of concurrent path validation processes
 e.g. by pacing them out or limiting the number of path initiation attempts
 over a certain time period.
```
mirjak, line 1445:
```suggestion


```

PR #444, opened 2024-10-02
Add more guidance on when to use PATH_ACK frames

### LINKED ISSUE #220: "SHOULD" use ACK_MP frames?

Yanmei-Liu:
To close #220

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -318,11 +318,10 @@ Path ID 0.
 
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
-also for acknowledging so far unacknowledged 0-RTT packets, using
-Path ID 0. Further more, endpoints SHOULD use PATH_ACK frames in all 1-RTT packets after
-the successful negotiation for multipath. If endpoints receive ACK frame
-after a succesful negotiation resulting from packet out-of-order arrival,
-endpoints could consider it as used for initial PATH ID 0 without causing connection error.
+including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0. 
+However, even after successful negotiation, endpoints MAY use ACK frames 
+to acknowledge 0-RTT packets or 1-RTT packets received with Path ID 0, 
+the peers process these ACK frames as for Path ID 0 without causing error.
 Similarly after a successful handshake, endpoints SHOULD also use
 the PATH_NEW_CONNECTION_ID frame to provide new connection IDs for Path ID 0 and,
 respectively, the PATH_RETIRE_CONNECTION_ID frame to retire connection IDs for
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -318,9 +318,9 @@ Path ID 0.
 
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
-including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0. 
-However, even after successful negotiation, endpoints MAY use ACK frames 
-to acknowledge 0-RTT packets or 1-RTT packets received with Path ID 0, 
+including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0.
+However, even after successful negotiation, endpoints MAY use ACK frames
+to acknowledge 0-RTT packets or 1-RTT packets received with Path ID 0,
 the peers process these ACK frames as for Path ID 0 without causing error.
 Similarly after a successful handshake, endpoints SHOULD also use
 the PATH_NEW_CONNECTION_ID frame to provide new connection IDs for Path ID 0 and,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -321,7 +321,7 @@ endpoints SHOULD use PATH_ACK frames instead of ACK frames,
 including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0.
 However, even after successful negotiation, endpoints MAY use ACK frames
 to acknowledge 0-RTT packets or 1-RTT packets received on initial path with Path ID 0,
-these ACK frames need to be processed correctly by peer.
+and these ACK frames need to be processed correctly by the peer.
 Similarly after a successful handshake, endpoints SHOULD also use
 the PATH_NEW_CONNECTION_ID frame to provide new connection IDs for Path ID 0 and,
 respectively, the PATH_RETIRE_CONNECTION_ID frame to retire connection IDs for
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -320,9 +320,9 @@ After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
 also for acknowledging so far unacknowledged 0-RTT packets, using
 Path ID 0. Further more, endpoints SHOULD use PATH_ACK frames in all 1-RTT packets after
-the successful negotiation for multipath. If endpoints receive ACK frame 
-after a succesful negotiation resulting from packet out-of-order arrival, 
-endpoints could consider it as used for initial PATH ID 0 without causing connection error. 
+the successful negotiation for multipath. If endpoints receive ACK frame
+after a succesful negotiation resulting from packet out-of-order arrival,
+endpoints could consider it as used for initial PATH ID 0 without causing connection error.
 Similarly after a successful handshake, endpoints SHOULD also use
 the PATH_NEW_CONNECTION_ID frame to provide new connection IDs for Path ID 0 and,
 respectively, the PATH_RETIRE_CONNECTION_ID frame to retire connection IDs for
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -319,7 +319,11 @@ Path ID 0.
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
 also for acknowledging so far unacknowledged 0-RTT packets, using
-Path ID 0. Similarly after a successful handshake, endpoints SHOULD also use
+Path ID 0. Further more, endpoints SHOULD use PATH_ACK frames in all 1-RTT packets after
+the successful negotiation for multipath. If endpoints receive ACK frame
```
mirjak, line 323: This additional sentence is not needed at this is already what the sentence above says. All packets after the handshake has concluded are 1-RTT packets.

draft-ietf-quic-multipath.md
```diff
@@ -319,7 +319,11 @@ Path ID 0.
...
+after a succesful negotiation resulting from packet out-of-order arrival,
```
mirjak, line 324: Given this is a SHOULD only, it is not only out-of-order arrivals. The peer can decide to keep using ACK frames for all Path ID 0 packets. However, we should explain when this might be useful, e.g. in order to same some bytes, especially when only one path is used.

mirjak (changes requested):
The new SHOULD is not needed as this is covered already by the existing first sentence in this paragraph.

draft-ietf-quic-multipath.md
```diff
@@ -319,7 +319,11 @@ Path ID 0.
...
+endpoints could consider it as used for initial PATH ID 0 without causing connection error.
+Similarly after a successful handshake, endpoints SHOULD also use
 the PATH_NEW_CONNECTION_ID frame to provide new connection IDs for Path ID 0 and,
 respectively, the PATH_RETIRE_CONNECTION_ID frame to retire connection IDs for
 Path ID 0.
```
huitema, line 329: I think I agree that the added text "Further more, endpoints SHOULD use PATH_ACK frames in all 1-RTT packets after
the successful negotiation for multipath. If endpoints receive ACK frame after a succesful negotiation resulting from packet out-of-order arrival, endpoints could consider it as used for initial PATH ID 0 without causing connection error." is not what we want. I would replace it by something like:

"However, even after successful negotiation, endpoints MAY use ACK frames  to acknowledge 0RTT packets or 1RTT packets received with PathID=0."

Stating the MAY implicitly states that arrival of such packets should not be treated as a connection error.

mirjak: I'm fine to add this sentence to make it explicit that those packets arrive. However, I'm not sure about the use of normative language here. Because this is not to say, you are aloof to send that (that's already included in the SHOULD above). It's to say, please be aware that they may arrive and you have to handle them. Maybe like this:

> Still, even after successful negotiation, any ACK frames that acknowledge 0-RTT packets or 1-RTT packets for path ID 0 need to be processed correctly.

draft-ietf-quic-multipath.md
```diff
@@ -318,8 +318,11 @@ Path ID 0.
 
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
-also for acknowledging so far unacknowledged 0-RTT packets, using
-Path ID 0. Similarly after a successful handshake, endpoints SHOULD also use
+including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0.
+However, even after successful negotiation, endpoints MAY use ACK frames
+to acknowledge 0-RTT packets or 1-RTT packets received on initial path with Path ID 0,
+these ACK frames need to be processed correctly by peer.
```
mirjak, line 324:
```suggestion
However, even after successful negotiation, endpoints can use ACK frames
to acknowledge 0-RTT packets or 1-RTT packets received on initial path with Path ID 0,
and these ACK frames need to be processed correctly by the peer.
```

huitema: I think that we should keep the normative MAY. You want to say something like "MAY but we don't like it", but there is no way to say that in a standard. Either it is a protocol violation or it isn't. If it isn't, then the peer MAY do it.

mirjak: This MAY is not needed because we already have the SHOULD in the sentence before. I don't think we should say the same things twice normatively.

huitema: That's one opinion. I have witnessed several IESG reviews asking "why do you say SHOULD and not MUST? If you say SHOULD, you must also have a clause explaining when the implementation MAY do something else." 

That construct is used at least once in RFC 9000, in section 11:

_The most appropriate error code ([Section 20](https://www.rfc-editor.org/rfc/rfc9000.html#error-codes)) SHOULD be included in the frame that signals the error. Where this specification identifies error conditions, it also identifies the error code that is used; though these are worded as requirements, different implementation strategies might lead to different errors being reported. In particular, an endpoint MAY use any applicable error code when it detects an error condition; a generic error code (such as PROTOCOL_VIOLATION or INTERNAL_ERROR) can always be used in place of specific error codes._

On the other hand, I don't care all that much...

qdeconinck, line 324: Having read the discussion, I am fine keeping the MAY. An alternative could also be to focus on receiving ACKs (indicate that they must be correctly processed) without explicitly mentioning that the endpoint can send it (I agree this is somewhat mentioned by the SHOULD).

Minor typos.
```suggestion
and these ACK frames need to be processed correctly by the peer.
```

mirjak: The point of adding this sentence is to make the receiver side aware and not to specify any requirements for the sender. There the sentence is rather stating a fact (like "note, this can happen and is valid") that setting a normative protocol requirement. If we want to phrase a requirement here, it should be on the receiver (as Quentin also says below). So basically saying something like:

`Endpoints MUST process ACK frames that acknowledge 0-RTT packets or 1-RTT packets for Path ID 0, even after successful negotiation of this extension.`

Further the intention of adding something here as discussed in the issue was to give a bit of a reasoning why there is a SHOULD in the above sentence and why it could be useful to still send ACK frames. So I propose to further add a sentence like this:

`For example, ACK frames might be preferred by the sender as long as only the initial path with Path ID 0 is in use.`

mirjak:
```suggestion
Endpoints MUST process ACK frames that acknowledge 0-RTT packets or 1-RTT packets
for Path ID 0, even after successful negotiation of this extension. For example,
ACK frames might be preferred by the sender as long as only the initial path
with Path ID 0 is in use.
```

mirjak, line 324:
```suggestion
and these ACK frames need to be processed correctly by the peer.
```

draft-ietf-quic-multipath.md
```diff
@@ -318,8 +318,11 @@ Path ID 0.
 
 After the handshake concluded if negotiation of multipath support succeeded,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
-also for acknowledging so far unacknowledged 0-RTT packets, using
-Path ID 0. Similarly after a successful handshake, endpoints SHOULD also use
+including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0.
+However, even after successful negotiation, endpoints MAY use ACK frames
+to acknowledge 0-RTT packets or 1-RTT packets received on initial path with Path ID 0,
+and these ACK frames need to be processed correctly by the peer.
```
mirjak, line 324: Sorry, I merged the nits and that made the conversation disappear here, so I'm repost my last comment and change proposal:

The point of adding this sentence is to make the receiver side aware and not to specify any requirements for the sender. There the sentence is rather stating a fact (like "note, this can happen and is valid") that setting a normative protocol requirement. If we want to phrase a requirement here, it should be on the receiver (as Quentin also says below). So basically saying something like:

Endpoints MUST process ACK frames that acknowledge 0-RTT packets or 1-RTT packets for Path ID 0, even after successful negotiation of this extension.

Further the intention of adding something here as discussed in the issue was to give a bit of a reasoning why there is a SHOULD in the above sentence and why it could be useful to still send ACK frames. So I propose to further add a sentence like this:

For example, ACK frames might be preferred by the sender as long as only the initial path with Path ID 0 is in use.

mirjak:
```suggestion
Endpoints MUST process ACK frames that acknowledge 0-RTT packets or 1-RTT packets
for Path ID 0, even after successful negotiation of this extension. For example, ACK frames
might be preferred by the sender as long as only the initial path with Path ID 0 is in use.
```

huitema: +1. This is a good change.

qdeconinck: Agree too and committed suggestion.

qdeconinck (approved):
With the new text, looks good to me.

Issue #414, opened 2024-07-10
Abandon Frame needs Error Code definition


gloinul:
Section 7.2:
Should Error Code reference the number space defined by transport or the IANA Registry?

Yanmei-Liu:
Actually I find that the current draft lack of the definition for path abandon close error code. The [error code](https://datatracker.ietf.org/doc/html/draft-ietf-quic-multipath-10#name-error-code-for-multipath-qu) registered in IANA Registry is used for Connection Close Frame. We definitely need to add this part.

huitema:
Do we really want to replicate the structure of connection close, with 2 frame types, one "abandon by application" and the other "abandon by transport"? Or do we want to follow the example of "Stream Reset" and "Stop Sending", which just carry an application defined code?

I am for "application defined", because I think there are only two cases in which the transport originates an Abandon packet "by itself": abandon on timer, and abandon in response to an abandon by the peer. Using error code=0 for these two cases seems like a fine simplification.

mirjak:
This was discussion in the session at IETF-120. Decision was to remove the Reason Phrase and define a MP specific error code space, e.g. including error codes like "closed by application", "out of ressources", ...

yfmascgy:
I am trying to address this issue with PR #445

mirjak:
Thanks for starting the PR @yfmascgy. In the discussion @marten-seemann proposed to define a separate error code space rather than using the existing QUIC error code space. Do people have opinions on that?

mirjak:
Also more more question: Should the PR also add information which error code to use in each case where we require/recommend to send a path abandon frame. E.g. if close a path because of a local timeout, you should probably use "NO_ERROR". If you close a path that the other end tries to open, you maybe use "NO_ERROR" or maybe also something more specific like "RESOURCE_LIMIT_REACHED" which indicates to the other end that it should not try to open this path (or any other path) right now again...?

huitema:
I think we should keep using the QUIC code space. That was pretty much the recommendation during the WG meeting.

mirjak:
Reposting here for the discussion what I just posted in the PR:  I think using the QUIC error code space is wrong. I think we should not call this error code at all. In case of connection close, you actually close the connection because there was an unexpected protocol behavior. In case of the multipath extension, it is the same: if there is something unexpected, you have to close the whole connection. Therefore closing a path is never an error but always an explicit decision.

mirjak:
[...]

PR #445, opened 2024-10-03
Dev/add error code

yfmascgy:
First attempt on #414 

Here is an initial attempt to remove the reason phrase from the PATH_ABANDON frame and add two error codes: APPLICATION_ABANDON and RESOURCE_LIMIT_REACHED. This follows the discussion in issue #414 and references the discussion notes from IETF-120. I have no strong opinion on the names of these two error codes and have set their values as TBD for both.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1338,9 +1338,22 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
-# Error Codes
+# Error Codes {#error-codes}
+
+Multipath QUIC transport error codes are 62-bit unsigned integers
+following {{QUIC-TRANSPORT}}. 
 
 The following error codes are defined for use in the PATH_ABANDON frame.
+
+APPLICATION_ABANDON (TBD): The endpoint is abandoning the path at the 
+request of the application. The application has determined that it no 
+longer needs this path. This error is used when the application layer 
+decides to stop using a specific path.
+
+RESOURCE_LIMIT_REACHED (TBD): The endpoint is abandoning the path because 
+it cannot allocate sufficient resources to maintain it. This is due to 
+limitations in the transport layer's capacity. This error indicates that 
+resource constraints prevent the continuation of the path.
 
 # IANA Considerations
 
@@ -1373,6 +1386,16 @@ TBD-06 (experiments use 0x15228c0a)                  | PATH_RETIRE_CONNECTION_ID
 TBD-07 (experiments use 0x15228c0c)                  | MAX_PATH_ID            | {{max-paths-frame}}
 TBD-08 (experiments use 0x15228c0d)                  | PATHS_BLOCKED    | {{paths-blocked-frame}}
 {: #frame-types title="Addition to QUIC Frame Types Entries"}
+
+The following transport error code defined in {{tab-error-code}} are to
+be added to the "QUIC Transport Error Codes" registry under
+the "QUIC Protocol" heading.
+
+Value                       | Code                  | Description                   | Specification
+----------------------------|-----------------------|-------------------------------|-------------------
+TBD | APPLICATION_ABANDON | Path abandoned at the application's request | {{error-codes}}
+TBD | RESOURCE_LIMIT_REACHED | Path abandoned due to resource limitations in the transport | {{error-codes}}
+{: #tab-error-code title="Error Codes for Multipath QUIC"}
 
 
 # Security Considerations
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1341,9 +1341,9 @@ Maximum Path Identifier:
 # Error Codes {#error-codes}
 
 QUIC transport error codes are 62-bit unsigned integers
-following {{QUIC-TRANSPORT}}.
-
-In addition to NO_ERROR(0x0) following {{QUIC-TRANSPORT}}, the following QUIC error codes are defined for use in the PATH_ABANDON frame.
+(see {{Section 20.1 of QUIC-TRANSPORT}}. In addition to
+NO_ERROR(0x0), the following QUIC error codes are defined
+for use in the PATH_ABANDON frame:
 
 APPLICATION_ABANDON (TBD-09): The endpoint is abandoning the path at the
 request of the application. The application has determined that it no
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1340,7 +1340,7 @@ Maximum Path Identifier:
 
 # Error Codes {#error-codes}
 
-Multipath QUIC transport error codes are 62-bit unsigned integers
+QUIC transport error codes are 62-bit unsigned integers
 following {{QUIC-TRANSPORT}}.
 
 The following additional QUIC error codes are defined for use in the PATH_ABANDON frame.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1341,18 +1341,18 @@ Maximum Path Identifier:
 # Error Codes {#error-codes}
 
 Multipath QUIC transport error codes are 62-bit unsigned integers
-following {{QUIC-TRANSPORT}}. 
+following {{QUIC-TRANSPORT}}.
 
 The following error codes are defined for use in the PATH_ABANDON frame.
 
-APPLICATION_ABANDON (TBD): The endpoint is abandoning the path at the 
-request of the application. The application has determined that it no 
-longer needs this path. This error is used when the application layer 
+APPLICATION_ABANDON (TBD): The endpoint is abandoning the path at the
+request of the application. The application has determined that it no
+longer needs this path. This error is used when the application layer
 decides to stop using a specific path.
 
-RESOURCE_LIMIT_REACHED (TBD): The endpoint is abandoning the path because 
-it cannot allocate sufficient resources to maintain it. This is due to 
-limitations in the transport layer's capacity. This error indicates that 
+RESOURCE_LIMIT_REACHED (TBD): The endpoint is abandoning the path because
+it cannot allocate sufficient resources to maintain it. This is due to
+limitations in the transport layer's capacity. This error indicates that
 resource constraints prevent the continuation of the path.
 
 # IANA Considerations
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1353,6 +1338,23 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+# Error Codes {#error-codes}
+
+Multipath QUIC transport error codes are 62-bit unsigned integers
+following {{QUIC-TRANSPORT}}.
+
+The following error codes are defined for use in the PATH_ABANDON frame.
```
mirjak, line 1346:
```suggestion
The following additional QUIC error codes are defined for use in the PATH_ABANDON frame.
```
I assume the intention is to be able also use e.g. NO_ERROR, right? Maybe it's good to say that explicitly?

draft-ietf-quic-multipath.md
```diff
@@ -1353,6 +1338,23 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+# Error Codes {#error-codes}
+
+Multipath QUIC transport error codes are 62-bit unsigned integers
+following {{QUIC-TRANSPORT}}.
+
+The following additional QUIC error codes are defined for use in the PATH_ABANDON frame.
```
yfmascgy, line 1346:
```suggestion
In addition to NO_ERROR(0x0) following {{QUIC-TRANSPORT}}, the following QUIC error codes are defined for use in the PATH_ABANDON frame.
```

yfmascgy (commented):
NO_ERROR (0x0) is described in the text above; explicitly mentioning NO_ERROR (0x0) again here helps to clarify.

draft-ietf-quic-multipath.md
qdeconinck: While I agree that NO_ERROR is a fine error message for a PATH_ABANDON, the current text is not really clear about the other error codes defined in RFC9000. For instance, what should a receiver do if it receives a PATH_ABANDON with error code FLOW_CONTROL_ERROR? This would sound wrong to me, though it is not clear how a receiver would react.

mirjak: I don't think the receiver would react it all. This is only for logging. If there are concerns about this that I think we would need to define a separate code space. Thinking more about this, I think using the QUIC error code space is wrong. I think we should not call this error code at all. In case of connection close, you actually close the connection because there was an unexpected protocol behavior. In case of the multipath extension, it is the same: if there is something unexpected, you have to close the whole connection. Therefore closing a path is never an error but always an explicit decision.

qdeconinck: Yes, I agree with your point. If we need such logging, we may standardise a few PATH_ABANDON-specific reason code points (the three mentioned above seem sufficient).

LPardue: As a reference, [HTTP/3's error code space](https://www.iana.org/assignments/http3-parameters/http3-parameters.xhtml#http3-parameters-error-codes) contains codes that cover various scenarios.

This include unexpected protocol errors, such as detecting the peer closed a stream it shouldn't and responding with a CONNECTION_CLOSE of H3_CLOSED_CRITICAL_STREAM.

But it also includes expected behaviours that are not errors, such as when an endpoint cancels a request (a bit like path abandon) by sending a REST_STREAM or STOP_SENDING using H3_REQUEST_CANCELLED.

In all these cases, an H3 implementation has space to ignore the specific language in the spec: use the wrong error code space (send a transport error when it was an application error), or send a generic code, or send an incorrect code. I don't think any of those cause any real problems for the recipient, as Mirja put it, it's mainly for debugging. And someone investigating these should always be skeptical of the peer not telling the ground truth.

On balance, I think its fine to use the QUIC error code space here. This is an example of an extension where we are building some more complexity into the transport and can expect to need to extend the shared codepoint space to suit.

huitema: I also think that using the QUIC error space is fine. It is fairly obvious that not all error codes defined in the QUIC error space apply to all scenarios. For example, a "transport parameter error" only makes sense when parsing transport parameters. The important property is that the error codes be unique.

I also do not think that any implementation should react to specific error code values, unless that's explicitly mentioned in the text of the standard. The purpose is documentation for helping debugging issues, or for passing information to the application and ultimately the user. We don't need a special registry for path abandon reasons.

draft-ietf-quic-multipath.md
```diff
@@ -1353,6 +1338,23 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+# Error Codes {#error-codes}
+
+Multipath QUIC transport error codes are 62-bit unsigned integers
```
yfmascgy, line 1343:
```suggestion
QUIC transport error codes are 62-bit unsigned integers
```

yfmascgy (commented):
Remove the word "multipath" as we are using the QUIC error space.

draft-ietf-quic-multipath.md
```diff
@@ -1343,7 +1343,7 @@ Maximum Path Identifier:
 QUIC transport error codes are 62-bit unsigned integers
 following {{QUIC-TRANSPORT}}.
 
-The following additional QUIC error codes are defined for use in the PATH_ABANDON frame.
+In addition to NO_ERROR(0x0) following {{QUIC-TRANSPORT}}, the following QUIC error codes are defined for use in the PATH_ABANDON frame.
```
huitema, line 1347: Should we add provisional values for the two error codes in the IANA section, to enable interop testing?

huitema (approved):
I think this is good. We could discuss whether "application abandon" is really different from "no error", but error codes are cheap and I would rather see this merged. I just have a small nit.

draft-ietf-quic-multipath.md
Yanmei-Liu: I support @huitema 's idea, we need two experimental error codes value here for interop testing.

draft-ietf-quic-multipath.md
```diff
@@ -1385,6 +1387,16 @@ TBD-07 (experiments use 0x15228c0c)                  | MAX_PATH_ID            |
 TBD-08 (experiments use 0x15228c0d)                  | PATHS_BLOCKED    | {{paths-blocked-frame}}
 {: #frame-types title="Addition to QUIC Frame Types Entries"}
 
+The following transport error code defined in {{tab-error-code}} are to
+be added to the "QUIC Transport Error Codes" registry under
+the "QUIC Protocol" heading.
+
+Value                       | Code                  | Description                   | Specification
+----------------------------|-----------------------|-------------------------------|-------------------
+TBD | APPLICATION_ABANDON | Path abandoned at the application's request | {{error-codes}}
```
Yanmei-Liu, line 1396:
```suggestion
TBD-09(experiments use 0x4150504C4142414E) | APPLICATION_ABANDON | Path abandoned at the application's request | {{error-codes}}
```

draft-ietf-quic-multipath.md
```diff
@@ -1385,6 +1387,16 @@ TBD-07 (experiments use 0x15228c0c)                  | MAX_PATH_ID            |
...
+TBD | RESOURCE_LIMIT_REACHED | Path abandoned due to resource limitations in the transport | {{error-codes}}
```
Yanmei-Liu, line 1397:
```suggestion
TBD-10(experiments use 0x5245534C494D4954) | RESOURCE_LIMIT_REACHED | Path abandoned due to resource limitations in the transport | {{error-codes}}
```

draft-ietf-quic-multipath.md
```diff
@@ -1353,6 +1338,23 @@ Maximum Path Identifier:
...
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+# Error Codes {#error-codes}
+
+QUIC transport error codes are 62-bit unsigned integers
+following {{QUIC-TRANSPORT}}.
+
+In addition to NO_ERROR(0x0) following {{QUIC-TRANSPORT}}, the following QUIC error codes are defined for use in the PATH_ABANDON frame.
+
+APPLICATION_ABANDON (TBD): The endpoint is abandoning the path at the
```
Yanmei-Liu, line 1348:
```suggestion
APPLICATION_ABANDON (TBD-09): The endpoint is abandoning the path at the
```

draft-ietf-quic-multipath.md
```diff
@@ -1353,6 +1338,23 @@ Maximum Path Identifier:
...
+request of the application. The application has determined that it no
+longer needs this path. This error is used when the application layer
+decides to stop using a specific path.
+
+RESOURCE_LIMIT_REACHED (TBD): The endpoint is abandoning the path because
```
Yanmei-Liu, line 1353:
```suggestion
RESOURCE_LIMIT_REACHED (TBD-10): The endpoint is abandoning the path because
```

Yanmei-Liu (approved):
Support the idea. 
I add suggestion about using 2 experimental value for interop tests, although we'd replace them with formal value later.

draft-ietf-quic-multipath.md
```diff
@@ -1353,6 +1338,23 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+# Error Codes {#error-codes}
+
+QUIC transport error codes are 62-bit unsigned integers
+following {{QUIC-TRANSPORT}}.
+
+In addition to NO_ERROR(0x0) following {{QUIC-TRANSPORT}}, the following QUIC error codes are defined for use in the PATH_ABANDON frame.
```
mirjak, line 1346:
```suggestion
In addition to NO_ERROR(0x0) (see {{Section 21.1 of QUIC-TRANSPORT}}, the following QUIC error codes are defined for use in the PATH_ABANDON frame.
```

mirjak, line 1346:
```suggestion
QUIC transport error codes are 62-bit unsigned integers (see {{Section 20.1 of QUIC-TRANSPORT}}.
In addition to NO_ERROR(0x0), the following QUIC error codes are defined
for use in the PATH_ABANDON frame:
```

qdeconinck (approved):

PR #402, opened 2024-07-02
Add MAX_PATH_ID_BLOCKED frame and related texts

yfmascgy:
This is an attempt to fix issue #342. I have added the MAX_PATH_ID_BLOCKED frame and related texts. Note that the same Path ID is used in both directions to address a path so the path identifier can be either limited by the sender or its peer. We should treat the two scenario differently and MAX_PATH_ID_BLOCKED should be used when the limit is set by the peer.

PR #448, opened 2024-10-21
Increasing the transport parameter value

mirjak:
At least #428 (remove MP error) and #402 (PATH_BLOCKED) are breaking changes and #445 will be as well.


Yanmei-Liu (approved):

Yanmei-Liu:
We also have an transport parameter reference in Section {#nego}, you can replace the old value "0x0f739bbc1b666d09" with the new one and then find it.

Yanmei-Liu (approved):
Looks good!

PR #456, opened 2024-10-31
PATH_ACK can still be sent after PATH_ABANDON

mirjak:


huitema (changes requested):
This is the wrong fix. The root cause is that the text in the last paragraph of "PATH_ABANDON Frame" section contradicts the specification in "Path Close". It should be replaced by a simple reference, e.g. "The use of the PATH_ABANDON frame to close paths is specified in section {{path-close}}."

mirjak:
@huitema agreed. Done!

Yanmei-Liu (approved):

qdeconinck (approved):

gloinul (approved):

Yanmei-Liu (approved):

Issue #461, opened 2024-11-03
PATH BLOCKED should be clearly labelled "informational"


huitema:
The interop tests showed mismatched expectations, with some stacks apparently expecting to promptly receive a new MAX_PATH_ID frame after sending a PATH_BLOCKED frame. Of course, that's not happening. If the number of allocated paths has reached the limit supported by the server, the MAX_PATH_ID value can only increase if some allocated paths are abandoned. We should maybe clarify that PATH_BLOCKED is purely informational, and that it is not expected to trigger changes in the state of the peer.

PR #463, opened 2024-11-04
Add clarification for PATHS_BLOCKED frame.

Yanmei-Liu:
Fix issue #461. 
Generally the PATH_BLOCKED frame is informational and doesn't change the state of the peer.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1346,8 +1346,11 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
-limit set by its peer. Note that PATH_BLOCKED frame is informational, as it isn’t
-expected to change the state of peer.
+limit set by its peer.
+Note that PATH_BLOCKED frame is informational, as it isn’t expected to change
+the state of peer. Sending a PATH_BLOCKED frame does not imply a particular action
+from the peer like updating the new Max Path ID value, but informs the peer that
+the maximum path identifier limit prevented the creation of new paths.
 
 PATHS_BLOCKED frames are formatted as shown in {{fig-paths-blocked-frame-format}}.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1347,7 +1347,7 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
 limit set by its peer.
-Note that PATH_BLOCKED frame is informational. Sending a PATH_BLOCKED frame does not
+Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
 imply a particular action from the peer like updating the new Max Path ID value,
 but informs the peer that the maximum path identifier limit prevented the creation of new paths.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1315,6 +1315,8 @@ processing of a RETIRE_CONNECTION_ID frame is only applied for Path ID 0.
 
 A MAX_PATH_ID frame (type=0x15228c0c) informs the peer of the maximum path identifier
 it is permitted to use.
+If the number of allocated paths has reached the local limit of max path identifier supported
+by the endpoint, it SHOULD send MAX_PATH_ID frame to increase the local limit and inform the peer.
```
mirjak, line 1319: I don`t think we should make a normative requirement here. it's really the receiver decision what to do here.

huitema: I agree with @mirjak. We cannot have an unqualified SHOULD here. If the peer has reached its internal active path limit, it is not going to increase MAX_PATH_ID. The client endpoint has a budget. If the peer only supports N path, it is not going to increase that to N+1. If the client wants to open a new path, it has to abandon an old one.

Yanmei-Liu: Okay. I removed normative language here.

huitema (approved):

draft-ietf-quic-multipath.md
```diff
@@ -1346,7 +1346,8 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
-limit set by its peer;
+limit set by its peer. Note that PATH_BLOCKED frame is informational, as it isn’t
+expected to change the state of peer.
```
mirjak, line 1350: I think we should be more explicit what we mean here, maybe like:

"Sending a PATH_BLOCKED frame does not imply a particular action from the peer like updating the new Max Path ID value."

draft-ietf-quic-multipath.md
```diff
@@ -1346,7 +1346,11 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
-limit set by its peer;
+limit set by its peer.
+Note that PATH_BLOCKED frame is informational, as it isn’t expected to change
+the state of peer. Sending a PATH_BLOCKED frame does not imply a particular action
```
qdeconinck, line 1351:
```suggestion
Note that PATH_BLOCKED frame is informational, as it is not expected to change
the state of the peer. Sending a PATH_BLOCKED frame does not imply a particular action
```

qdeconinck (approved):

mirjak (approved):

JonathanLennox:
Is it PATHS_BLOCKED or PATH_BLOCKED?

draft-ietf-quic-multipath.md
```diff
@@ -1346,7 +1346,11 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
-limit set by its peer;
+limit set by its peer.
+Note that PATH_BLOCKED frame is informational. Sending a PATH_BLOCKED frame does not
```
JonathanLennox, line 1350: These should both say PATHS_BLOCKED rather than PATH_BLOCKED

Yanmei-Liu, line 1350:
```suggestion
Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
```

Yanmei-Liu: Nice catch! Updated.

gloinul (approved):

PR #466, opened 2024-11-11
Add two more specific error codes for path abandon frame

Yanmei-Liu:
To fix issue #414, according to the discussion during IETF 121.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1400,9 +1400,8 @@ limitations in the transport layer's capacity. This error indicates that
 resource constraints prevent the continuation of the path.
 
 UNSTABLE_INTERFACE (TBD-11): The endpoint is abandoning the path because
-of unstable interfaces. This error is used when endpoints find that
-the network interfaces are unstable due to weak signal or other possible situations.
-Endpoints could also choose to use this error when detecting a black hole on the specific path.
+the used interface is considered to be unstable. This conition can occur, e.g., 
+due to a weak wireless signal or frequent mobility events.
 
 NO_CID_AVAILABLE (TBD-12): The endpoint is abandoning the path due to
 the lack of an available connection ID for this path.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1142,7 +1142,7 @@ Path Identifier:
 Error Code:
 : A variable-length integer that indicates the reason for abandoning
   this path. NO_ERROR(0x0) indicates that the path is being abandoned
-  without any error being encountered.
+  without any error being encountered. Other error codes can be found in {{error-codes}}.
 
 PATH_ABANDON frames are ack-eliciting. If a packet containing
 a PATH_ABANDON frame is considered lost, the peer SHOULD repeat it.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1384,6 +1384,22 @@ it cannot allocate sufficient resources to maintain it. This is due to
 limitations in the transport layer's capacity. This error indicates that
 resource constraints prevent the continuation of the path.
 
+UNSTABLE_INTERFACE (TBD-11): The endpoint is abandoning the path because
+of unstable interfaces. This error is used when endpoints find that
```
gloinul, line 1388: Should this be interfaces in singular?

draft-ietf-quic-multipath.md
```diff
@@ -1384,6 +1384,22 @@ it cannot allocate sufficient resources to maintain it. This is due to
...
+the network interfaces are unstable due to weak signal or other possible situations.
+Endpoints could also choose to use this error when detecting a black hole on the specific path.
```
gloinul, line 1390: I think black hole can be replace by very high degrees of persistent packet loss, rather than using jargon.

draft-ietf-quic-multipath.md
```diff
@@ -1399,6 +1399,22 @@ it cannot allocate sufficient resources to maintain it. This is due to
...
```
mirjak, line 1405:
```suggestion
UNSTABLE_INTERFACE (TBD-11): The endpoint is abandoning the path because
the used interface is considered to be unstable. This conition can occur, e.g., 
due to a weak wireless signal or frequent mobility events.
```

draft-ietf-quic-multipath.md
```diff
@@ -1399,6 +1399,17 @@ it cannot allocate sufficient resources to maintain it. This is due to
...
 
+UNSTABLE_INTERFACE (TBD-11): The endpoint is abandoning the path because
+the used interface is considered to be unstable. This conition can occur, e.g., 
+due to a weak wireless signal or frequent mobility events.
+
+NO_CID_AVAILABLE (TBD-12): The endpoint is abandoning the path due to
+the lack of a connection ID for this path.
+This may occur when the peer initiates a new path
+but has not provided a corresponding connection ID for the path ID
+(or the packet containing the connection IDs has not arrived yet).
```
gloinul, line 1410: Should this error position itself in relation to the PATH_CID_BLOCKED in #468. Commonly a endpoint would first send PATH_CID_BLOCKED to indicate the lack of CID, and eventually it might decide to close the path using PATH_ABANDON with the NO_CID_AVAILABLE error code.

Yanmei-Liu: Yes, that's true. But we need to merge #468 first or we'll have an incorrect reference, and we also need to change the error code TBD-number at that point. I'd like to change these two parts after #468 is merged.

mirjak: Not sure if an endpoint would typically do that. It's is equally valid to just abandon the path right away and not have an half-open path dangling around. I'm not sure we want to give any guidance here at all.

draft-ietf-quic-multipath.md
```diff
@@ -1399,6 +1399,17 @@ it cannot allocate sufficient resources to maintain it. This is due to
 limitations in the transport layer's capacity. This error indicates that
 resource constraints prevent the continuation of the path.
 
+UNSTABLE_INTERFACE (TBD-11): The endpoint is abandoning the path because
+the used interface is considered to be unstable. This conition can occur, e.g.,
```
mirjak, line 1403:
```suggestion
the used interface is considered to be unstable. This conition can occur, e.g.,
```

draft-ietf-quic-multipath.md
```diff
@@ -1399,6 +1399,17 @@ it cannot allocate sufficient resources to maintain it. This is due to
...
```
qdeconinck, line 1403:
```suggestion
the used interface is considered to be unstable. This condition can occur, e.g.,
```

qdeconinck (approved):
I am fine with these error codes. Just a typo.

mirjak (approved):

draft-ietf-quic-multipath.md
```diff
@@ -1399,6 +1399,17 @@ it cannot allocate sufficient resources to maintain it. This is due to
 limitations in the transport layer's capacity. This error indicates that
 resource constraints prevent the continuation of the path.
 
+UNSTABLE_INTERFACE (TBD-11): The endpoint is abandoning the path because
+the used interface is considered to be unstable. This condition can occur, e.g.,
+due to a weak wireless signal or frequent mobility events.
```
yfmascgy, line 1404: What does "frequent mobility events" mean here? Maybe consider making this more clear.

mirjak: This is e.g. when you walk constantly in and out of coverage. However, these are really just supposed to be examples, so I don't think we need to go too much into detail. Which part is not clear? Do you have an alternative wording that you think would be more clear?

gloinul (approved):

draft-ietf-quic-multipath.md
gloinul: After some additional thought I think this is fine. Implementers will likely figure out what I wrote, or if not even that will be clear for the ones receiving PATH_ABANDON with this error code.

huitema (approved):

draft-ietf-quic-multipath.md
Yanmei-Liu:
```suggestion
due to a weak wireless signal or frequent handover events during high-speed mobility.
```

yfmascgy (approved):

Issue #457, opened 2024-11-02
Handling of PTO


gloinul:
Something that is not obvious how to handle in the draft is how one deals with cases when a path for a while drops all packets. 
The draft says in section: 

Data associated with the transmission and reception such RTT measurements, congestion control state, or loss recovery are maintained per packet number space and as such per Path ID.

So if one follows this one a PTO will eventually be triggered when there are no room to send more due to the congestion window. At this point from the paths perspective there are two things that can happen: 
1) A PTO functions and results in outstanding data being declared as lost. 
2) One abandon the path based on some other higher watch timer. 

There appear however be one issue with the above that is multipath specific. Namely that when the PTO fires the outstanding data could also be retransmitted on another path. But, following RFC 9002 the data will not be declared lost until a Probe actually makes it through to the peer and is being acked. 

Think it is worth clarifying this in the draft.

huitema:
On 11/2/2024 1:51 PM, Magnus Westerlund wrote:
> Something that is not obvious how to handle in the draft is how one deals with cases when a path for a while drops all packets.
> The draft says in section:
>
> Data associated with the transmission and reception such RTT measurements, congestion control state, or loss recovery are maintained per packet number space and as such per Path ID.
>
> So if one follows this one a PTO will eventually be triggered when there are no room to send more due to the congestion window.

This is actually a common scenario with Wi-Fi paths. We observe 
"suspension periods" of 100 or 200ms, much larger than a typical RTT of 
10 to 20ms.

>   At this point from the paths perspective there are two things that can happen:
> 1) A PTO functions and results in outstanding data being declared as lost.
> 2) One abandon the path based on some other higher watch timer.
Picoquic does something a bit different. If other paths are available, 
the packets will be declared lost, so their content can be resend 
immediately without risking delays, but the copy of the packets will be 
kept for a while. The lossy path will be placed in a "standby" state, no 
more content will be queued there. The stack will do a series of path 
challenges on the path, to check whether it comes back. If the path 
comes back it will be placed in available state again. If it turns out 
that packets were merely delayed instead of lost, it the acks will be 
processed, and the frames that were placed in the retransmission queue 
will be marked as received.

> There appear however be one issue with the above that is multipath specific. Namely that when the PTO fires the outstanding data could also be retransmitted on another path. But, following RFC 9002 the data will not be declared lost until a Probe actually makes it through to the peer and is being acked.
>
> Think it is worth clarifying this in the draft.

[...]

PR #467, opened 2024-11-11
Add guidance on handling of PTO

Yanmei-Liu:
To fix issue #457. 
We could add some implementation guidance, but how aggressive depends on the implementations. (according to the discussion in IETF 121.)

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -994,6 +994,17 @@ expected to be preserved when data is retransmitted. Especially when STREAM
 frames have to be retransmitted over a different path with a smaller MTU limit,
 new smaller STREAM frames might need to be sent instead.
 
+## Handling PTO
+
+An implementation should follow the mechanism specified in {{QUIC-RECOVERY}}
+for detecting packet loss on each individual path.
+When an endpoint transmits a significant number of packets on a specific path,
+and the path turned into a blackhole while acknowledgements can not be received from the path,
+endpoint waits for PTO on the specific path following {{QUIC-RECOVERY}}.
+Meanwhile, as a potential advantage of multipath extension, when endpoints detect
+that one of the paths has turned into a blackhole, endpoints could choose to
+retransmit on other available paths if the congestion control window allows.
+
 
 ## Handling different PMTU sizes
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -997,14 +997,16 @@ new smaller STREAM frames might need to be sent instead.
 ## Handling PTO
 
 An implementation should follow the mechanism specified in {{QUIC-RECOVERY}}
-for detecting packet loss on each individual path.
-When an endpoint transmits a significant number of packets on a specific path,
-and the path turned into a blackhole while acknowledgements can not be received from the path,
-endpoint waits for PTO on the specific path following {{QUIC-RECOVERY}}.
-Meanwhile, as a potential advantage of multipath extension, when endpoints detect
-that one of the paths has turned into a blackhole, endpoints could choose to
-retransmit on other available paths if the congestion control window allows.
+for detecting packet loss on each individual path. A special case happens when
+the PTO timer expires. According to {{QUIC-RECOVERY}}, no packet will be declared
+lost until either the packet sender receives a new acknowledgement for this path,
+or the path itself is finally declared broken. This cautious process minimizes
+the risk of spurious retransmissions, but it may cause significant delivery delay
+for the frames contained in these "lost packets".
 
+Endpoints could take advantage of the multipath extension, and retransmit the content
+of the delayed packets on other available paths if the congestion control window on these
+paths allows.
 
 ## Handling different PMTU sizes
 


draft-ietf-quic-multipath.md
```diff
@@ -994,6 +994,17 @@ expected to be preserved when data is retransmitted. Especially when STREAM
 frames have to be retransmitted over a different path with a smaller MTU limit,
 new smaller STREAM frames might need to be sent instead.
 
+## Handling PTO
+
+An implementation should follow the mechanism specified in {{QUIC-RECOVERY}}
+for detecting packet loss on each individual path.
+When an endpoint transmits a significant number of packets on a specific path,
+and the path turned into a blackhole while acknowledgements can not be received from the path,
```
gloinul, line 1002: I think the part after "into a blackhole" is a bit strange. One reason is that either acks sent on this path can't be received, or the equally relevant reason they may not be sent is that no forward packets on this path reaches the endpoint. Thus I suggest: 

When an endpoint transmits a significant number of packets on a specific path,
and the path turned into a blackhole resulting in that either no ACK is sent when no packets are received, or no ACKs sent on this path arrive at the packet sender, then the packet sender's probe timeout (PTO) will trigger following {{QUIC-RECOVERY}}. However, no packet's will be declared as lost until the packet sender receives an ACK for this path. To utilise the advantages of the multipath extension, when endpoints detect
that one of the paths has turned into a blackhole, endpoints could choose to
retransmit on other available paths if the congestion control window allows.

huitema: I think "black hole" is jargon. How about:

An implementation should follow the mechanism specified in {{QUIC-RECOVERY}}
for detecting packet loss on each individual path. A special case happens when
the PTO timer expires. According to {{QUIC-RECOVERY}}, no packet's will be declared
as lost until the packet sender receives an ACK for this path, or the path itself is finally declared
broken. This cautious process minimizes the risk of undue retransmissions,
but is may cause significant delivery delay for the frames contained in these "lost packets".

Endpoints could take advantage of the multipath extension, and retransmit the content
of the delayed packets on other available paths if the congestion control window on these
paths allows.

mirjak:
Thanks @huitema, your proposed text looks good. I would suggest to not only say "retransmit on another path" but say "retransmit on another path earlier even before the originally used path is abandon" or something. Do we even want to recommend or hint when to retransmit?

Also do we need to add some normative language in the connection closure section that all packets need to be declared lost after path abandon?

huitema:
We already have this text in Path Close:

_...However, knowledge of the connection identifiers received from the peer and of the state of the number space associated to the path SHOULD be retained while packets from the peer might still be in transit, i.e., for a delay of 3 PTO after the PATH_ABANDON frame has been received from the peer, both to avoid generating spurious stateless packets as specified in {{spurious-stateless-reset}} and to be able to acknowledge the last packets received from the peer as specified in {{ack-after-abandon}}._

_After receiving or sending a PATH_ABANDON frame, the endpoints SHOULD promptly send PATH_ACK frames to acknowledge all packets received on the path and not yet acknowledged, as specified in {{ack-after-abandon}}. When an endpoint finally deletes all resource associated with the path, the packets sent over the path and not yet acknowledged MUST be considered lost._

I think that's pretty clear, no need to add more details on when packets are considered lost.

draft-ietf-quic-multipath.md
gloinul: This proposal works for me.

gloinul (approved):

mirjak:
Ah, thanks I missed that last MUST. So that is pretty clear and already addresses. No change needed.

Then only my other question remains: Do we want to give more clear advice when to retransmit? Like after one PTO?

huitema:
I don't think we should give more guidance. Mostly because we don't really know. When we will have lots of deployment experience, maybe. But for now, just trust implementers. They cannot merely translate a spec into code, they have to think, and I really expect that different implementers will need to make different tradeoffs based on their deployments.

huitema:
oops! Wrong window. I was working on a PR on a different project and just merged this one. Sorry.

mirjak:
There is a "revert" button if needed. However, I guess this was anyway more or less ready to merge.

I would just have had a few more editorial things but I guess I can also create another (editorial) PR.

For this sentence I think we should say somehow that you can retransmit _earlier_ before the packet is declared lost, otherwise the term "retransmit" might be confusing:

OLD
Endpoints could take advantage of the multipath extension, and retransmit the content
of the delayed packets on other available paths if the congestion control window on these
paths allows.

NEW
Endpoints can take advantage of the multipath extension, and retransmit the content
of the delayed packets on other available paths before they are declared lost
if the congestion control window on these paths allows.

Also not sure if we need the "if the congestion control window on these paths allows" part but I guess it doesn't hurt.

And then I would propose to replace "the path itself is finally declared broken" with "the path is explicitly abandon". However, also note that this is exactly the issue because as we don't have/require an idle timeout, you can basically keep a broken path open forever (potentially hoping it will come back one day) and then never declare those packets as lost. I think we need to mention that explicitly somehow.

Issue #459, opened 2024-11-03
Clarify interaction between PATHS_BLOCKED and PATH_NEW_CONNECTION_ID


gloinul:
In the MP-QUIC interop testing at IETF 121 it was observed that an issue when testing PATHS_BLOCKED is that some servers where blocking path creation by not providing CIDS for some of the path IDs that it allowed. Thus, the client may desire to open more paths, there are agreed on path IDs, but lacks the CIDs. If the PATHS_BLOCKED is usable, how does one handle the situation when the server block the client by not providing CIDs? 

I believe some clarifications and possible requirements are needed. I think there are two roads to take here:

1) Require that endpoints provide PATH_NEW_CONNECTION_ID when there are no CIDs for an unused PATH ID, as long as below MAX_PATH_IDS. 
2) Create an equivalent or parameter on PATHS_BLOCKED indicating what is the issue.

huitema:
For the record, the observation during interop was due to an off-by-1 bug in picoquic, which has since be fixed. But the sitiation could be repeated.

There are a couple reasons why an endpoint might delay provision of CIDs. The Path CID frames can only be sent after the max path id is negotiated, so there will always be a delay. That delay will be longer if packets are lost and need to be repeated. If the client is getting impatient, I think that something like "PATH_CID_BLOCKED" with a path_id parameter would make sense. If after sending that the client still does not receive the expected CID, it can escalate. Maybe abandon the path.

huitema:
Should we change the title of the issue? It seems to be about providing CID for all paths. But this is not the only clarification needed. The interop tests also show mismatched expectations, with some stacks apparently expecting to promptly receive a new MAX_PATH_ID frame after sending a PATH_BLOCKED frame. Of course, that's not happening. If the number of allocated paths has reached the limit supported by the server, the MAX_PATH_ID value can only increase if some allocated paths are abandoned. We should maybe clarify that PATH_BLOCKED is purely informational, and that it is not expected to trigger changes in the state of the peer.

gloinul:
So picoquiche had the off-by-one issue. QUICHE does not provide CIDs period up to the level of max_initial_path_ids. Thus, I think we need to think about how to deal with this situation. I agree that there are reasons for delaying provisioning CIDs for Paths, but what a peer can expect needs to be made clear.

mirjak:
Based on the discussion at IETF 121 we will propose a PR to add a field to the PATHS_BLOCKED frame and an error code the PATH_ABANDON frame.

PR #468, opened 2024-11-18
Add new frame type: PATH_CIDS_BLOCKED frame

Yanmei-Liu:
To solve issue #459 according to comments in IETF-121

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1358,20 +1358,16 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
 limit set by its peer.
-If the Type field within the PATHS_BLOCKED frame is TBD-09 (used in experiments as 0x15228c0e),
-this signifies that new paths are blocked because there are no unused Connection IDs available
-for the corresponding Path IDs.
 
 Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
 imply a particular action from the peer like updating the new Max Path ID value,
-but informs the peer that the maximum path identifier limit prevented the creation of new paths,
-or there are no available Connection IDs for the path.
+but informs the peer that the maximum path identifier limit prevented the creation of new paths.
 
 PATHS_BLOCKED frames are formatted as shown in {{fig-paths-blocked-frame-format}}.
 
 ~~~
 PATHS_BLOCKED Frame {
-  Type (i) = TBD-08..TBD-09 (experiments use 0x15228c0d.. 0x15228c0e),
+  Type (i) = TBD-08 (experiments use 0x15228c0d),
   Maximum Path Identifier (i),
 }
 ~~~
@@ -1385,9 +1381,30 @@ Maximum Path Identifier:
   the currently allowed maximum value, this frame can be ignored.
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
-  If the type field is TBD-09(experiments use 0x15228c0e), this indicates
-  that the path identifier cannot be used to create new paths because
-  there are no unused connection IDs available for this path.
+
+
+## PATH_CID_BLOCKED frames {#path-cid-blocked-frame}
+
+A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
+it wishes to open a path but is unable to do so because there are no unused Connection IDs available
+for the corresponding Path IDs.
+
+Note that PATH_CID_BLOCKED frame is informational. Sending a PATH_CID_BLOCKED frame does not
+imply a particular action from the peer, but informs the peer that the absence of unused Connection IDs
+prevented the creation of this path.
+
+PATH_CID_BLOCKED frames are formatted as shown in {{fig-path-cid-blocked-frame-format}}.
+~~~
+PATHS_BLOCKED Frame {
+  Type (i) = TBD-09 (experiments use 0x15228c0e),
+  Path Identifier (i),
+}
+~~~
+{: #fig-path-cid-blocked-frame-format title="PATH_CID_BLOCKED Frame Format"}
+
+Path Identifier:
+  : Identifier of the path for which unused Connection IDs are not available.
+
 
 # Error Codes {#error-codes}
 
@@ -1436,7 +1453,7 @@ TBD-05 (experiments use 0x15228c09)                  | PATH_NEW_CONNECTION_ID   
 TBD-06 (experiments use 0x15228c0a)                  | PATH_RETIRE_CONNECTION_ID| {{mp-retire-conn-id-frame}}
 TBD-07 (experiments use 0x15228c0c)                  | MAX_PATH_ID            | {{max-paths-frame}}
 TBD-08 (experiments use 0x15228c0d)                  | PATHS_BLOCKED    | {{paths-blocked-frame}}
-TBD-09 (experiments use 0x15228c0e)                  | PATHS_BLOCKED    | {{paths-blocked-frame}}
+TBD-09 (experiments use 0x15228c0e)                  | PATH_CID_BLOCKED | {{path-cid-blocked-frame}}
 {: #frame-types title="Addition to QUIC Frame Types Entries"}
 
 The following transport error code defined in {{tab-error-code}} are to
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1371,7 +1371,7 @@ for the corresponding Path IDs.
 Note that PATHS_BLOCKED and PATH_CIDS_BLOCKED frames are informational.
 Sending a PATHS_BLOCKED or a PATH_CIDS_BLOCKED frame does not imply a particular action from the peer
 like updating the new Max Path ID value, but informs the peer that the maximum path identifier limit
-or the absence of unused connection IDs prevented the creation of new paths.
+or the absence of unused connection IDs prevented the creation or the usage of paths.
 
 PATHS_BLOCKED frames are formatted as shown in {{fig-paths-blocked-frame-format}}.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -369,8 +369,7 @@ path identifiers,
 - PATHS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
 to notify the peer of being blocked to open new paths as
 the limit of active paths set by the peer has been reached
-- PATH_CIDS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
-to notify the peer there are no unused connection IDs available
+or there are no unused connection IDs available
 for the corresponding Path ID.
 
 All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1386,7 +1386,8 @@ Maximum Path Identifier:
 ## PATH_CID_BLOCKED frames {#path-cid-blocked-frame}
 
 A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
-it wishes to open a path but is unable to do so because there are no unused connection IDs available
+it wishes to open a path or change the CID on an established path
+but is unable to do so because there are no unused connection IDs available
 for the corresponding Path IDs.
 
 Note that PATH_CID_BLOCKED frame is informational. Sending a PATH_CID_BLOCKED frame does not
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -371,7 +371,7 @@ to notify the peer of being blocked to open new paths as
 the limit of active paths set by the peer has been reached
 - PATH_CIDS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
 to notify the peer there are no unused connection IDs available
-for the corresponding Path IDs.
+for the corresponding Path ID.
 
 All new frames are sent in 1-RTT packets {{QUIC-TRANSPORT}}.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1359,11 +1359,11 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
 ## PATHS_BLOCKED and PATH_CIDS_BLOCKED frames {#paths-and-cids-blocked-frame}
 
-A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
+A sender can send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
 limit set by its peer.
 
-A sender SHOULD send a PATH_CIDS_BLOCKED frame (type=0x15228c0e) when
+A sender can send a PATH_CIDS_BLOCKED frame (type=0x15228c0e) when
 it wishes to open a path with a valid Path ID or change the CID on an established path
 but is unable to do so because there are no unused connection IDs available
 for the corresponding Path ID.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1358,16 +1358,20 @@ MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
...
+If the Type field in the PATHS_BLOCKED frame is TBD-09(experiments use 0x15228c0e),
+it indicates that new paths are blocked because of lacking unused Connection IDs
+for the corresponding Path ID.
+
 Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
 imply a particular action from the peer like updating the new Max Path ID value,
-but informs the peer that the maximum path identifier limit prevented the creation of new paths.
-
+but informs the peer that the maximum path identifier limit prevented the creation of new paths
+or the lacking of new Connection IDs.
```
gloinul, line 1368: I don't think "lacking" works here grammatically, either use "lack of Connection IDs for any unused path ID".

gloinul (commented):
Some language adjustments needed, but otherwise fine.

huitema (approved):
I think the text could be a bit simpler, but this is what we want to do.

mirjak:
I don't think it is useful to make this new frame a variant of the PATHS_BLOCKD frame because for the this signal the max_path_id value is not useful. I would prefer for clarity to have a new PATHS_CID_BLOCK frames which either has no fields or a path ID field that indicates an unused path ID that you would like to get CIDs for or something like that if people think such information would be useful...?

huitema:
There is no technical difference between what @mirjak suggests and the current PR. In both cases, we get two frame types, with two different parameters, and a path_id values. The big difference is having two frame names (@mirjak) instead of one (current). The "two frame name" variant is easier to document, and probably easier to understand by implementers. For the PR, that would mean:

* remove the proposed changes to PATHS_BLOCKED frames 
* add the description of a PATH_CID_BLOCKED frame

That description could be:
```
## PATH_CID_BLOCKED frames {#path-cid-blocked-frame}
A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
it wishes to open a path but is unable to do so because there are no unused Connection IDs available
for the corresponding Path IDs.

Note that PATH_CID_BLOCKED frame is informational. Sending a PATH_CID_BLOCKED frame does not
imply a particular action from the peer like updating the new Max Path ID value,
but informs the peer that the absence of unused Connection IDs prevented the creation of this path.

PATH_CID_BLOCKED frames are formatted as shown in {{fig-path-cid-blocked-frame-format}}.
~~~
PATHS_BLOCKED Frame {
  Type (i) = TBD-09 (experiments use 0x15228c0e),
  Path Identifier (i),
}
~~~
{: #fig-path-cid-blocked-frame-format title="PATH_CID_BLOCKED Frame Format"}

Path Identifier:
  : Identifier of the path for which unused Connection IDs are not available.
```

Note that there is a difference between PATHS_BLOCKED and PATH_CID_BLOCKED frame formats: in one case the parameter is a number of paths, in the other case a path identifier.

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,30 @@ Maximum Path Identifier:
...
+it wishes to open a path but is unable to do so because there are no unused Connection IDs available
+for the corresponding Path IDs.
+
+Note that PATH_CID_BLOCKED frame is informational. Sending a PATH_CID_BLOCKED frame does not
+imply a particular action from the peer, but informs the peer that the absence of unused Connection IDs
+prevented the creation of this path.
+
+PATH_CID_BLOCKED frames are formatted as shown in {{fig-path-cid-blocked-frame-format}}.
+~~~
+PATHS_BLOCKED Frame {
```
mirjak, line 1398:
```suggestion
PATHS_CID_BLOCKED Frame {
```

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,30 @@ Maximum Path Identifier:
...
+
+A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
+it wishes to open a path but is unable to do so because there are no unused Connection IDs available
+for the corresponding Path IDs.
+
+Note that PATH_CID_BLOCKED frame is informational. Sending a PATH_CID_BLOCKED frame does not
+imply a particular action from the peer, but informs the peer that the absence of unused Connection IDs
+prevented the creation of this path.
+
+PATH_CID_BLOCKED frames are formatted as shown in {{fig-path-cid-blocked-frame-format}}.
```
mirjak, line 1396:
```suggestion
PATH_CID_BLOCKED frames are formatted as shown in {{fig-path-cid-blocked-frame-format}}.

```

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,31 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+
+## PATH_CID_BLOCKED frames {#path-cid-blocked-frame}
+
+A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
+it wishes to open a path but is unable to do so because there are no unused Connection IDs available
```
mirjak, line 1389:
```suggestion
it wishes to open a path but is unable to do so because there are no unused connection IDs available
```

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,31 @@ Maximum Path Identifier:
...
+for the corresponding Path IDs.
+
+Note that PATH_CID_BLOCKED frame is informational. Sending a PATH_CID_BLOCKED frame does not
+imply a particular action from the peer, but informs the peer that the absence of unused Connection IDs
```
mirjak, line 1393:
```suggestion
imply a particular action from the peer, but informs the peer that the absence of unused connection IDs
```

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,31 @@ Maximum Path Identifier:
...
+~~~
+PATHS_CID_BLOCKED Frame {
+  Type (i) = TBD-09 (experiments use 0x15228c0e),
+  Path Identifier (i),
+}
+~~~
+{: #fig-path-cid-blocked-frame-format title="PATH_CID_BLOCKED Frame Format"}
+
+Path Identifier:
+: Identifier of the path for which unused Connection IDs are not available.
```
mirjak, line 1407:
```suggestion
: Identifier of the path for which unused connection IDs are not available.
```

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,31 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+
+## PATH_CID_BLOCKED frames {#path-cid-blocked-frame}
+
+A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
```
mirjak, line 1388: I think we should actually change this "SHOULD" to "can". There is no requirement in interoperability that require the sender to send this. Or if we want normative language we should use "MAY".

The same comment of course applies to the PATHS_BLOCKED frame of course. Sorry I overlooked this earlier.

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,31 @@ Maximum Path Identifier:
   Receipt of a value that is higher than the local maximum value MUST
   be treated as a connection error of type PROTOCOL_VIOLATION.
 
+
+## PATH_CID_BLOCKED frames {#path-cid-blocked-frame}
```
mirjak, line 1386: Is this PATH_CID_BLOCKED or PATH_CIDS_BLOCKED?

We use plural for PATHS_BLOCKED to align with STREAMS_BLOCKED in rfc9000

draft-ietf-quic-multipath.md
```diff
@@ -1382,6 +1382,31 @@ Maximum Path Identifier:
...
+
+A sender SHOULD send a PATH_CID_BLOCKED frame (type=0x15228c0e) when
+it wishes to open a path but is unable to do so because there are no unused connection IDs available
```
mirjak, line 1389:
```suggestion
it wishes to open a path or change the CID on an established path
but is unable to do so because there are no unused connection IDs available
```
I think for CIDs we need to cover this case as well.

mirjak:
I think it good to have a separate frame but we could editorially consider to still specify the two frames in the same section as we do for PATH_BACKUP and PATH_AVAILABLE. Note sure which option is better but given we have more or less some duplicate sentences here, I wanted to bring this option up.

draft-ietf-quic-multipath.md
Yanmei-Liu: Updated.

draft-ietf-quic-multipath.md
```diff
@@ -1353,15 +1357,21 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
...
+## PATHS_BLOCKED and PATH_CIDS_BLOCKED frames {#paths-and-cids-blocked-frame}
 
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
 limit set by its peer.
-Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
-imply a particular action from the peer like updating the new Max Path ID value,
-but informs the peer that the maximum path identifier limit prevented the creation of new paths.
 
+A sender SHOULD send a PATH_CIDS_BLOCKED frame (type=0x15228c0e) when
```
mirjak, line 1366: I think we should actually change this "SHOULD" to "can". There is no requirement in interoperability that require the sender to send this. Or if we want normative language we should use "MAY".

The same comment of course applies to the PATHS_BLOCKED frame of course. Sorry I overlooked this earlier.

Yanmei-Liu: I'd like to vote for "SHOULD" here. Basically it's a recommendation action in this kind of situation.

BTW, PATHS_BLOCKED and PATH_CIDS_BLOCKED are very similar to STREAMS_BLOCKED and we could find a "SHOULD" in [RFC 9000 Section 19.14](https://datatracker.ietf.org/doc/html/rfc9000#name-streams_blocked-frames):

> A sender SHOULD send a STREAMS_BLOCKED frame (type=0x16 or 0x17) when it wishes to open a stream but is unable to do so due to the maximum stream limit set by its peer...

draft-ietf-quic-multipath.md
```diff
@@ -358,16 +358,20 @@ on different 4-tuples due to, e.g., NAT rebinding. In such a case, the receiver
...
 - MAX_PATH_ID frame (see {{max-paths-frame}}) for increasing the limit of
-path identifiers, and PATHS_BLOCKED frame (see {{paths-blocked-frame}})
+path identifiers,
+- PATHS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
 to notify the peer of being blocked to open new paths as
-the limit of active paths set by the peer has been reached.
+the limit of active paths set by the peer has been reached
+- PATH_CIDS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
+to notify the peer there are no unused connection IDs available
+for the corresponding Path IDs.
```
qdeconinck, line 374:
```suggestion
for the corresponding Path ID.
```

draft-ietf-quic-multipath.md
```diff
@@ -1353,15 +1357,21 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
...
 it wishes to open a path but is unable to do so due to the maximum path identifier
 limit set by its peer.
-Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
-imply a particular action from the peer like updating the new Max Path ID value,
-but informs the peer that the maximum path identifier limit prevented the creation of new paths.
 
+A sender SHOULD send a PATH_CIDS_BLOCKED frame (type=0x15228c0e) when
+it wishes to open a path or change the CID on an established path
+but is unable to do so because there are no unused connection IDs available
+for the corresponding Path IDs.
```
qdeconinck, line 1369:
```suggestion
for the corresponding Path ID.
```

draft-ietf-quic-multipath.md
```diff
@@ -1353,15 +1357,21 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
...
 
 A sender SHOULD send a PATHS_BLOCKED frame (type=0x15228c0d) when
 it wishes to open a path but is unable to do so due to the maximum path identifier
 limit set by its peer.
-Note that PATHS_BLOCKED frame is informational. Sending a PATHS_BLOCKED frame does not
-imply a particular action from the peer like updating the new Max Path ID value,
-but informs the peer that the maximum path identifier limit prevented the creation of new paths.
 
+A sender SHOULD send a PATH_CIDS_BLOCKED frame (type=0x15228c0e) when
+it wishes to open a path or change the CID on an established path
```
qdeconinck, line 1367: To make the distinction clearer than with the PATHS_BLOCKED frame case.
```suggestion
it wishes to open a path with a valid Path ID or change the CID on an established path
```

draft-ietf-quic-multipath.md
```diff
@@ -1353,15 +1357,21 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
...
+but is unable to do so because there are no unused connection IDs available
+for the corresponding Path IDs.
+
+Note that PATHS_BLOCKED and PATH_CIDS_BLOCKED frames are informational.
+Sending a PATHS_BLOCKED or a PATH_CIDS_BLOCKED frame does not imply a particular action from the peer
+like updating the new Max Path ID value, but informs the peer that the maximum path identifier limit
+or the absence of unused connection IDs prevented the creation of new paths.
```
qdeconinck, line 1374: A endpoint may avoid using the same CID when it experienced active migration.
```suggestion
or the absence of unused connection IDs prevented the creation or the usage of paths.
```

yfmascgy (approved):

draft-ietf-quic-multipath.md
mirjak: Actually I think this is slightly different. In case of streams there is no really good reason to limit streams too much. Of course you want to have a limit to avoid excessive resource usage because that would open an attack, however, under normal operation you should not be blocked. In case of the number of paths, there can be other reasons to limit, e.g. you may have a policy to not use more than a small number of paths. Therefore I argue that MAY would make more sense here than SHOULD. However, I don't have a really strong opinion. I would like to hear other people's thoughts.

huitema (approved):
Approve, but as @mirjak replace SHOULD by MAY or can.

draft-ietf-quic-multipath.md
Yanmei-Liu: That's fair enough. For the reason of anti-attack, I agree to use "can" here.

draft-ietf-quic-multipath.md
```diff
@@ -358,16 +358,20 @@ on different 4-tuples due to, e.g., NAT rebinding. In such a case, the receiver
...
+This proposal adds six multipath control frames for path management:
 
 - PATH_ABANDON frame for the receiver side to abandon the path
 (see {{path-abandon-frame}}),
 - PATH_BACKUP and PATH_AVAILABLE frames to express a preference
 in path usage (see {{path-backup-available-frame}}), and
 - MAX_PATH_ID frame (see {{max-paths-frame}}) for increasing the limit of
-path identifiers, and PATHS_BLOCKED frame (see {{paths-blocked-frame}})
+path identifiers,
+- PATHS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
```
mirjak, line 369: Editorial suggestion to combine these to one bullet point again (total 3 proposed commits): 
```suggestion
- PATHS_BLOCKED and PATH_CID_BLOCKED frames (see {{paths-and-cids-blocked-frame}})
```

draft-ietf-quic-multipath.md
```diff
@@ -358,16 +358,20 @@ on different 4-tuples due to, e.g., NAT rebinding. In such a case, the receiver
...
 to notify the peer of being blocked to open new paths as
-the limit of active paths set by the peer has been reached.
+the limit of active paths set by the peer has been reached
+- PATH_CIDS_BLOCKED frame (see {{paths-and-cids-blocked-frame}})
+to notify the peer there are no unused connection IDs available
+for the corresponding Path ID.
```
mirjak, line 374:
```suggestion
or there are no unused connection IDs available
for the corresponding Path ID.
```

PR #469, opened 2024-11-28
recommend use of new paths instead of migration

### LINKED ISSUE #458: Clarification on Path Migration in relation to opening a new path
Path Migration for the case when the client is not aware of the change, like for NAT address rebinding case can't be avoided. Thus, this needs to be supported by server as something that can occur. However, if a client have a new address it could technically select to migrate an existing path to this new path. However, it appears more suitable to have a very strong recommendation here to actually open a new path in these cases. Combined possibly with abandoning another existing path. 

This issue is based on the interop testing.

### LINKED ISSUE #485: Managing ACK and repeat of PATH_CID_BLOCKED frame
Suppose that the packet carrying a PATH_CID_BLOCKED frame is lost. How does the implementation decide whether that frame should be repeated?

For the paths blocked frame, we have a good logic: if a the packet containing a path blocked frame is lost, compared the "pax path id" in the blocked frame to the highest value of the max path id in a PATHS_BLOCKED frame acknowledged by the peer. If the value is higher than the retained value, repeat. Otherwise, don't. But we don't have anything like that for the PATH_CID_BLOCKED blocked frame.

One possibility is to repeat it no matter what, but this may create extra traffic. Another is to never repeat it, but that means losing the signal. Yet another would be to remember if the PATH_CID_BLOCKED frame for that path was already acknowledged, but what if we need to send several such frames for the same path over time? We could also test whether the path is still blocked.

I think this is a bit under-specified.

mirjak:
fixes #458

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1028,7 +1028,7 @@ Different applications will likely require different strategies.
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
-## Connection ID Changes 
+## Connection ID Changes
 
 {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
 can change the connection ID it uses to another available one
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1028,13 +1028,15 @@ Different applications will likely require different strategies.
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
-## Connection ID Changes and NAT Rebindings {#migration}
+## Connection ID Changes 
 
 {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
 can change the connection ID it uses to another available one
 at any time during the connection. As such a sole change of the Connection
 ID without any change in the address does not indicate a path change and
 the endpoint can keep the same congestion control and RTT measurement state.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1028,14 +1028,16 @@ Different applications will likely require different strategies.
 Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
 on each of these paths before the idle timeout expires.
 
-## Connection ID Changes and NAT Rebindings {#migration}
+## Connection ID Changes
```
mirjak, line 1031:
```suggestion
## Connection ID Changes
```

draft-ietf-quic-multipath.md
```diff
@@ -1050,6 +1052,13 @@ multipath extension is used and an endpoint has multiple addresses that
 could lead to switching between different paths, it should rather maintain
 multiple open paths instead.
 
+More generally, while migration cannot be avoided in case of network-based
+NAT rebindings, opening a new path instead of active client migration
+should be strongly perferened when the multipath extension is supported.
+This enables a smoother handover and allows a simplified migration
+handling at the server as NAT rebindings imply immediate loss of the old
+address.
+
```
huitema, line 1061: typo: perferened -> preferred.

huitema (approved):
Looks fine.

Yanmei-Liu (approved):

gloinul (approved):

yfmascgy (approved):
Looks good to me.

qdeconinck (approved):
Good for me

PR #474, opened 2024-12-11
Update transport parameter for draft-12

Yanmei-Liu:
As we have add new frame type PATH_CIDS_BLOCKED frame, we'd need a new transport parameter type.
BTW, as the transport parameter type is hex value, I used 0x0c for draft-12.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1456,7 +1456,7 @@ the "QUIC Transport Parameters" registry under the "QUIC Protocol" heading.
 
 Value                                         | Parameter Name.   | Specification
 ----------------------------------------------|-------------------|-----------------
-TBD (current version uses 0x0f739bbc1b666d12) | initial_max_path_id | {{nego}}
+TBD (current version uses 0x0f739bbc1b666d0c) | initial_max_path_id | {{nego}}
 {: #transport-parameters title="Addition to QUIC Transport Parameters Entries"}
 
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -252,7 +252,7 @@ the use of the multipath extension during the connection handshake,
 as specified in {{QUIC-TRANSPORT}}. The new transport parameter is
 defined as follows:
 
-- initial_max_path_id (current version uses 0x0f739bbc1b666d0c): a
+- initial_max_path_id (current version uses 0x0f739bbc1b666d12): a
   variable-length integer specifying the maximum path identifier
   an endpoint is willing to maintain at connection initiation.
   This value MUST NOT exceed 2^32-1, the maximum allowed value for the Path ID due to
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -252,7 +252,7 @@ the use of the multipath extension during the connection handshake,
 as specified in {{QUIC-TRANSPORT}}. The new transport parameter is
 defined as follows:
 
-- initial_max_path_id (current version uses 0x0f739bbc1b666d11): a
+- initial_max_path_id (current version uses 0x0f739bbc1b666d0c): a
   variable-length integer specifying the maximum path identifier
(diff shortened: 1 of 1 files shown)


mirjak:
yes, it's an hex value but I thought the intention was to map the last to bits to the decimal revision number of the draft. At least that was my understanding of what we did so far, so let's not change this now.

draft-ietf-quic-multipath.md
```diff
@@ -252,7 +252,7 @@ the use of the multipath extension during the connection handshake,
 as specified in {{QUIC-TRANSPORT}}. The new transport parameter is
 defined as follows:
 
-- initial_max_path_id (current version uses 0x0f739bbc1b666d11): a
+- initial_max_path_id (current version uses 0x0f739bbc1b666d0c): a
```
Yanmei-Liu, line 255:
```suggestion
- initial_max_path_id (current version uses 0x0f739bbc1b666d12): a
```

draft-ietf-quic-multipath.md
```diff
@@ -1456,7 +1456,7 @@ the "QUIC Transport Parameters" registry under the "QUIC Protocol" heading.
 
 Value                                         | Parameter Name.   | Specification
 ----------------------------------------------|-------------------|-----------------
-TBD (current version uses 0x0f739bbc1b666d11) | initial_max_path_id | {{nego}}
+TBD (current version uses 0x0f739bbc1b666d0c) | initial_max_path_id | {{nego}}
```
Yanmei-Liu, line 1459:
```suggestion
TBD (current version uses 0x0f739bbc1b666d12) | initial_max_path_id | {{nego}}
```

Yanmei-Liu:
> yes, it's an hex value but I thought the intention was to map the last two values to the decimal revision number of the draft. At least that was my understanding of what we did so far, so let's not change this now.

As an engineer, I thought it would be strange to mix hex and decimal values together, a little bit tricky. 
But it's not a major issue and I updated  the last two values with "0x12", please have a look again!

huitema:
We have been using hex values for QUIC version numbers for some time. See https://github.com/quicwg/base-drafts/wiki/QUIC-Versions. The convention was to use version number '0xFF0000xx' for the draft N, were 'xx' was the hexadecimal notation for n. For example, the version for draft-34 was '0xFF000022'.

I think we could just keep the '0c' value for draft 12.

draft-ietf-quic-multipath.md
```diff
@@ -252,7 +252,7 @@ the use of the multipath extension during the connection handshake,
 as specified in {{QUIC-TRANSPORT}}. The new transport parameter is
 defined as follows:
 
-- initial_max_path_id (current version uses 0x0f739bbc1b666d11): a
+- initial_max_path_id (current version uses 0x0f739bbc1b666d12): a
```
Yanmei-Liu, line 255:
```suggestion
- initial_max_path_id (current version uses 0x0f739bbc1b666d0c): a
```

draft-ietf-quic-multipath.md
```diff
@@ -1456,7 +1456,7 @@ the "QUIC Transport Parameters" registry under the "QUIC Protocol" heading.
 
 Value                                         | Parameter Name.   | Specification
 ----------------------------------------------|-------------------|-----------------
-TBD (current version uses 0x0f739bbc1b666d11) | initial_max_path_id | {{nego}}
+TBD (current version uses 0x0f739bbc1b666d12) | initial_max_path_id | {{nego}}
```
Yanmei-Liu, line 1459:
```suggestion
TBD (current version uses 0x0f739bbc1b666d0c) | initial_max_path_id | {{nego}}
```

Yanmei-Liu (commented):
According to @huitema 's comments, I keep the hex value version in this commit if we decide to keep it.

mirjak:
Sure we can do one or the other but we have used the (decimal) draft version number so far, so why should we change it now?

huitema (approved):

PR #475, opened 2024-12-11
Note that path state cannot be removed without reception of PATH_ABANDON

### LINKED ISSUE #471: If the peer does not reply with a path_abandon frame, state can never be removed
Currently the draft only says:

"knowledge of the connection identifiers received from the peer and of the state of the number space associated to the path SHOULD be retained while packets from the peer might still be in transit, i.e., for a delay of 3 PTO after the PATH_ABANDON frame has been received from the peer"

Therefore state removal is triggered by reception of the path abandon frame. There is MUST in the text that my peer must send me a path_abandon frame if it has received mine. What if it doesn't send a path_abandon frame ever?

The sentence above is only a SHOULD, so I guess you could remove state any reasonable time after sending the path_abandon frame, however, maybe we should say more explicitly that it makes sense to also start a timer after the ACK for the path_abandon frame was received, or stop retransmitting the path_abandon frame after X tries and then just remove state?

mirjak:
fixes #471

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -544,7 +544,7 @@ last packets received from the peer as specified in {{ack-after-abandon}}.
 If a peer sends a PATH_ABANDON frame but never receives
 a corresponding PATH_ABANDON frame, it might not be able to remove path state.
 It is left to the implementation to handle this unexpected
-behavior as it does not impact interoperability. If the endpoint is no longer 
+behavior as it does not impact interoperability. If the endpoint is no longer
 willing to process the issued connection IDs for the abandoned path,
 it MAY close the connection, but SHOULD wait at least 3 PTOs after
 sending the PATH_ABANDON frame.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -544,8 +544,10 @@ last packets received from the peer as specified in {{ack-after-abandon}}.
 If a peer sends a PATH_ABANDON frame but never receives
 a corresponding PATH_ABANDON frame, it might not be able to remove path state.
 It is left to the implementation to handle this unexpected
-behavior as it does not impact interoperability. However, an endpoint SHOULD NOT
-remove path state earlier than 3 PTOs after sending the PATh_ABANDON frame.
+behavior as it does not impact interoperability. If the endpoint is no longer 
+willing to process the issued connection IDs for the abandoned path,
+it MAY close the connection, but SHOULD wait at least 3 PTOs after
+sending the PATH_ABANDON frame.
 
 After receiving or sending a PATH_ABANDON frame, the endpoints SHOULD
 promptly send PATH_ACK frames to acknowledge all packets received on
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -542,11 +542,11 @@ both to avoid generating spurious stateless packets as specified in
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
 This means path state removal is supposed to be triggered by the reception of a
-PATH_ABANDON frame. If a peer sends an PATH_ABANDON frame but never receives
+PATH_ABANDON frame. If a peer sends a PATH_ABANDON frame but never receives
 a corresponding PATH_ABANDON frame, it either might not be able to remove path state or
 could use additional timers, e.g. after receiving the corresponding ACK frame,
 or other conditions like a maximum number of retransmissions to remove state.
-If path state is remove without receiving a PATH_ABANDON frame this could lead
+If path state is removed without receiving a PATH_ABANDON frame this could lead
 to packet loss. It is left to the implementation to handle this unexpected
 behavior as it does not impact interoperability.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -541,6 +541,15 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+This means path state removal is supposed to be triggered by the reception of a
+PATH_ABANDON frame. If a peer sends an PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it either might not be able to remove path state or
+could use additional timers, e.g. after receiving the corresponding ACK frame,
+or other conditions like a maximum number of retransmissions to remove state.
+If path state is remove without receiving a PATH_ABANDON frame this could lead
+to packet loss. It is left to the implementation to handle this unexpected
+behavior as it does not impact interoperability.
+
 After receiving or sending a PATH_ABANDON frame, the endpoints SHOULD
 promptly send PATH_ACK frames to acknowledge all packets received on
 the path and not yet acknowledged, as specified in {{ack-after-abandon}}.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -544,7 +544,8 @@ last packets received from the peer as specified in {{ack-after-abandon}}.
 If a peer sends a PATH_ABANDON frame but never receives
 a corresponding PATH_ABANDON frame, it might not be able to remove path state.
 It is left to the implementation to handle this unexpected
-behavior as it does not impact interoperability.
+behavior as it does not impact interoperability. However, an endpoint SHOULD NOT
+remove path state earlier than 3 PTOs after sending the PATh_ABANDON frame.
 
 After receiving or sending a PATH_ABANDON frame, the endpoints SHOULD
 promptly send PATH_ACK frames to acknowledge all packets received on
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -541,13 +541,9 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
-This means path state removal is supposed to be triggered by the reception of a
-PATH_ABANDON frame. If a peer sends a PATH_ABANDON frame but never receives
-a corresponding PATH_ABANDON frame, it either might not be able to remove path state or
-could use additional timers, e.g. after receiving the corresponding ACK frame,
-or other conditions like a maximum number of retransmissions to remove state.
-If path state is removed without receiving a PATH_ABANDON frame this could lead
-to packet loss. It is left to the implementation to handle this unexpected
+If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it might not be able to remove path state.
+It is left to the implementation to handle this unexpected
 behavior as it does not impact interoperability.
 
 After receiving or sending a PATH_ABANDON frame, the endpoints SHOULD


draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,15 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+This means path state removal is supposed to be triggered by the reception of a
+PATH_ABANDON frame. If a peer sends an PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it either might not be able to remove path state or
+could use additional timers, e.g. after receiving the corresponding ACK frame,
+or other conditions like a maximum number of retransmissions to remove state.
+If path state is remove without receiving a PATH_ABANDON frame this could lead
```
mirjak, line 549:
```suggestion
If path state is removed without receiving a PATH_ABANDON frame this could lead
```

qdeconinck (approved):
LGTM.

huitema (changes requested):
I do not think this paragraph is necessary. I also think it is probably wrong. Suppose that I put a path in a "locally abandoned" state, and:

* send a last ACK for the packets received so far,
* consider all the packets sent so far and not acknowledged,
* drop all incoming packets on that path without processing them.

I have at that point removed 99% of the path state -- the only resource that I need to keep is the path identifier, and I will only remove that when the ABANDON is received. But I have removed the bulk of the memory allocation: the copy of packets waiting for an ACK, the list of packets received from the peer that should be acked, probably the CID allocated for the path as well. 

Yes, that will lead to packet loss, but so what? I have sent the ABANDON already, the peer should not be using the path, if it keeps doing so, too bad. There is never a guarantee that a packet will be received, and there is no protocol violation if the packets that were dropped are not acked.


Which points that

mirjak:
@huitema looks like you comment above is half finished only...?

So what exactly is wrong about the current text?

I proposed this PR because the agreement in issue #471 that we wanted to note this issue. Please continue discussion there if you changed your mind and think we should not do anything.

draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,15 @@ both to avoid generating spurious stateless packets as specified in
...
 
+This means path state removal is supposed to be triggered by the reception of a
+PATH_ABANDON frame. If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it either might not be able to remove path state or
+could use additional timers, e.g. after receiving the corresponding ACK frame,
+or other conditions like a maximum number of retransmissions to remove state.
+If path state is removed without receiving a PATH_ABANDON frame this could lead
+to packet loss. It is left to the implementation to handle this unexpected
+behavior as it does not impact interoperability.
+
```
huitema, line 552: Less is more. I would remove extra text and just keep the following:

If a peer sends a PATH_ABANDON frame but never receives
a corresponding PATH_ABANDON frame, it ~~either~~ might not be able to remove path state ~~or
could use additional timers, e.g. after receiving the corresponding ACK frame,
or other conditions like a maximum number of retransmissions to remove state~~.
~~If path state is removed without receiving a PATH_ABANDON frame this could lead
to packet loss.~~ It is left to the implementation to handle this unexpected
behavior as it does not impact interoperability.

huitema:
The only recommendation that we should make is to not take any drastic action too soon. For example, it would be bad if application waited less than 3 PTO before closing the connection because they have not received a PATH_ABANDON yet.

draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,15 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+This means path state removal is supposed to be triggered by the reception of a
+PATH_ABANDON frame. If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it either might not be able to remove path state or
+could use additional timers, e.g. after receiving the corresponding ACK frame,
+or other conditions like a maximum number of retransmissions to remove state.
+If path state is removed without receiving a PATH_ABANDON frame this could lead
+to packet loss. It is left to the implementation to handle this unexpected
```
mirjak, line 550:
```suggestion
If a peer sends a PATH_ABANDON frame but never receives
a corresponding PATH_ABANDON frame, it might not be able to remove path state.
It is left to the implementation to handle this unexpected
```

mirjak:
I entered your proposed change which is fine for me. But based on your second comment, do you want to add anther sentence like:

"An endpoint SHOULD NOT remove path state earlier than 3 PTOs after sending the PATh_ABANDON frame".

?

huitema:
Yes, please add the warning against removing state too soon.

draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,11 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it might not be able to remove path state.
+It is left to the implementation to handle this unexpected
+behavior as it does not impact interoperability.
```
```suggestion
behavior as it does not impact interoperability. However, an endpoint SHOULD NOT
remove path state earlier than 3 PTOs after sending the PATh_ABANDON frame.
```

draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,12 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it might not be able to remove path state.
+It is left to the implementation to handle this unexpected
+behavior as it does not impact interoperability. However, an endpoint SHOULD NOT
+remove path state earlier than 3 PTOs after sending the PATh_ABANDON frame.
+
```
huitema, line 549: What state are we speaking about? I think that we are saying either too much or too little. There many kinds of "states":

1) Lists of packets that were sent on the path and not yet acked by the peer,
2) Packets that were received from the peer and processed but not yet acked,
3) the total number of paths considered active, which some implementation use to decide when to increase the Max Path ID.

Suppose that an implementation has abandoned a path, had not received a Path Abandon from the peer, and is losing patience. It may do a number of things:

1) It may decide to drop any incoming packet on that path. That has no interoperability impact, as long as those dropped packets are never acknowledged. Indeed, that's exactly what would happen if the path had become bad and packets were dropped in the network. But it has some performance impact, since some packets will be in transit when the Path Abandon is sent. It is recommended to not do that earlier than 3 PTOs after sending the PATH_ABANDON frame.

2) It may decide to consider all packets send on that path as lost, and resend the frames through another path. That's fine. It might cause some duplicate transmission, but that also happens for other reasons (e.g., bad timers) and has no interoperability impact. The rules there are murky, because this is a tradeoff -- resend sooner minimizes the end-to-end latency caused by losses, but resend sooner also increases the risk of duplicate transmission, which reduces efficiency. The 3PTO rule is probably one end of the tradeoff: good mitigation of duplicate transmission, poor management of latency. A 1PTO rule would be a different tradeoff, more focused on managing latency.

3) It should send the "missing ACKs" on another path before clearing the missing ACK state. Again, that's an efficiency tradeoff. If the node does not send acknowledgements, it risks causing spurious retransmission. 

4) It may well decide to never increase the MAX_PATH_ID until the PATH_ABANDON has been received. That's a reasonable way for an implementation to conserve resource.

huitema: I would just say:

If a peer sends a PATH_ABANDON frame but never receives
a corresponding PATH_ABANDON frame, it might not be able to remove path state.
It is left to the implementation to handle this unexpected
behavior.

huitema (commented):
I don't think we should add this paragraph. If we do, I would just keep the first sentence and the first words of the second. The alternative is a full page of text, maybe in the "implementation" section, explaining the various tradeoffs. But I have no appetite for that.

huitema (approved):

huitema:
OK, I need to explain why I said first "Yes, please add the warning against removing state too soon", and then "just keep the first sentence and the first words of the second." I did indeed change my mind. That's because most of the issues about releasing state too soon are already discussed in the draft, in particular is [3.3.1. ](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-3.3.1)[Avoiding Spurious Stateless Resets](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#name-avoiding-spurious-stateless).

The "spurious stateless reset" issue is the main reason to be cautious with unilateral removal of state. Dropping packets is kind of bad, but will not crash the connection. Dropping the CID sent to the peer before receiving the PATH_ABANDON from the peer is worse, because of the risk of generating a spurious stateless reset and thus killing the connection. 

So maybe we should say:
```
It is left to the implementation to handle this unexpected
behavior. Implementation MAY shed some of the path state and reduce
resource consumption if the PATH_ABANDON is
not arriving in a reasonable time, but they SHOULD
take steps and avoid sending spurious stateless reset (see {{spurious-stateless-reset}}).
```

draft-ietf-quic-multipath.md
mirjak: > What state are we speaking about? I think that we are saying either too much or too little. There many kinds of "states":
> 
> 1. Lists of packets that were sent on the path and not yet acked by the peer,
> 2. Packets that were received from the peer and processed but not yet acked,
> 3. the total number of paths considered active, which some implementation use to decide when to increase the Max Path ID.

Packet handling as mentioned in point 1 and 2 seems actually partly orthogonal here. If you have send the path abandon frame, you should probably not have any outstanding packets that need to be acked on that path. However, you should also never just forgot about outstanting/non-acked packets. If you decide to remove state about non-ack packets, you need to declare them lost and resent on another path. I guess if you really want to sent a path abandon frame even though you have outstanding packets, you should probably declare them lost and resent on another path right away.

About your point 2, I don't think there is an issue here and I'm not sure when this case could/would happen. If you have received packets, I guess you should first send the ACK and then the path abandon. Of course you can also send ACKs on other paths still.

The important part is about packet handling of new incoming packets and you should be able to process those for at least 3 PTOs after sending the path abandon frame. This means, you need to remember the path ID and corresponding CID(s) as well as the packet number space state. However, as soon as you remove the CID(s), you cannot process any new incoming packets anymore and as such you can then also remove all packet number space state as well for received packets. For out-going packets, it's really on you if you remove state even though there are ACKs missing; however, as I said above that means you have to retransmit on another paths.

The interesting case is actually your point 3. If you remove path state that means you remove the path ID and will consider it as not active anymore. As such you could send a new max path ID. However, as the other end is not sending you a path abandon frame it will consider the path ID as active and not increase its max path ID. Respectively, you might not be able to open a new path. But I guess that's what we have the path blocked frame for. So I guess there is no problem either way.

So should we maybe say this to be more clear:

> However, an endpoint SHOULD process incoming packets for the abandoned Path ID and corresponding connection IDs for at least 3 PTOs after sending the PATh_ABANDON frame.

draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,12 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it might not be able to remove path state.
+It is left to the implementation to handle this unexpected
+behavior as it does not impact interoperability. However, an endpoint SHOULD NOT
+remove path state earlier than 3 PTOs after sending the PATh_ABANDON frame.
```
mirjak, line 548:
```suggestion
behavior.  Implementation MAY shed some of the path state and reduce
resource consumption if the PATH_ABANDON is
not arriving in a reasonable time, but they SHOULD
take steps and avoid sending spurious stateless reset (see {{spurious-stateless-reset}}).
```

mirjak:
I'm not sure about your proposed text above, Christian. First it's very vague saying "shed _some_ path state and reduce resource consumption" as well as "in a reasonable time". However, I don't think it addresses the issue. If you send a path_abandon frame you immediately retire all CIDs that you received from the other end which means you can't not sent on the path anymore. That avoids connection closure of spurious reset, as also mentioned in the paragraph just before his new proposed text, and it exactly what path_abandon stand for - it a clear promise that you won't sent on the path anymore.

The problem we are trying to address is about retiring the CIDs that you provided to the peer. Because those you are expected to only retire when you receive a path_abandon from the other end. And here I think we should be clear that there is also a recommend minimal time of 3 PTO before retiring the CID that were issues to the other end. But I think we could be more concrete in the proposed text and not talk about "path state" but more specifically about "issued CIDs". Would that help (see proposed commit)?

draft-ietf-quic-multipath.md
mirjak, line 548:
```suggestion
remove issued CIDs earlier than 3 PTOs after sending the PATh_ABANDON frame.
```

huitema: There is no way to "(retire) the CIDs that you provided to the peer." Only the peer can do that, either by sending individual RETIRE_CONNECTION_ID frames or by sending a PATH_ABANDON frame.

Of course, you can decide to unilaterally "forget" these CIDs. The practical effect is that if the peer uses them to send packets, the forgotten CIDs will trigger a stateless reset, thus closing the connection. If you want to do that, you can just as well close the connection with some appropriate error code. A frank closure of the connection seems better than a passive-aggressive booby trap...

Maybe we should define an error code for that scenario. "I waited for the path abandon for more that 3 PTOs, and did not receive the expected path abandon, so bye. Try to fix your bugs next time."

huitema: Actually, you could send a "NEW CONNECTION ID" with a "retire prior", but that's bad because you have abandoned the path already? Plus, the new connection ID will have the path ID of the abandoned path, so that does not let you abandon the path ID.

mirjak, line 548:
```suggestion
behavior as it does not impact interoperability. If the endpoint is no longer 
willing to process the issued connection IDs for the abandoned path,
it MAY close the connection, but SHOULD wait at least 3 PTOs after
sending the PATh_ABANDON frame.
```

mirjak: Actually RFC9000 says:

> "An endpoint might need to stop accepting previously issued connection IDs in certain circumstances. Such an endpoint can cause its peer to retire connection IDs by sending a NEW_CONNECTION_ID frame with an increased Retire Prior To field. The endpoint SHOULD continue to accept the previously issued connection IDs until they are retired by the peer. If the endpoint can no longer process the indicated connection IDs, it MAY close the connection."

So effectively even a "NEW CONNECTION ID" with a "retire prior" will not retire them. And given the text above I guess closing the connection is the right/only option. Therefore see proposed new text aligning with language in RFC9000.

I'm still proposing this PR because I still think we should spell out normatively that you should wait at least 3 PTO before doing anything. Given this is not enforceable I would stick with a SHOULD but effectively this is meant to be a MUST.

huitema: Yes, that works. It avoids saying creative things, and it sends the right message to developers: reply by ABANDON to ABANDON, or risk losing the connection. Note the typo "PATh_ABANDON".

Yanmei-Liu: I believe it works.

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
qdeconinck: Yes, this works for me.

qdeconinck (approved):

draft-ietf-quic-multipath.md
```diff
@@ -541,6 +541,14 @@ both to avoid generating spurious stateless packets as specified in
 {{spurious-stateless-reset}} and to be able to acknowledge the
 last packets received from the peer as specified in {{ack-after-abandon}}.
 
+If a peer sends a PATH_ABANDON frame but never receives
+a corresponding PATH_ABANDON frame, it might not be able to remove path state.
+It is left to the implementation to handle this unexpected
+behavior as it does not impact interoperability. If the endpoint is no longer
```
```suggestion
behavior as it does not impact interoperability. If the endpoint is no longer
```

PR #477, opened 2024-12-20
always recommend to send path_abandon on another path

mirjak:
I created this PR based on the discussion in (the now-closed) PR #473.

This is a slight change of the normative recommendation as it generally recommends to send the PATH_ABANDON on another path.

I understood that this is what we want to recommend but if not, we cannot also not apply this PR. Please comment!

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -551,10 +551,10 @@ After a path is abandoned, the Path ID MUST NOT be reused
 for new paths, as the Path ID is part of the nonce calculation {{multipath-aead}}.
 
 PATH_ABANDON frames can be sent on any path,
-not only the path that is intended to be closed. Thus,
-even if connectivity on that path is already broken
-but there is still another usable path, it is RECOMMENDED to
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -551,10 +551,10 @@ After a path is abandoned, the Path ID MUST NOT be reused
 for new paths, as the Path ID is part of the nonce calculation {{multipath-aead}}.
 
 PATH_ABANDON frames can be sent on any path,
-not only the path that is intended to be closed. Thus,
-even if connectivity on that path is already broken
-but there is still another usable path, it is RECOMMENDED to
-send the PATH_ABANDON frames on another path.
+not only the path that is intended to be closed.
```
mirjak, line 554:
```suggestion
not only the path that is intended to be closed.
```

huitema (approved):
Looks good.

Yanmei-Liu (approved):
Looks good to me.

qdeconinck (approved):

Issue #454, opened 2024-10-30
Cross-path acknowledgments should be congestion controlled


michael-eriksson:
QUIC packets that only contain an ACK frame are not congestion controlled. [RFC 9000 states](https://www.rfc-editor.org/rfc/rfc9000.html#section-13.2.1-3) that:

> _Since packets containing only ACK frames are not congestion controlled, an endpoint MUST NOT send more than one such packet in response to receiving an ack-eliciting packet._

When ACK-only packets are sent on another path than where the ack-eliciting packets were received, this requirement is obviously violated. A receiver of a big flow of packets could easily swamp a low-latency, low-bandwidth path by sending the acknowledgments there.

michael-eriksson:
Christian wrote:
> The requirement in RFC 9000 is not about congestion control. It is about not having an infinite cycle of sending an ACK, receiving an ACK in return, ACKing that, etc.

Hmm, I think that the quote from RFC 9000 very explicitly mentions congestion control...

But I agree that congestion controlled ACKs can easily lead to infinite ack cycles.

huitema:
The sentence that you quoted does mention congestion control. The discussion of infinite loops happens a couple paragraphs later -- _An endpoint that is only sending non-ack-eliciting packets might choose to occasionally add an ack-eliciting frame to those packets to ensure that it receives an acknowledgment; see [Section 13.2.4](https://www.rfc-editor.org/rfc/rfc9000.html#ack-tracking). In that case, an endpoint MUST NOT send an ack-eliciting frame in all packets that would otherwise be non-ack-eliciting, to avoid an infinite feedback loop of acknowledgments._

The spec does mention congestion control, as you said: _Since packets containing only ACK frames are not congestion controlled, an endpoint MUST NOT send more than one such packet in response to receiving an ack-eliciting packet_. That mean that if a node receives just one packet, it should send at most one ACK for that specific packet. It also implies that if you receive 2 or 3 packets, it is fine to send 2 or 3 ACKs. For example, if a node receives packets 1 and 2, it would be perfectly fine to send ACK 1-2 on path 1, and a copy of ACK 1-2 on path 2. Which would not address our problem, "avoid saturating a narrow return path".

The solution really is to send fewer ACKs overall, and that means adapting the ACK FREQUENCY draft for multipath.

mirjak:
I agree that the problem discussed is not specific to a specific path and as such not specific to the multipath extension. I proposed to close the issue with no action.

[...]

PR #478, opened 2024-12-20
acks are not congestion control -> be careful

mirjak:
this PR is a clarification based on the discussion in issue #454. Of course it's optimal to add this recommendation or not. However, it's in the implementation considerations section and I think at least noting the issue is a good thing.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -979,7 +979,10 @@ faster if the scheduling strategy is stable, but besides that
 implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
-and thus better performance.
+and thus better performance. As packets that only carries PATH_ACK frames
+are not congestion controlled, sending only PATH_ACK frames on a path
+should carefully consider the load induced by these packets, especially
+if the capacity is unknown on that path.
 
 ## Retransmissions
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -980,7 +980,7 @@ implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
 and thus better performance. However, since packets that only carry PATH_ACK frames
-are not congestion controlled, senders should carefully consider the load induced
+are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}), senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
 ## Retransmissions
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -980,7 +980,8 @@ implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
 and thus better performance. However, since packets that only carry PATH_ACK frames
-are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}), senders should carefully consider the load induced
+are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
+senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
 ## Retransmissions
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -979,10 +979,9 @@ faster if the scheduling strategy is stable, but besides that
 implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
-and thus better performance. As packets that only carries PATH_ACK frames
-are not congestion controlled, sending only PATH_ACK frames on a path
-should carefully consider the load induced by these packets, especially
-if the capacity is unknown on that path.
+and thus better performance. However, since packets that only carries PATH_ACK frames
+are not congestion controlled, senders should carefully consider the load induced
+by these packets, especially if the capacity is unknown on that path.
 
 ## Retransmissions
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -979,7 +979,10 @@ faster if the scheduling strategy is stable, but besides that
 implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
-and thus better performance.
+and thus better performance. As packets that only carries PATH_ACK frames
+are not congestion controlled, sending only PATH_ACK frames on a path
+should carefully consider the load induced by these packets, especially
+if the capacity is unknown on that path.
```
huitema, line 986: The additional sentence uses the passive form and is a bit hard to read. If we want something, I would suggest:

However, since packets that only carries PATH_ACK frames
are not congestion controlled, senders should carefully consider the load induced by these packets, especially
if the capacity is unknown on that path.

huitema (commented):
I still think this is not needed. I think that we should expurgate that kind of text from the draft, and place it is a separate informational document, but if you really want to add it, I wont stand in the the way. In that case, I suggest a minor rewrite.

draft-ietf-quic-multipath.md
```diff
@@ -979,7 +979,10 @@ faster if the scheduling strategy is stable, but besides that
...
```
mirjak, line 985:
```suggestion
and thus better performance. However, since packets that only carries PATH_ACK frames
are not congestion controlled, senders should carefully consider the load induced
by these packets, especially if the capacity is unknown on that path.
```

draft-ietf-quic-multipath.md
```diff
@@ -979,7 +979,9 @@ faster if the scheduling strategy is stable, but besides that
 implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
-and thus better performance.
+and thus better performance. However, since packets that only carries PATH_ACK frames
+are not congestion controlled, senders should carefully consider the load induced
```
Yanmei-Liu, line 983:
```suggestion
are not congestion controlled(following {{Section 7 of QUIC-RECOVERY}}), senders should carefully consider the load induced
```

Yanmei-Liu: I think it's contained in RFC 9002 Section 7:
> Similar to TCP, packets containing only ACK frames do not count toward bytes in flight and are not congestion controlled.

But as PATH_ACK is a new frame added by multipath extension, I'm not against on adding this part to make things more clear. Just minor fix.

qdeconinck (commented):
No strong opinion here. I think it is fine in the implementation section, but this section starts being very large...

draft-ietf-quic-multipath.md
huitema: I believe that the way RFC9002 is written is wrong. It wants to say that ACK can be sent whatever the state of congestion control, but it says that by stating that "ACK packets do count as bytes in flight", which is wrong -- they create queues in the network just like other packets. Mirja's text nicely avoids that issue. I do not think that we should add a reference to RFC9002.

draft-ietf-quic-multipath.md
huitema: Typo. Should be " As packets that only carry", not "carries" (subject is plural).

huitema (approved):
Looks OK, besides on nit.

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
Yanmei-Liu: @huitema You're right about the packets only contains ACK still create queues on switching devices and cause process cost on both sides. My suggestion is about that "PATH_ACK is not count into congestion control" isn't something new added by multipath extension, but is inherit from RFC 9000 and 9002.

mirjak: Sec 7 of RFC9002 says:

"Similar to TCP, packets containing only ACK frames do not count toward bytes in flight and are not congestion controlled."

Further the next sentence in RFC9002 is actually:

"Unlike TCP, QUIC can detect the loss of these packets and MAY use that information to adjust the congestion controller or the rate of ACK-only packets being sent, but this document does not describe a mechanism for doing so."

So I guess it's fine to add a reference here or it might actually be useful for more context.

mirjak:
> No strong opinion here. I think it is fine in the implementation section, but this section starts being very large...

@qdeconinck the implementation section is currently 4.5 pages. It has a bunch of subsections but most of them are short. I think that is fine and I think we should more concentrate to have all the other sections concise and to the point. However, again, I'm planning to do a full editorial pass and I can imagine that some alignment in the implementation section might also make that part a bit shorter. Therefore I think we should merge for now.

qdeconinck (approved):

draft-ietf-quic-multipath.md
huitema: Not really a comment on this PR, but the part about QUIC using acknowledgments of ACK packets may be a bit optimistic. ACK-only packets do not trigger ACK. Implementation may trigger that ACK by bundling a Ping with their ACK, but they cannot do that too frequently -- typically no more than once every N ACK, with N at least 2, and no more than once per RTT. So you get very scarce feedback, and any estimation of capacity based on that is imprecise. Plus, in the general case, the ACK traffic is way lower than link capacity, so you don't get an accurate estimate of capacity -- this is an app-limited scenario.

mirjak: Yes, ACK congestion control is an open issue but that's not specific for the multipath case, as also initially discussed in issue #454. However, given we discuss in this text in the implementation guidance explicitly that ACKs could be send on another path, with is different than in the single path case, I think it is fair to add an additional warning about congestion control and point to the respective text in RFC9002 that already takes about ACK congestion control. Therefore I propose to merge this PR incl. the proposed commit with the reference to RFC9002; I believe the discussed showed there is agreement to merge.

draft-ietf-quic-multipath.md
```diff
@@ -979,7 +979,9 @@ faster if the scheduling strategy is stable, but besides that
 implementations can choose between multiple strategies such as sending
 PATH_ACK frames on the path they acknowledge packets, or sending
 PATH_ACK frames on the shortest path, which results in shorter control loops
-and thus better performance.
+and thus better performance. However, since packets that only carry PATH_ACK frames
+are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}), senders should carefully consider the load induced
+by these packets, especially if the capacity is unknown on that path.
```
mirjak, line 984:
```suggestion
are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
senders should carefully consider the load induced
by these packets, especially if the capacity is unknown on that path.
```

PR #486, opened 2025-02-04
ack-eliciting/retrans of PATHS_BLOCKED, PATH_ID_BLOCKED, MAX_PATHS

### LINKED ISSUE #485: Managing ACK and repeat of PATH_CID_BLOCKED frame

mirjak:
fixes #485

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1376,6 +1376,8 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
 a smaller Maximum Path Identifier value than was previously received.
 MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
+MAX_PATH_ID frames are ack-elicting and SHOULD be retransmitted when lost.
+
 ## PATHS_BLOCKED and PATH_CIDS_BLOCKED frames {#paths-and-cids-blocked-frame}
 
 A sender can send a PATHS_BLOCKED frame (type=0x15228c0d) when
@@ -1426,6 +1428,8 @@ Path Identifier:
 Receipt of a value of Maximum Path Identifier or Path Identifier that is higher than
 the local maximum value MUST be treated as a connection error of type PROTOCOL_VIOLATION.
 
+PATHS_BLOCKED and PATH_CIDS_BLOCKED frames are ack-eliciting and MAY be retransmitted
+if the path is still blocked when the lost is detected.
 
 # Error Codes {#error-codes}
 


huitema (approved):
One minor nit: there is no point retransmitting a frame that will be ignored by the peer.

draft-ietf-quic-multipath.md
```diff
@@ -1376,6 +1376,8 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
 a smaller Maximum Path Identifier value than was previously received.
 MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
+MAX_PATH_ID frames are ack-elicting and SHOULD be retransmitted when lost.
```
mirjak, line 1379:
```suggestion
MAX_PATH_ID frames are ack-eliciting and SHOULD be retransmitted when lost
and no more recent MAX_PATH_ID frame has been sent in the mean time.
```

mirjak:
@huitema I added another half sentence to address your comment! Please check!

huitema (approved):
Works for me. Thanks.

draft-ietf-quic-multipath.md
```diff
@@ -1376,6 +1376,9 @@ Loss or reordering can cause an endpoint to receive a MAX_PATH_ID frame with
 a smaller Maximum Path Identifier value than was previously received.
 MAX_PATH_ID frames that do not increase the path limit MUST be ignored.
 
+MAX_PATH_ID frames are ack-eliciting and SHOULD be retransmitted when lost
+and no more recent MAX_PATH_ID frame has been sent in the mean time.
```
qdeconinck, line 1380:
```suggestion
and no more recent MAX_PATH_ID frame has been sent in the meantime.
```

qdeconinck (approved):
Minor nit.

PR #488, opened 2025-02-04
Shorter intro

mirjak:
This PR is an editorial pass of the intro. The aim is to provide a shorter, more clear text. This PR does not (really) modify the subsections of the intro, however, we could consider if sec 1.1. and maybe also the others should simply be removed...?

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -88,73 +88,41 @@ enable the simultaneous usage of multiple paths for a single connection.
 
 This document specifies an extension to QUIC version 1 {{QUIC-TRANSPORT}}
 to enable the simultaneous usage of multiple paths for a single
-connection. This contrasts with
-the base QUIC protocol {{QUIC-TRANSPORT}} that includes a connection migration mechanism that
-selects only one path at a time to exchange such packets.
+connection, using the same or different 4-tuples. 
 
-The path management specified in {{Section 9 of QUIC-TRANSPORT}}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -88,74 +88,42 @@ enable the simultaneous usage of multiple paths for a single connection.
 
 This document specifies an extension to QUIC version 1 {{QUIC-TRANSPORT}}
 to enable the simultaneous usage of multiple paths for a single
-connection. This contrasts with
-the base QUIC protocol {{QUIC-TRANSPORT}} that includes a connection migration mechanism that
-selects only one path at a time to exchange such packets.
+connection, using the same or different 4-tuples.
```
mirjak, line 91:
```suggestion
connection, using the same or different 4-tuples.
```

mirjak:
Actually we wouldn't be able to just remove section 1.2 (path ID) and 1.3 entirely (multiple packet spaces), however, we could move the needed text somewhere else.

qdeconinck (approved):
Having compared side-by-side the current version and this PR, I think this text nicely summarises the previous text without losing any relevant information. It also now clearly states that paths are created by the client, which was somewhat unclear in the previous draft. I see that the introduction assumes the proposed ToC of #489 -- which is fine for me.

mirjak:
Ups I referenced wrongly this PR instead of issue #489 in PR #503 that I just merged. Re-opening...

PR #493, opened 2025-02-24
Error code for receiving MP specific frame in other than 1-RTT packet

jalume:
Aligning error code with RFC 9000. 
RFC 9000 states: "An endpoint MUST treat receipt of a frame in a packet type that is not permitted as a connection error of type PROTOCOL_VIOLATION."


mirjak (approved):
Yes, I believe this is right to just use PROTOCOL_VIOLATION instead

huitema (approved):

qdeconinck (approved):

PR #494, opened 2025-02-24
Editorial proposals

jalume:
I have some editorial proposals.

1. Introduction
'packets' have not been mentioned before. --> 'QUIC packets' instead of 'such packets'

3. Path Management
No other way to start using multiple paths was presented before. --> Removing 'also'

3.3.2. Handling PATH_ACK for abandoned paths
As far as I understand the 'other packets' are a subset of the unacknowledged packets. --> 'Some of these' instead of 'Some other'


qdeconinck (approved):
Overall good for me, but this would conflict with #488.

qdeconinck:
Alright, let's merge then.

PR #497, opened 2025-02-28
Clarify use of PATH_CIDS_BLOCKED

### LINKED ISSUE #496: Provide more guidance when to send PATH_CID_BLOCKED frame
We do specify the PATH_CID_BLOCKED in the frame section but we actually don't mention it at all in the CID management section. We should provide a bit more guidance when/how to actually use this frame.

mirjak:
fixes #496


qdeconinck (approved):
Sounds good.

PR #501, opened 2025-03-01
recommend one CID for each unused path

mirjak:
alternative to #495

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -657,8 +657,9 @@ Respectively, the connection IDs used during the handshake belong to the initial
 with Path ID 0.
 The PATH_NEW_CONNECTION_ID frame is used to issue new connection IDs for all paths.
 In order to let the peer open new paths, it is RECOMMENDED to proactively
-issue at least one Connection ID for each unused Path ID up to the peer's
-Maximum Path ID limit.
+issue at least one Connection ID for each unused Path ID up to the
+minimum of the peer's and the local
+Maximum Path ID limits.
 
 When issuing path-specific connection IDs, an endpoint associates a sequence number
 as specified in {{Section 5.1.1 of QUIC-TRANSPORT}}. Each Path ID has its own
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -657,8 +657,8 @@ Respectively, the connection IDs used during the handshake belong to the initial
 with Path ID 0.
 The PATH_NEW_CONNECTION_ID frame is used to issue new connection IDs for all paths.
 In order to let the peer open new paths, it is RECOMMENDED to proactively
-issue a Connection ID for at least one unused Path ID, as long as it remains
-compatible with the peer's Maximum Path ID limit.
+issue at least one Connection ID for each unused Path ID up to the peer's
+Maximum Path ID limit.
```
huitema, line 662: Should be:

```
up to the minimum of the peer's and the local Maximum Path ID limits.
```

This reinforces the desired behavior: do not increase the path ID limit if you are not going to make the paths usable.

huitema (approved):
Yes, that's the right way to say it. But we need to mention the local path ID limit.

draft-ietf-quic-multipath.md
```diff
@@ -657,8 +657,8 @@ Respectively, the connection IDs used during the handshake belong to the initial
 with Path ID 0.
 The PATH_NEW_CONNECTION_ID frame is used to issue new connection IDs for all paths.
 In order to let the peer open new paths, it is RECOMMENDED to proactively
-issue a Connection ID for at least one unused Path ID, as long as it remains
-compatible with the peer's Maximum Path ID limit.
+issue at least one Connection ID for each unused Path ID up to the peer's
```
mirjak, line 660:
```suggestion
issue at least one Connection ID for each unused Path ID up to the
minimum of the peer's and the local
```

draft-ietf-quic-multipath.md
```diff
@@ -657,8 +657,8 @@ Respectively, the connection IDs used during the handshake belong to the initial
...
+Maximum Path ID limit.
```
mirjak, line 661:
```suggestion
Maximum Path ID limits.
```

draft-ietf-quic-multipath.md
mirjak: Done.

Yanmei-Liu (approved):
Looks good.

qdeconinck (approved):

PR #504, opened 2025-03-12
Remove design consideration section

mirjak:
I don't think we will need this section for the final RFC, do we?


huitema (commented):
Good. Anything that removes redundant text is good. The design is the design. Text like that will fit better in a blog or in a research paper.

huitema (approved):

Yanmei-Liu (approved):
Looks good!

PR #508, opened 2025-03-16
shorter error code definition

mirjak:
I also wonder if we should move this section into a subsection of the Abandon frame because this is only for this frame...?


qdeconinck (approved):
Shorter is better.

Yanmei-Liu (approved):

PR #509, opened 2025-03-16
initial pass - (re)move subsections in impl guidance sec

mirjak:
This is not a full path on the implementation guidance section but it removes/merges one subsection and moves some sections around to align content better. However, there might be some editorial improvement needed on the rest of the text still.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1197,7 +1197,7 @@ nor the packet number space.
 
 While endpoints assign a connection ID to a specific sending 4-tuple,
 networks events such as NAT rebinding may make the packet's receiver
-observe a different 4-tuple. Though it changes the 4-tuple of the path,
+observe a different 4-tuple. This changes the 4-tuple of the path but
 the connection ID remains the same, and so the related path identifier.
 Servers observing a 4-tuple change will
 perform path validation (see {{Section 9 of QUIC-TRANSPORT}}).
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1184,7 +1184,7 @@ This section provides informational guidance for implementors.
 
 ## CID Changes, Migration, and NAT Rebindings {#migration}
 
-With the multipath extension each
+With the multipath extension, each
 path uses a separate packet number space.
 This is a major difference from
 {{QUIC-TRANSPORT}}, which only defines three number spaces (Initial,
@@ -1193,12 +1193,12 @@ Handshake and Application packets).
 The relation between packet number spaces and paths is fixed.
 Connection IDs are separately allocated for each Path ID.
 Rotating the connection ID on a path does not change the Path ID
-nor packet number space.
+nor the packet number space.
 
 While endpoints assign a connection ID to a specific sending 4-tuple,
 networks events such as NAT rebinding may make the packet's receiver
 observe a different 4-tuple. Though it changes the 4-tuple of the path,
-also does not change connection ID and as such also not the path identifier.
+the connection ID remains the same, and so the related path identifier.
 Servers observing a 4-tuple change will
 perform path validation (see {{Section 9 of QUIC-TRANSPORT}}).
 If path validation process succeeds, the endpoints set
@@ -1215,7 +1215,7 @@ More generally, while migration cannot be avoided in case of network-based
 NAT rebindings, opening a new path instead of active client migration
 should be strongly preferred when the multipath extension is supported.
 This enables a smoother handover and allows a simplified migration
-handling at the server as NAT rebindings imply immediate loss of the old
+handling at the server side as NAT rebindings imply immediate loss of the old
 address.
 
 ## Using multiple paths on the same 4-tuple
@@ -1235,7 +1235,7 @@ migrations. For example:
 
 * Client starts path 1 from address 192.0.2.1 to server address 198.51.100.1
 * Client starts path 2 from address 192.0.2.2 to server address 198.51.100.1
-* both paths are used for a while.
+* Both paths are used for a while.
 * Server sends packet from address 198.51.100.1 to client address 192.0.2.1, with CID indicating path=2.
 * Client receives packet, recognizes a path migration, update source address of path 2 to 192.0.2.1.
 
@@ -1272,7 +1272,7 @@ control schemes have been proposed for Multipath TCP such as {{OLIA}}.
 
 {{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
 can change the connection ID it uses to another available one
-at any time during the connection. As such a sole change of the Connection
+at any time during the connection. As such, a sole change of the Connection
 ID without any change in the address does not indicate a path change and
 the endpoint can keep the same congestion control and RTT measurement state.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1180,9 +1180,9 @@ but has not provided a corresponding connection ID for the path ID
 
 # Implementation Considerations {#impl-consideration}
 
-## Number Spaces
+## CID Changes, Migration, and NAT Rebindings {#migration}
 
-As stated in {{introduction}}, when the multipath extension is negotiated, each
+With the multipath extension each
 path uses a separate packet number space.
 This is a major difference from
 {{QUIC-TRANSPORT}}, which only defines three number spaces (Initial,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,77 @@ but has not provided a corresponding connection ID for the path ID
 
 # Implementation Considerations {#impl-consideration}
```
mirjak, line 1182:
```suggestion

This section provides informational guidance for implementors.
```

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,78 @@ but has not provided a corresponding connection ID for the path ID
...
 
-## Number Spaces
+This section provides informational guidance for implementors.
```
mirjak, line 1183:
```suggestion
This section provides informational guidance for implementors.

```

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
-also does not change the path identifier.
-The packet number space does not change when connection ID
-rotation happens within a given Path ID.
+Rotating the connection ID on a path does not change the Path ID
+nor packet number space.
+
+While endpoints assign a connection ID to a specific sending 4-tuple,
+networks events such as NAT rebinding may make the packet's receiver
+observe a different 4-tuple. Though it changes the 4-tuple of the path,
+also does not change connection ID and as such also not the path identifier.
```
qdeconinck, line 1201: This sentence is hard to parse, but I think that I understand the underlying meaning

```suggestion
the connection ID remains the same, and so the related path identifier.
```

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
+a peer address if that address has been seen recently. However, when the
+multipath extension is used and an endpoint has multiple addresses that
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead.
+
+More generally, while migration cannot be avoided in case of network-based
+NAT rebindings, opening a new path instead of active client migration
+should be strongly preferred when the multipath extension is supported.
+This enables a smoother handover and allows a simplified migration
+handling at the server as NAT rebindings imply immediate loss of the old
```
qdeconinck, line 1218:
```suggestion
handling at the server side as NAT rebindings imply immediate loss of the old
```

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
+There may be cases in which paths are created with different 4-tuples,
+but end up using the same 4-tuples as a consequence of path
+migrations. For example:
 
-Data associated with the transmission and reception such RTT measurements,
-congestion control state, or loss recovery are maintained per packet number
-space and as such per Path ID.
+* Client starts path 1 from address 192.0.2.1 to server address 198.51.100.1
+* Client starts path 2 from address 192.0.2.2 to server address 198.51.100.1
+* both paths are used for a while.
```
qdeconinck, line 1238:
```suggestion
* Both paths are used for a while.
```

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
+* Server sends packet from address 198.51.100.1 to client address 192.0.2.1, with CID indicating path=2.
+* Client receives packet, recognizes a path migration, update source address of path 2 to 192.0.2.1.
```
qdeconinck, line 1240: While I agree with the provided example, it implicitly makes the assumption that both paths use the same source and destination ports; otherwise, it would not lead to the same 4-tuple.

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
+
+Such unintentional use of the same 4-tuple on different paths ought to
+be rare. When they happen, the two paths would be redundant, and the
+endpoint will want to close one of them.
```
qdeconinck, line 1244:
```suggestion
endpoint may want to close one of them.
```

draft-ietf-quic-multipath.md
```diff
@@ -1222,6 +1270,12 @@ specified in {{RFC6356}} to solve this problem.  This scheme can
 immediately be adapted to Multipath QUIC. Other coupled congestion
 control schemes have been proposed for Multipath TCP such as {{OLIA}}.
 
+{{Section 5.1.2 of QUIC-TRANSPORT}} indicates that an endpoint
+can change the connection ID it uses to another available one
+at any time during the connection. As such a sole change of the Connection
```
qdeconinck, line 1275:
```suggestion
at any time during the connection. As such, a sole change of the Connection
```

qdeconinck (approved):
Minor typos and maybe a need to explicitly state assumptions in an example, but for the rest LGTM.

draft-ietf-quic-multipath.md
mirjak: I also was close to change this but if I remember correctly I think when we created this text we wanted to use anything close to normative language here. But I agree it does not read well.

draft-ietf-quic-multipath.md
mirjak: Yes, should we add that or is it okay to just not say it explicitly?

draft-ietf-quic-multipath.md
qdeconinck: In that case, why not "could" instead of "may"?

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead.
+
+More generally, while migration cannot be avoided in case of network-based
+NAT rebindings, opening a new path instead of active client migration
+should be strongly preferred when the multipath extension is supported.
+This enables a smoother handover and allows a simplified migration
+handling at the server side as NAT rebindings imply immediate loss of the old
+address.
+
```
huitema, line 1220: OK, now I kind of understand what you wanted to say before. But we end up with a mix of considerations for NAT
rebinding and path migration. That section is a bit weird, because it is unclear whether it applies to the initiator of the migration or to the peer. If we want to prevent guidance, we could say:

For any given path, we may expect to observe the Connection ID rotation, NAT Rebinding and Migration
as specified by RFC 9000, although happening in a single path. These events do not change the path ID,
and do not affect the packet number space associated with the path. As a general rule, implementations
should strongly prefer using multipath mechanisms such as creating a new path to the path migration
mechanisms of RFC 9000. This enables a smoother handover and allows a simplified migration
handling at the server side as NAT rebindings imply immediate loss of the old
address. However, if the server requests migration to a "preferred address", clients must initiate an
RFC9000-style migration of path 0 to that preferred address.

Rotating the connection ID on a path does not change the Path ID
nor the packet number space.

While endpoints assign a connection ID to a specific sending 4-tuple,
networks events such as NAT rebinding may make the packet's receiver
observe a different 4-tuple. Though it changes the 4-tuple of the path,
the connection ID remains the same, and so the related path identifier.

Servers observing a 4-tuple change will
perform path validation (see {{Section 9 of QUIC-TRANSPORT}}).
If path validation process succeeds, the endpoints set
the path's congestion controller and round-trip time
estimator according to {{Section 9.4 of QUIC-TRANSPORT}}.

huitema (commented):
I left quite a few comments. This needs some work.

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
-NAT rebinding, though it changes the 4-tuple of the path,
-also does not change the path identifier.
-The packet number space does not change when connection ID
-rotation happens within a given Path ID.
+Rotating the connection ID on a path does not change the Path ID
+nor the packet number space.
+
+While endpoints assign a connection ID to a specific sending 4-tuple,
+networks events such as NAT rebinding may make the packet's receiver
+observe a different 4-tuple. Though it changes the 4-tuple of the path,
```
mirjak, line 1200:
```suggestion
observe a different 4-tuple. This changes the 4-tuple of the path but
```

draft-ietf-quic-multipath.md
```diff
@@ -1180,31 +1180,79 @@ but has not provided a corresponding connection ID for the path ID
...
 
-Data associated with the transmission and reception such RTT measurements,
-congestion control state, or loss recovery are maintained per packet number
-space and as such per Path ID.
+{{Section 9.3 of QUIC-TRANSPORT}} allows an endpoint to skip validation of
+a peer address if that address has been seen recently. However, when the
+multipath extension is used and an endpoint has multiple addresses that
+could lead to switching between different paths, it should rather maintain
+multiple open paths instead.
+
```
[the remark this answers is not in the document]
mirjak: This PR is a bit messy as I removed most of the section on "Number Spaces" (as we do need an explicitly section in implementation guidance anymore as this is full integrated in the main spec) and then moved up the section on "Connection ID Changes" and "Migration and NAT Rebindings" (see below for the removed parts) and merged them together with the remaining parts of the "Number Spaces section. These three paragraphs I only copied over from below and did not change. Note I also did not write any new text.

I agree that this need some clean up but I wanted to do the merging first. Would it be okay with you @huitema if I merge this PR now as it is and then start a separate PR with you proposed text on this section below?

draft-ietf-quic-multipath.md
qdeconinck: I would say, rather be explicit than implicit. This is not blocking though.

PR #511, opened 2025-03-17
clarification about when to send path-abandon

### LINKED ISSUE #505: Abandon path, 3*PTO and idle timer
Reading the draft-13 more closely, I see a bit of a contradiction between 3.3.3 and 3.4.

Section 3.3.3 on "idle timeout" says: "_An endpoint can decide to close a path at any time, whether the path is in active use or not, by sending a PATH_ABANDON frame.  It is not required to send a PATH_ABANDON frame at any specific point in time.  For example, an endpoint may wait until it will anyway send another frame._"

Section 3.4 on "Path Close" says: "_If a peer sends a PATH_ABANDON frame but never receives a corresponding PATH_ABANDON frame, it might not be able to remove path state. It is left to the implementation to handle this unexpected behavior as it does not impact interoperability. If the endpoint is no longer willing to process the issued connection IDs for the abandoned path, it MAY close the connection, but SHOULD wait at least 3 PTOs after sending the PATH_ABANDON frame._"

The text in 3.3.3 is motivated in part by radio management. On a mobile device, sending any packet requires powering up the radio, which consumes energy. It is better to power the radio once and send a few packets in a batch than to power the radio multiple time. But if the endpoint waits until "something else has to be sent", it may wait more than 3*PTO, and the peer MAY close the connection. Not good.

mirjak:
fixes #505

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1422,7 +1422,8 @@ in {{QUIC-TRANSPORT}}.
 
 This document does not specify per-path idle timeouts. An endpoint
 can decide to close a path at any time, whether the path is in active
-use or not, by sending a PATH_ABANDON frame. It is not required
+use or not, by sending a PATH_ABANDON frame. For the initial PATH_ABANDON
+frame to close a path, it is not required
 to send a PATH_ABANDON frame at any specific point in time.
 For example, an endpoint may wait until it will anyway send another frame.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1422,10 +1422,10 @@ in {{QUIC-TRANSPORT}}.
 
 This document does not specify per-path idle timeouts. An endpoint
 can decide to close a path at any time, whether the path is in active
-use or not, by sending a PATH_ABANDON frame. For the initial PATH_ABANDON
-frame to close a path, it is not required
-to send a PATH_ABANDON frame at any specific point in time.
-For example, an endpoint may wait until it will anyway send another frame.
+use or not. For example, an endpoint may wait to send
+the initial PATH_ABANDON frame until it anyway sends another frame.
+Note that the receiver of an initial PATH_ABANDON frame is, however,
+required to immediately reply (see {{path-close}}).
 
 If a path is not actively used for a while, it might not be usable anymore,
 e.g. due to middlebox timeouts. To avoid such path breakage, endpoints


mirjak:
I think this is ready to merge as we agree to the text in the issue.

draft-ietf-quic-multipath.md
```diff
@@ -1422,7 +1422,8 @@ in {{QUIC-TRANSPORT}}.
 
 This document does not specify per-path idle timeouts. An endpoint
 can decide to close a path at any time, whether the path is in active
-use or not, by sending a PATH_ABANDON frame. It is not required
+use or not, by sending a PATH_ABANDON frame. For the initial PATH_ABANDON
+frame to close a path, it is not required
 to send a PATH_ABANDON frame at any specific point in time.
 For example, an endpoint may wait until it will anyway send another frame.
```
qdeconinck, line 1428: Based on https://github.com/quicwg/multipath/issues/505#issuecomment-2727266283, shouldn't we explicitly state that we want endpoints to immediately react to the reception of a PATH_ABANDON? Also the normative nature of the should depends if we keep this in implementation guidance or not...

```suggestion
However, the endpoint that did not initiated the path abandonment should immediately
reply with a PATH_ABANDON frame upon reception of a PATH_ABANDON frame (see {{path-close}}).

```

qdeconinck (commented):
Based on https://github.com/quicwg/multipath/issues/505#issuecomment-2727266283, shouldn't we explicitly state that we want endpoints to immediately react to the reception of a PATH_ABANDON? Also the normative nature of the should depends if we keep this in implementation guidance or not...

draft-ietf-quic-multipath.md
mirjak: We can add this sentence if that makes it even more clearer. However why did you remove the other sentence?

qdeconinck: Oh, that was an oversight, will fix this!

draft-ietf-quic-multipath.md
```diff
@@ -1422,7 +1422,8 @@ in {{QUIC-TRANSPORT}}.
 
 This document does not specify per-path idle timeouts. An endpoint
 can decide to close a path at any time, whether the path is in active
-use or not, by sending a PATH_ABANDON frame. It is not required
+use or not, by sending a PATH_ABANDON frame. For the initial PATH_ABANDON
+frame to close a path, it is not required
 to send a PATH_ABANDON frame at any specific point in time.
```
huitema, line 1427: The sentence "For the initial PATH_ABANDON
frame to close a path, it is not required
to send a PATH_ABANDON frame at any specific point in time" is rather hard to parse. How about removing it entirely? We have already said that "An endpoint
can decide to close a path at any time."

huitema (commented):
More comments...

draft-ietf-quic-multipath.md
mirjak: I marking this comment as resolved now because you it added in somewhere in the middle and that stops me from suggesting a change. Please see the change I will suggest next and let me know that's better. Otherwise we can maybe also remove this part.

draft-ietf-quic-multipath.md
mirjak, line 1428:
```suggestion
use or not. For example, an endpoint may wait to send
the initial PATH_ABANDON frame until it anyway sends another frame.
Note that the receiver of an initial PATH_ABANDON frame is, however,
required to immediately reply (see {{path-close}}).
```

mirjak: Does that work for you @huitema and @qdeconinck ?

huitema: Works for me.

PR #515, opened 2025-03-23
editorial pass on security considerations section

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 
 ## Memory Allocation for Per-Path Resources
 
-The initial_max_path_id transport parameter and the Max Path ID field
-in the MAX_PATH_ID frame limit the number of paths an endpoint is willing
-to maintain and accordingly limit the associated path resources.
-
+The maximum Path ID limit in initial_max_path_id or MAX_PATH_ID frame
+limit the number of paths an endpoint is willing
+to maintain and accordingly thereby also limit the associated path resources.
```
huitema, line 1502: You can strike out either accordingly or thereby or both. Fewer words is better.

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
-the number of Path ID that it issues connection IDs for.
+same Path ID before an endpoint can open a path, each endpoint could also
+control the per-path resource usage by only
+issuing connection IDs for a limited number of paths. However, using
+the maximum Path ID limit in initial_max_path_id or the MAX_PATH_ID frame is preferred.
 
-Therefore, to avoid unnecessarily resource usage, that potentially could be exploited
-in a resource exhaustion attack, endpoints should allocate those additional path resource,
+To avoid unnecessarily resource usage, that potentially could be exploited
+in a resource exhaustion attack, endpoints SHOULD allocate additional path resources,
```
huitema, line 1510: Suggestion:

To avoid unnecessarily resource usage that could be exploited

Fewer words, and also no comma before "that"

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 followed to limit the amplification risk.
 
-However, while {{QUIC-TRANSPORT}} only allows the use of one path simultaneously
-and therefore only one path migration at the time should be validated,
-this extension allows for multiple open paths, that could in theory be migrated
-all at the same time, and it allows for multiple paths that could be initialized
-simultaneously. Therefore, each path could be used to further amplify an attack.
-Endpoints need to limit the number of maximum paths and might consider
+{{QUIC-TRANSPORT}} only allows the use of one path
+and as auch there can be only one path validation at a time.
```
huitema, line 1524: Suggestion:

and thus only one path validation at a time.

huitema (commented):
Looks good, but I wonder whether we are missing something. The discussion of congestion mentioned the risk to create a "thundering herd" if multiple connections use the same path as default, and the same other path as backup. If the default path breaks, we get a sudden influx of resource on the backup path. This is more a congestion control issue than a security issue, except if third parties can somehow orchestrate the scenario -- maybe if the control the default path and can guess the backup path.

draft-ietf-quic-multipath.md
mirjak, line 1502:
```suggestion
to maintain and thereby also limit the associated path resources.
```

draft-ietf-quic-multipath.md
mirjak, line 1524:
```suggestion
and as such only one path validation at a time.
```

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
-control the per-path resource usage (beyond the connection IDs) by limiting
-the number of Path ID that it issues connection IDs for.
+same Path ID before an endpoint can open a path, each endpoint could also
+control the per-path resource usage by only
+issuing connection IDs for a limited number of paths. However, using
+the maximum Path ID limit in initial_max_path_id or the MAX_PATH_ID frame is preferred.
 
-Therefore, to avoid unnecessarily resource usage, that potentially could be exploited
-in a resource exhaustion attack, endpoints should allocate those additional path resource,
+To avoid unnecessarily resource usage, that potentially could be exploited
```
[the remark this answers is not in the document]
LPardue: Can s/unnecessarily/unnecessary

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 
 
 ## Memory Allocation for Per-Path Resources
 
-The initial_max_path_id transport parameter and the Max Path ID field
-in the MAX_PATH_ID frame limit the number of paths an endpoint is willing
-to maintain and accordingly limit the associated path resources.
-
+The maximum Path ID limit in initial_max_path_id or MAX_PATH_ID frame
+limit the number of paths an endpoint is willing
```
qdeconinck, line 1501:
```suggestion
limits the number of paths an endpoint is willing
```

draft-ietf-quic-multipath.md
qdeconinck:
```suggestion
to maintain and thereby also limits the associated path resources.
```

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 successful handshake when both endpoints indicate support for this extension.
 Respectively, all new frames defined in this extension are only used in 1-RTT packets.
 As the handshake is not changed by this extension, the transport security mechanisms
 as specified in {{QUIC-TLS}}, such as encryption key exchange and peer authentication,
-remain unchanged as well and the respective security considerations in {{QUIC-TLS}} applied unaltered.
-Note that with the use of this extension, multiple nonces can be in use simultaneously
-for the same AEAD key.
-
-Further note, that the limits as discussed on Appendix B of {{QUIC-TLS}}
+remain unchanged. As such also the respective security considerations in {{QUIC-TLS}} apply unaltered. However, note that the limits as discussed on Appendix B of {{QUIC-TLS}}
```
qdeconinck, line 1541:
```suggestion
remain unchanged. As such, the respective security considerations in {{QUIC-TLS}} apply unaltered. However, note that the limits as discussed on Appendix B of {{QUIC-TLS}}
```

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 apply to the total number of packets sent on all paths.
 
 This specification changes the AEAD calculation by using the path identifier as part of
-AEAD encryption nonce (see {{multipath-aead}}). To ensure a unique nonce, path identifiers
-are limited to 32 bits and cannot be reused for another path in the same connection.
-
+AEAD nonce (see {{multipath-aead}}). To ensure a unique nonce, path identifiers
+are limited to 32 bits and cannot be reused for another path of the same connection.
+However, note that with the use of this extension, multiple nonces can be in use simultaneously
+for the same AEAD key.
```
qdeconinck, line 1548: I am not sure that this sentence is needed or even useful; a correct usage of nonces ensures this by design (i.e., no nonce reuse for the same AEAD key).

```suggestion
```

mirjak: This was added to ensure people encounter for use of multiple nonces in their implementation. Yes, it's hard to get this wrong as you have multiple packet number spaces but it seemed important enough that we wanted to mention it. If there is nothing wrong about this sentence, I propose to keep it.

qdeconinck: Alright then, though I wonder if we rather want to keep several "nonce computation states" (Path ID + PN) that leads to unique nonces for a same AEAD key.

mirjak: would be fine for me. Do you want to propose a wording?

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
```
qdeconinck, line 1548: This is a first shot, do not hesitate
```suggestion
This implies maintaining some per-path nonce computation state (a Path ID
and a packet number). Altogether, this ensures unique nonces over all the paths for
a same AEAD key.
```

mirjak: Should we even say:

"this implies maintaining per-path nonce values and computation state..."

?

mirjak: Okay, how about

"The multipath extension maintains unique per-path nonces for a same AEAD key."

That makes the state aspect a bit more clear...?

mirjak:
```suggestion
The multipath extension maintains unique per-path nonces for a same AEAD key.
```
Is that okay @qdeconinck ?

mirjak:
> We have some warning about sending ack on another path in the implementation guidance section here: https://quicwg.org/multipath/draft-ietf-quic-multipath.html#name-packet-scheduling
> 
> If you think it's useful we could also a few words to the security consideration section (with a pointer to that section). If so, please open a new issue and we can address this in a separate PR.

@huitema do you want to add one sentence with a reference to the Packet scheduling section or do you think the text in that section is sufficient (without adding anything in the security considerations)?

draft-ietf-quic-multipath.md
```diff
@@ -1487,68 +1487,65 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 followed to limit the amplification risk.
 
-However, while {{QUIC-TRANSPORT}} only allows the use of one path simultaneously
-and therefore only one path migration at the time should be validated,
-this extension allows for multiple open paths, that could in theory be migrated
-all at the same time, and it allows for multiple paths that could be initialized
-simultaneously. Therefore, each path could be used to further amplify an attack.
-Endpoints need to limit the number of maximum paths and might consider
+{{QUIC-TRANSPORT}} only allows the use of one path
+and as such only one path validation at a time.
```
huitema, line 1524: That text is a bit imprecise. RFC 9000 completely allows the peer to probe multiple address tuples at a time. This is only limited by the number of CID that are available. We have to define what we mean by validation. Validation happens when the server receives a validating packet from the client; if the 4-tuple is already validated (server received a valid response to a path challenge), the server automatically migrates to that new address. It would be simpler to state:

{{QUIC-TRANSPORT}} only allows the use of one path at a time.

huitema (commented):
Mostly OK. I may have other comments on the security section, but they would not be editorial.

huitema:
The "thundering herd" issue is explained in a note that I wrote on [QUIC and congestion control](https://privateoctopus.com/2025/03/24/multipath-congestion-control.html). Note that the issue is due to having multiple QUIC connections switching to the backup path at the same time. It is more an operation issue that a security issue -- but OK, maybe attackers could set up a wifi network, accept lots of naive customers that will have a backup path set to the target wireless network, and then stop the Wi-Fi network in order to do the backup.

If we want to discuss that, we have to do that in an "operation consideration" section, and then maybe reference that section in the security section.

draft-ietf-quic-multipath.md
mirjak: Or would be it correct to say:

"{{QUIC-TRANSPORT}} only allows the use of one path
and as such only one path migration at a time."

While you are right that multiple path validation could be on-going (if there are enough CIDs) at the end you con only switch over to one path, so I think saying "one path migration at a time" is correct, no?

mirjak:
> Mostly OK. I may have other comments on the security section, but they would not be editorial.

@huitema can you open an issue (and potentially even create a separate PR) if you have further non-editorial comments on the security considerations? Or what exactly do you mean?

draft-ietf-quic-multipath.md
qdeconinck: The only issue I have with this wording is that one may thing nonces are unique per path, which is not correct (nonces are computed in a way that two different paths will always have different values, hence guaranteeing their uniqueness over the whole connection). I fear this last version is ambiguous on this.

mirjak: Should we just remove the word "unique" then? "nonces for a same AEAD key" implies that they are unique, no?

draft-ietf-quic-multipath.md
huitema: No. The focus of the paragraph is on "address spoofing". The typical spoofed challenge / response scenario is:

1. Client sends PATH_CHALLENGE on tuple X
2. Server picks unique CID sends PATH_CHALLENGE + PATH_RESPONSE on symmetric of tuple X
3. Client does not reply because the source address was spoofed.
4. Server repeats PATH_CHALLENGE + PATH_RESPONSE a few time, subject to amplification limits.

Damage happens if the packet sent by the server is accepted and processed by the target because it somehow mimics a valid protocol.

This can be run in parallel for multiple tuples in both RFC9000 and multipath. In RFC9000, the protocol limit is the number of available CID at the server. With multipath, it is limited by the sum of CID available on all paths.

So the multipath risk level is pretty much the same as RFC9000, except maybe there are more CIDs available.

draft-ietf-quic-multipath.md
huitema: The method described in {{multipath-aead}} defines the AEAD nonce as a the combination of the IV, the 32 bit path ID, and the 62 bit packet sequence number. The path ID is unique to the path, and a given 62 bit sequence number is only use for a single packet on that path. Combining path ID and sequence number guarantees that AEAD will use a different nonce for each packet sent on the connection.

draft-ietf-quic-multipath.md
mirjak: Okay, you are right but I think we need to fix this in a separate PR. I would like to propose this following small fit for now and then merge this PR for now and start a new (slightly more than editorial PR):
```suggestion
and the number of concurrent path validation attempts is limited by number of issued connection IDs.
```

draft-ietf-quic-multipath.md
mirjak: yes, the point was it's not only unique for the path but actually unique for the whole connection. However, the fact they are unique is not really the point of this sentence, so that's why I proposed to just remove that word here.

huitema: It boils down to the definition of "nonce". By definition, "nonce" means "only once", and thus "unique". What the AEAD specification says is a bit fuzzy, with nonce used to designate either the combination of IV+sequence number or the sequence number itself. My preference would be:

The method described in {{multipath-aead}} provides unique AEAD nonces for all packets across all path.

qdeconinck (approved):
I think this is good, except for the last sentence regarding nonces. Either we commit my suggestion (https://github.com/quicwg/multipath/pull/515#discussion_r2011684026) or we instead remove that sentence.

draft-ietf-quic-multipath.md
```diff
@@ -1386,68 +1386,66 @@ TBD-13 (experiments use 0x004e4f5f4349445f) | NO_CID_AVAILABLE | Path abandoned
...
 apply to the total number of packets sent on all paths.
 
 This specification changes the AEAD calculation by using the path identifier as part of
-AEAD encryption nonce (see {{nonce}}). To ensure a unique nonce, path identifiers
-are limited to 32 bits and cannot be reused for another path in the same connection.
-
+AEAD nonce (see {{nonce}}). To ensure a unique nonce, path identifiers
+are limited to 32 bits and cannot be reused for another path of the same connection.
+However, note that with the use of this extension, multiple nonces can be in use simultaneously
+for the same AEAD key.
```
mirjak, line 1448: This proposed change removes the last sentences but also edits the previous sentence a bit.
```suggestion
AEAD nonce (see {{nonce}}). To ensure unique nonces for simultaneous use over all paths, path identifiers
are limited to 32 bits and cannot be reused for another path of the same connection.
```
Does this work? @huitema @qdeconinck ?

huitema: I don't think that helps. Just remove the last sentence, from "However...".

huitema: But yes, do replace "a unique nonce" by "unique nonces".

mirjak:
```suggestion
AEAD nonce (see {{nonce}}). To ensure unique nonces, path identifiers
are limited to 32 bits and cannot be reused for another path of the same connection.
```

PR #518, opened 2025-03-24
new paragraph on use cases in intro

### LINKED ISSUE #512: Text about use cases in Intro
Maybe also add a bit of a warning that general purpose use needs more logic that is not specified in this draft.

mirjak:
fixes #512

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -116,14 +116,14 @@ of a path, but it has to validate a new path created by a client.
 This extension does not cover address discovery and management. Addresses
 and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
-defining mechanisms to address the remaining scenarios.
+defining mechanisms to cope with the remaining scenarios.
 
 Further, this document does not specify detailed algorithms that define
 how multiple, simultaneously open paths are used to send packets.
 As these differ depending on the application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
 
-Specifically, while failover between WiFi
+Specifically, while failover between Wi-Fi
 and mobile networks is a well-known multipath use case,
 it only uses two paths at the same time temporarily in order
 to avoid transmission pauses.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
 initiation {{path-initiation}}, handling of per-path connection IDs {{consume-retire-cid}},
 signaling of preferred path usage {{path-state}}, and explicit
 removal of paths that have been abandoned {{path-close}}.
-However, this document does not specify detailed algorithms that define
+Note that in this extention, a QUIC server does not initiate the creation
+of a path, but it has to validate a new path created by a client.
+
+This proposal does not cover address discovery and management. Addresses
+and the actual decision to setup or tear down paths are assumed
+to be handled by the application. But this document does not prevent future extensions from
+defining mechanisms to address the remaining scenarios.
+
+Further, this document does not specify detailed algorithms that define
 how multiple, simultaneously open paths are used to send packets.
 As these differ depending on the application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
 
-Further, this proposal does also not cover address discovery and management. Addresses
-and the actual decision process to setup or tear down paths are assumed
-to be handled by the application that is using the QUIC multipath
-extension. However, this document does not prevent future extensions from
-defining mechanisms to address the remaining scenarios.
-
-In this extention, a QUIC server does not initiate the creation
-of a path, but it has to validate a new path created by a client.
+Specifically, while failover between WiFi
+and mobile networks is a well-known multipath use case,
+it only uses two paths at the same time temporily in order
+to avoid transmission pauses.
+Simultaneous path usage generally, however, needs more consideration
+than specified in this document to avoid negative performance
+impacts, e.g. when stream data is distributed over multiple paths with
+different delays.
 
 ## Basic Design Points
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -110,10 +110,10 @@ This document specifies the needed path management mechanisms for path
 initiation {{path-initiation}}, handling of per-path connection IDs {{consume-retire-cid}},
 signaling of preferred path usage {{path-state}}, and explicit
 removal of paths that have been abandoned {{path-close}}.
-Note that in this extention, a QUIC server does not initiate the creation
+Note that in this extension, a QUIC server does not initiate the creation
 of a path, but it has to validate a new path created by a client.
 
-This proposal does not cover address discovery and management. Addresses
+This extension does not cover address discovery and management. Addresses
 and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to address the remaining scenarios.
@@ -125,7 +125,7 @@ only some basic implementation guidance is discussed in {{impl-consideration}}.
 
 Specifically, while failover between WiFi
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporily in order
+it only uses two paths at the same time temporarily in order
 to avoid transmission pauses.
 Simultaneous path usage generally, however, needs more consideration
 than specified in this document to avoid negative performance


gorryfair (approved):
Looks better to me, thanks.

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
...
 signaling of preferred path usage {{path-state}}, and explicit
 removal of paths that have been abandoned {{path-close}}.
-However, this document does not specify detailed algorithms that define
+Note that in this extention, a QUIC server does not initiate the creation
+of a path, but it has to validate a new path created by a client.
+
+This proposal does not cover address discovery and management. Addresses
+and the actual decision to setup or tear down paths are assumed
+to be handled by the application. But this document does not prevent future extensions from
+defining mechanisms to address the remaining scenarios.
```
LPardue, line 119: Perhaps avoid using `address` here since it means something different to the earlier address discovery mention

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
 initiation {{path-initiation}}, handling of per-path connection IDs {{consume-retire-cid}},
 signaling of preferred path usage {{path-state}}, and explicit
 removal of paths that have been abandoned {{path-close}}.
-However, this document does not specify detailed algorithms that define
+Note that in this extention, a QUIC server does not initiate the creation
```
LPardue, line 113:
```suggestion
Note that in this extension, a QUIC server does not initiate the creation
```

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
...
-and the actual decision process to setup or tear down paths are assumed
-to be handled by the application that is using the QUIC multipath
-extension. However, this document does not prevent future extensions from
-defining mechanisms to address the remaining scenarios.
-
-In this extention, a QUIC server does not initiate the creation
-of a path, but it has to validate a new path created by a client.
+Specifically, while failover between WiFi
+and mobile networks is a well-known multipath use case,
+it only uses two paths at the same time temporily in order
```
LPardue, line 128:
```suggestion
it only uses two paths at the same time temporarily in order
```

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
 initiation {{path-initiation}}, handling of per-path connection IDs {{consume-retire-cid}},
 signaling of preferred path usage {{path-state}}, and explicit
 removal of paths that have been abandoned {{path-close}}.
-However, this document does not specify detailed algorithms that define
+Note that in this extention, a QUIC server does not initiate the creation
+of a path, but it has to validate a new path created by a client.
+
+This proposal does not cover address discovery and management. Addresses
```
LPardue, line 116: Maybe?

```suggestion
This extension does not cover address discovery and management. Addresses
```

draft-ietf-quic-multipath.md
martinduke, line 113:
```suggestion
Note that in this extension, a QUIC server does not initiate the creation
```

draft-ietf-quic-multipath.md
martinduke, line 128:
```suggestion
it only uses two paths at the same time temporarily in order
```

martinduke (approved):
Thank you for this text.

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
...
-In this extention, a QUIC server does not initiate the creation
-of a path, but it has to validate a new path created by a client.
+Specifically, while failover between WiFi
+and mobile networks is a well-known multipath use case,
+it only uses two paths at the same time temporily in order
+to avoid transmission pauses.
+Simultaneous path usage generally, however, needs more consideration
+than specified in this document to avoid negative performance
+impacts, e.g. when stream data is distributed over multiple paths with
+different delays.
```
larseggert, line 133:
```suggestion
Simultaneous transmission on multiple paths is not supported by the extensions defined
in this document. It requires substantially more machinery than defined here
to avoid negative performance impacts, e.g., when data is to be transmitted 
over multiple paths with different characteristics.
```

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
...
 signaling of preferred path usage {{path-state}}, and explicit
 removal of paths that have been abandoned {{path-close}}.
-However, this document does not specify detailed algorithms that define
+Note that in this extension, a QUIC server does not initiate the creation
+of a path, but it has to validate a new path created by a client.
+
+This extension does not cover address discovery and management. Addresses
+and the actual decision to setup or tear down paths are assumed
+to be handled by the application. But this document does not prevent future extensions from
+defining mechanisms to address the remaining scenarios.
```
qdeconinck, line 119:
```suggestion
defining mechanisms to cope with the remaining scenarios.
```

draft-ietf-quic-multipath.md
```diff
@@ -110,19 +110,27 @@ This document specifies the needed path management mechanisms for path
...
 
-Further, this proposal does also not cover address discovery and management. Addresses
-and the actual decision process to setup or tear down paths are assumed
-to be handled by the application that is using the QUIC multipath
-extension. However, this document does not prevent future extensions from
-defining mechanisms to address the remaining scenarios.
-
-In this extention, a QUIC server does not initiate the creation
-of a path, but it has to validate a new path created by a client.
+Specifically, while failover between WiFi
```
qdeconinck, line 126:
```suggestion
Specifically, while failover between Wi-Fi
```

qdeconinck (approved):
Minor nits but for the rest, good to ship.

draft-ietf-quic-multipath.md
larseggert: I thought we agreed that this doc should call out that it is not a complete multipath solution at PS level. "Can experiment" to me sounds again like this should be Experimental?

mirjak: This is covered already in the previous paragraph where it does say that this extension does not "define
how multiple, simultaneously open paths are used to send packets" (and always did). However, there is nothing in this spec that stops people from using multiple paths, so your sentence is simply incorrect. Also we want to enable that; I thought we just agreed to add a warning.

gorryfair: At least a MP transport also needs to specify a scheduler, and I expect the IETF to care how that scheduler operates from a CC perspective.

The other thing I noted, was that we now have RFC 9743. 
When I spoke at the Mic, I was suggesting that it might be possible for MP-QUIC to state the requirements for the multipath scheduler that is used with the MP-DCCP spec needs to conform to the statement on "Multipath Transport", in Section 7.10 of RFC 9743. This explicitly differentiates between failover and concurrent multipath scheduling.

LPardue: From the perspective of the Internet, what is the difference between an application thst opens multiple connections and one that opens multiple paths? 

In my opinion, none.

There seems to be a desire to put requirements on this document that don't apply to usage of RFC 9000.

LPardue: > I thought we agreed that this doc should call out that it is not a complete multipath solution at PS level. "Can experiment" to me sounds again like this should be Experimental?

I don't know what the definition of "complete" is supposed to be. As I mentioned in IETF 122, we designed a multiplexed transport and shipped it as PS without any mechanism to signal prioritization or to schedule data across streams. The only guidance was in https://datatracker.ietf.org/doc/html/rfc9000#section-2.3

> Stream multiplexing can have a significant effect on application performance if resources allocated to streams are correctly prioritized.

> QUIC does not provide a mechanism for exchanging prioritization information. Instead, it relies on receiving priority information from the application.

> A QUIC implementation SHOULD provide ways in which an application can indicate the relative priority of streams. An implementation uses information provided by the application to determine how to allocate resources to active streams.

Hence, I don't think we should be holding this extension to some higher bar of expectation.

LPardue: > @gorry per-path congestion control is for QUIC already a requirement in RFC9000 because of migration. And we discuss this already in the implementation guidance in section 6.2: https://quicwg.org/multipath/draft-ietf-quic-multipath.html#name-congestion-control
> 
> As you can see that section also already discusses the problem of non-disjoint paths and references coupled congestion control.

I've proposed we add a pointer to RFC 9743 in our section 6.2 for any future CC enthusiasts that want to tackle that work - https://github.com/quicwg/multipath/pull/520

martinduke: > From the perspective of the Internet, what is the difference between an application thst opens multiple connections and one that opens multiple paths?
> 
> In my opinion, none.
> 
> There seems to be a desire to put requirements on this document that don't apply to usage of RFC 9000.

RFC 9000 does not support the delivery of application data simultaneously over multiple paths, which is at issue here. If I have dramatically misunderstood this document, and it's just a souped-up failover, then I withdraw my comments.

As always, "Multipath" is an almost useless term, as it conflates failover mechanisms (which we have been doing in proposed standards for 25 years) and concurrent delivery of application data over multiple paths, which can work well in some cases but doesn't have a general theory that accounts for different application flows and path properties.

As for the comparison to opening multiple connections, I have several responses:
1) Procedurally, we just ran 9743 through the consensus process and no one said a peep about the concurrent multipath text.

2) I would not say that opening multiple connections is exactly encouraged, although defects in TCP and HTTP<3 have incentivized this behavior. But we are not the protocol police; it is common behavior.

3) Opening multiple connections tends to, IIUC, stripe separate operations along different connections. In the general case, doing something like striping a very large file over multiple connections would be quite difficult.

4) When the paths have different properties, it is not at all clear what to do in terms of scheduling.

martinduke: > @gorryfair @martinduke any further comments from you side (as you approved the PR already as is)?

I'm good, thanks.

gorryfair: I thought I'd made a comment, I'll find out where this went and repost today.

mirjak: the only earlier comment I see from you is "Looks better to me, thanks." which I interpreted as you are fine with the proposed text and it addresses your concern...?

LPardue: > doing something like striping a very large file over multiple connections would be quite difficult.

Not all all. HTTP range requests makes this trivial. In fact, this is a common design in Amazon S3 and all of its clones, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

I think we are past the point of this PR being a siitable discussion venue. 

Chair hat on: Mirja please merge this PR once you are ready

Everyone else: if you have remaining concerns or other points to makr please open an issue with clear, actionable text.

gorryfair: As promised. What I said at the Mic was related to RFC5033.bis, now published as RFC 9743 and a part of [BCP 133].  
I was hoping that the I-D would note that realising a MP protocol include selecting an appropriate scheduling algorithm and that any new IETF-specified multipath scheduling algorithm would need to be evaluated using the methods defined in RFC 9743 (see section 7.10 on "Multipath Transport").

LPardue: I agree that's good suggestion and I've proposed text to that effect in https://github.com/quicwg/multipath/pull/520

huitema: I wrote a small note on the subject of [multipath and congestion control](https://privateoctopus.com/2025/03/24/multipath-congestion-control.html). I do not think that RFC 9473 applies to scheduling algorithms. Scheduling algorithms define which packets goes on what path, and there are very many algorithms that could be tried, such as:

* associate QUIC streams to specific paths, so as to minimize the amount of reordering between packets received on the same path (some folks do that with picoquic)
* or, spread packets of a long stream on multiple paths, based on whichever path has congestion control credits available at time of scheduling (picoquic does that by default)
* or, send a stream twice, with one copy on each path, to get the best delivery delay,
* or, use some kind of FEC, send a stream on one path and redundant FEC packets on the other, to get the best delays without being too piggish.

All of those have been experimented. All depends on following whatever limits are imposed by per path congestion control.

gorryfair: This raises an interesting question about the maturity of the method. To me the list above suggests experimental use.  I see aligns with the suggestion to revisit the document status: If the desire is to publish as EXP, then I still think that RFC 9473, is relevant to specifying future standards-track methods.

LPardue: PRs are not the appropriate place for these discussions so I've locked it down. 

Please open separate issue(s) if there are other points to follow up with.

mirjak:
I'm merging this PR as several people have approved it. As indicated by the chair, for further comments please open new issues and we can work on another PR if needed.

PR #521, opened 2025-03-28
Add paragraph on concurrent path use

mirjak:
As requested by Gorry this PR add the more or less same wording as used in the MP-DCCP document.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1274,6 +1274,12 @@ are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
 senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
+The specification of scheduling for concurrent multipath and related the
+congestion control algorithms and re-ordering methods for use in the general
+Internet are outside the scope of this document. If, and when, the IETF
+specifies a method for concurrent usage of multiple paths for the general
+Internet, it can be used with the multipath extension specified in this document.
+
 ## Retransmissions
 
 Simultaneous use of multiple paths enables different
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1274,8 +1274,14 @@ are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
 senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
+There are various ways that multipath techniques can be used (see {{Section 7.10 of rfc9743}}). 
+One example use is to provide failover from one path to another when
+the original path is no longer viable or provides inferior performance.
+Another example use is concurrent multipath, where the transport
+protocol simultaneously schedules a flow to aggregate the capacity
+across multiple paths.
 The specification of scheduling for concurrent multipath and related the
-congestion control algorithms and re-ordering methods for use in the general
+congestion control algorithms for use in the general
 Internet are outside the scope of this document. If, and when, the IETF
 specifies a method for concurrent usage of multiple paths for the general
 Internet, it can be used with the multipath extension specified in this document.


draft-ietf-quic-multipath.md
```diff
@@ -1274,6 +1274,12 @@ are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
 senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
+The specification of scheduling for concurrent multipath and related the
+congestion control algorithms and re-ordering methods for use in the general
```
LPardue, line 1278: I appreciate this text comes from almost directly from MP-DCCP. However, 

1. "concurrent mulitpath" doesn't seem to be a term or phrase this draft uses anywhere
2. I think there's mistake or bad grammar in "concurrent multipath and related the congestion control algorithms"
3. don't understand the "re-ordering methods" part - is that even relevant to this design?

qdeconinck: I agree with the above comments (especially point 3, as there should not be any packet reordering over a given path, but rather some "STREAM data" reordering, which is handled by design in (single-path) QUIC). More generally, do we need to add more than "Further scheduling considerations are implementation specific and are out of scope of this document."?

qdeconinck (commented):
I agree with the above comments (especially point 3, as there should not be any packet reordering over a given path, but rather some "STREAM data" reordering, which is handled by design in (single-path) QUIC). More generally, do we need to add more than "Further scheduling considerations are implementation specific and are out of scope of this document."?

draft-ietf-quic-multipath.md
```diff
@@ -1274,6 +1274,18 @@ are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
 senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
+There are various ways that multipath techniques can be used (see {{Section 7.10 of rfc9743}}).
```
mirjak, line 1277: Note we already added a reference to RFC9743 in section 5.3 on congestion control:

"Designers of congestion control algorithms specialized for Multipath QUIC are advised to follow BCP 133; see [Section 7.10](https://rfc-editor.org/rfc/rfc9743#section-7.10) of [[RFC9743](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#RFC9743)].[¶](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-5.3-2)"

draft-ietf-quic-multipath.md
```diff
@@ -1274,6 +1274,18 @@ are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
 senders should carefully consider the load induced
 by these packets, especially if the capacity is unknown on that path.
 
+There are various ways that multipath techniques can be used (see {{Section 7.10 of rfc9743}}). 
+One example use is to provide failover from one path to another when
+the original path is no longer viable or provides inferior performance.
+Another example use is concurrent multipath, where the transport
+protocol simultaneously schedules a flow to aggregate the capacity
+across multiple paths.
```
mirjak, line 1282: This seems to repeat text from RFC9743 and also what we already say in the introduction:

"Specifically, while failover between Wi-Fi and mobile networks is a well-known multipath use case, it only uses two paths at the same time temporarily in order to avoid transmission pauses. Simultaneous path usage generally, however, needs more consideration than specified in this document to avoid negative performance impacts, e.g. when stream data is distributed over multiple paths with different delays.[¶](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-1-9)"

I assume the intention is to introduce the term "concurrent multipath", however, I don't think this is necessary because we should rather use the term that is already introduced in this document, which is "simultaneous path usage".

draft-ietf-quic-multipath.md
```diff
@@ -1274,6 +1274,18 @@ are not congestion controlled (see {{Section 7 of QUIC-RECOVERY}}),
...
+The specification of scheduling for concurrent multipath and related the
+congestion control algorithms for use in the general
+Internet are outside the scope of this document. If, and when, the IETF
+specifies a method for concurrent usage of multiple paths for the general
+Internet, it can be used with the multipath extension specified in this document.
```
mirjak, line 1287: Do we really need this last sentence? I think this goes without saying....

mirjak (commented):
This edit tries to align the wording, however, in uses a lot of words to introduce the term "concurrent multipath" while the term used in this document is "simultaneous path usage". Let's just stick to this.

Further note that this section on is packet scheduling and there is a separate section on congestion control. I also think these two things are different enough that different considerations apply. Congestion control is similar to single path use; there are some additional considerations on fairness but that applies also to situations were applications open multiple connections on the same path, so I don't think this is a completely unknown situation.

For scheduling, however, we have less experience but as the text implies this is a general question for all multipath protocols as well as application layer multipath and I don't think it is really necessary to say that any future work is as applicable to this extension than other multipath protocols.

So the only thing left is maybe to explicitly say that scheduling was out-of scope for this doc, however, we already say this in the intro:

"Further, this document does not specify detailed algorithms that define how multiple, simultaneously open paths are used to send packets. As these differ depending on the application requirements, only some basic implementation guidance is discussed in [Section 5](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#impl-consideration).[¶](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-1-8)"

If we want to make that more explicit, I suggest to edit this sentence further up in this section:

OLD
"The scheduling is a local decision, based on the preferences of the application and the
implementation."

NEW
"Scheduling is a local decision, based on the preferences of the application and the
implementation, and is out of scope for this document."

draft-ietf-quic-multipath.md
gorryfair: Yes, but please make sure somewhere you differentiate between path failover and concurrent/simultaneous usage.

draft-ietf-quic-multipath.md
gorryfair: Yes, I really would expect that sentence to appear.

draft-ietf-quic-multipath.md
mirjak: RFC9743 does that - explaining the difference between path failover and concurrent use. Why do we need to repeat this here if we don't give any additional advise?

draft-ietf-quic-multipath.md
LPardue: This document is not about IP though. So IP packet reordering methods seem totally put of scope of discussion. There's literally nothing I can do with that statement as an implementer of this extension

gorryfair:
Your text seems to hint you could do it ,but there are more things to consider; whereas my proposed text says more like the IETF has specified one but not the other. I think we could find text that does both, andthis ought to be a change to the intro reflecting the key points.
I think using the word scheduler in the intro would be good.
On "concurrent": if your usage means that "simulatenous: is the same as concurrent (i.e. data on both paths) then you can state this - That I would appreciate.
I've decided to abandon the rest of my thread here because it would be better to start a new PR that is based on the latest text. I will of course also review the entire document when ready, and I think the final text can be consistent. Thanks.

mirjak:
After discussion with Gorry, I created PR #538 to add one more sentence to the intro text covering the missing piece of this PR. The other parts should be covered already in the intro (but using different words). Therefore closing this PR in agreement with Gorry to proceed with PR #538 instead.

PR #523, opened 2025-03-28
Edits on migration guidance section

mirjak:
Proposed edits by Christian; slightly adapted

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1086,15 +1086,13 @@ as specified in {{QUIC-TRANSPORT}} might occur, similar as on a single path.
 These events do not change the Path ID,  and do not affect the packet number
 space associated with the path.
 
-In case of network-based NAT rebindings, path migration cannot be avoided.
-However, as a general rule, when the multipath extension is supported,
-implementations  should otherwise strongly prefer opening a new path instead
-of path migration  as specified in {{QUIC-TRANSPORT}}.
-This enables a smoother handover and allows a simplified migration
-handling at the server side as NAT rebindings imply immediate loss of the old
-address. However, if the server requests migration to a "preferred address",
-clients must initiate migration of path 0 to that preferred address as
-specified in {{QUIC-TRANSPORT}}.
+It is generally preferable to use multipath mechanisms such as
+creating a new path and later abandoning the old path,
+rather than doing migration of a single path as specified in {{QUIC-TRANSPORT}}.
+This enables a smoother handover and allows a more controlled migration handling
+at the server side. However, migration of a single path cannot be
+avoided in case of NAT rebinding, or if the server requests migration
+to a "preferred address" during the handshake.
 
 {{Section 9.3 of QUIC-TRANSPORT}} allows an endpoint to skip validation of
 a peer address if that address has been seen recently. However, when the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1081,15 +1081,15 @@ This is a major difference from
...
-In case of network-based NAT rebindings path migration cannot be avoided.
-However otherwise, as a general rule, when the multipath extension is supported,
-implementations
 should strongly prefer opening a new path instead of path migration

-as specified in {{QUIC-TRANSPORT}}.
+In case of network-based NAT rebindings, path migration cannot be avoided.
+However, as a general rule, when the multipath extension is supported,
+implementations
 should otherwise strongly prefer opening a new path instead
```
huitema, line 1091: you could remove "otherwise"

draft-ietf-quic-multipath.md
```diff
@@ -1081,15 +1081,15 @@ This is a major difference from
...
+of path migration
 as specified in {{QUIC-TRANSPORT}}.
```
huitema, line 1092: It might be better to specify the rule first, and the exceptions later. Something like:

It is generally preferable to use multipath mechanisms such as creating a new path and later abandoning the old path, rather than doing migration of a single path as specified in {{QUIC-TRANSPORT}}. This enables a smoother handover and allows a more controlled migration handling at the server side. However, migration of a single path cannot be
avoided in case of NAT rebinding, or if the server requests migration to a "preferred address" during the handshake.

huitema (commented):
Kinda good, but we could improve the "prefer multipath" paragraph.

draft-ietf-quic-multipath.md
mirjak: This will be gone with the proposed edit below. Resolving this comment for now to add the commit.

draft-ietf-quic-multipath.md
```diff
@@ -1081,33 +1081,32 @@ This is a major difference from
...
+This enables a smoother handover and allows a simplified migration
+handling at the server side as NAT rebindings imply immediate loss of the old
+address. However, if the server requests migration to a "preferred address",
+clients must initiate migration of path 0 to that preferred address as
+specified in {{QUIC-TRANSPORT}}.
```
mirjak, line 1097:
```suggestion
It is generally preferable to use multipath mechanisms such as
creating a new path and later abandoning the old path,
rather than doing migration of a single path as specified in {{QUIC-TRANSPORT}}.
This enables a smoother handover and allows a more controlled migration handling
at the server side. However, migration of a single path cannot be
avoided in case of NAT rebinding, or if the server requests migration
to a "preferred address" during the handshake.
```

huitema (approved):

draft-ietf-quic-multipath.md
```diff
@@ -1081,33 +1081,30 @@ This is a major difference from
...
-networks events such as NAT rebinding may make the packet's receiver
-observe a different 4-tuple. This changes the 4-tuple of the path but
-the connection ID remains the same, and so the related path identifier.
-Servers observing a 4-tuple change will
-perform path validation (see {{Section 9 of QUIC-TRANSPORT}}).
-If path validation process succeeds, the endpoints set
-the path's congestion controller and round-trip time
-estimator according to {{Section 9.4 of QUIC-TRANSPORT}}.
+For any given path, connection ID rotation, NAT rebinding, or client initiated migration
+as specified in {{QUIC-TRANSPORT}} might occur, similar as on a single path.
```
qdeconinck, line 1085:
```suggestion
as specified in {{QUIC-TRANSPORT}} might occur, like on a single path.
```

qdeconinck (approved):
Minor suggestion, but good to ship.

PR #525, opened 2025-04-07
small editorial edits on PATH_ACK usage

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -240,12 +240,11 @@ ACK frames when used with the multipath extension acknowledge packets for the pa
...
 Initial and Handshake packets are sent using ACK frames.
 
-After the handshake concluded and both endpoints support the multipath extension,
+After the handshake concluded with support for the multipath extension,
 endpoints SHOULD use PATH_ACK frames instead of ACK frames,
-including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0.
-Endpoints MUST process ACK frames that acknowledge 0-RTT packets or 1-RTT packets,
-even when this extension is used. For example,
-ACK frames might be preferred by the sender as long as only the initial path
+also so far unacknowledged 0-RTT packets using Path ID 0.
```
huitema, line 245: The initial writing is better. "also" is a bit ambiguous, and "also so far" does not really convey the meaning. I suggest reverting to "including for acknowledging so far unacknowledged 0-RTT packets using Path ID 0."

draft-ietf-quic-multipath.md
```diff
@@ -240,12 +240,11 @@ ACK frames when used with the multipath extension acknowledge packets for the pa
...
+Endpoints MUST still process ACK frames that acknowledge 0-RTT packets or 1-RTT packets.
+For example, ACK frames might be preferred by the sender as long as only the initial path
 with Path ID 0 is in use.
```
huitema, line 248: No strong opinion, but this will be interpreted as "please use ACK frames for path 0 if it is the only path." Maybe replace by "ACK frames might be preferred by some senders for various reasons."

huitema (commented):
Some changes suggested...

draft-ietf-quic-multipath.md
qdeconinck: Agree with Christian here. The "including" version is better.

draft-ietf-quic-multipath.md
qdeconinck, line 248: I also agree that this wording is a bit strange. I am not sure to understand the "only" here.
```suggestion
For example, ACK frames might be preferred by the sender to acknowledge packets
of the initial path with Path ID 0.
```

mirjak: The intention here was to give an example. The sentence you proposed is a bit meaningless. If there is a real concern, we should just remove the sentence. But I think it's useful to give an example as this was the one case that was brought up. I don't think this is a risk of being see as recommendation as I think our recommendation is clear. At the end it will be an implementation decision potentially mostly related to implementation complexity. However, the important point here is that the other end needs to be able to handle this case.

mirjak: Actually this changes the meaning of the example. I though the case that was brought up is that you might want to negotiate multipath support (for potential later use) but be able to have a quic connection that is behaving completely in the same way as a single path connection as long as you allow only for one path. That's why it says "only".

qdeconinck: With the provided explanation, this somewhat makes sense. However, without such context, I find this sentence a bit confusing. But this is not blocking.

mirjak: @qdeconinck any proposal how to improve the sentence?

qdeconinck: My proposal would be to add more context to the example, but the "example text" would start being larger. I will do a suggestion.

```suggestion
For example, a sender may negotiate multipath support for later use and keep
only the initial path with path ID 0 for a while. During this single-path period,
ACK frames might be preferred by the sender.
```

qdeconinck (approved):
My suggestion is only a suggestion; for the rest, we can merge.

draft-ietf-quic-multipath.md
mirjak: wpm

Yanmei-Liu (approved):
Looks good.

PR #533, opened 2025-04-07
Merge idle timeout and keep alive section

mirjak:


draft-ietf-quic-multipath.md
```diff
@@ -1308,23 +1307,11 @@ each path. The benefit of such an approach is to simplify retransmission
...
 
-## Keep Alive
-
-The QUIC specification defines an optional keep alive process, see {{Section 5.3 of QUIC-TRANSPORT}}.
-Implementations of the multipath extension should map this keep alive process to a number of paths.
-Some applications may keep only one path alive, while others could prefer to maintain
-liveliness on two or more paths during the connection lifetime.
-Different applications will likely require different strategies.
-Once the implementation has decided which paths to keep alive, it can do so by sending Ping frames
-on each of these paths before the idle timeout expires.
```
mirjak, line 1319: Note that is moved a shorter version of this text further down.

draft-ietf-quic-multipath.md
```diff
@@ -1337,12 +1325,15 @@ If a path is not actively used for a while, it might not be usable anymore,
...
+{{Section 5.3 of QUIC-TRANSPORT}} defines an optional keep alive process.
+This process can be applied to each path separately depending on application needs.
+Some applications may decide to not keep any not-actively used path alive,
+keep only one additonal path alive, or multiple paths, e.g. for more redunancy.
+As discussed in {{Section 10.1.2 of QUIC-TRANSPORT}}, the keep-alive interval
+needs to incorperate timeouts in middleboxes on the path.
+
+If a path was not actively used for a while and no keep alives have been sent,
+an endpoint can probe it before switching to active use if there are still other paths
 that are currently usable.
```
huitema, line 1338: The weird part of that paragraph is the conditional "if there are still other paths
that are currently usable." What we are saying is that and endpoint that want to use an idle path should probe it first, to verify that it is still usable. Of course if there is no other path available, the probing is a bit pointless, because if that last idle path does not work the connection will just break. But we we don't really need to say that.

huitema (commented):
Maybe take time to rethink the very last paragraph?

draft-ietf-quic-multipath.md
mirjak: Yes, but I don't see a real problem just saying it anyway and keeping the sentence as it is...

qdeconinck (approved):
Ok.

draft-ietf-quic-multipath.md
```diff
@@ -1337,12 +1325,15 @@ If a path is not actively used for a while, it might not be usable anymore,
...
-If a path was not actively used for a while, an endpoint can
-probe it before switching to active use if there are still other paths
+{{Section 5.3 of QUIC-TRANSPORT}} defines an optional keep alive process.
+This process can be applied to each path separately depending on application needs.
+Some applications may decide to not keep any not-actively used path alive,
+keep only one additonal path alive, or multiple paths, e.g. for more redunancy.
+As discussed in {{Section 10.1.2 of QUIC-TRANSPORT}}, the keep-alive interval
+needs to incorperate timeouts in middleboxes on the path.
+
+If a path was not actively used for a while and no keep alives have been sent,
```
Yanmei-Liu, line 1335:
```suggestion
If a path was not actively used for a while and no more packets have been sent,
```

Yanmei-Liu (approved):
Basically I think the new sentences are much more clearer. A few nits.

draft-ietf-quic-multipath.md
mirjak: I don't think this change puts it right. Active use effectively already implies that no packets have been sent, however, this second part is here to be explicit that also no keep alives have been sent because that is the kind of probing we talk about. If we want to change this sentence we could just remove the second part about keep alives but I think it's okay to be explicit.

huitema: I think I agree with @mirjak.

huitema (approved):

draft-ietf-quic-multipath.md
mirjak: Okay, I will merge the PR now as it is.

PR #536, opened 2025-04-14
note on sending data frames on different paths

### LINKED ISSUE #535: Should we note anything on streams and scheduling?
Maybe all we could do is just noting that scheduling data frames that below to the same on different path will increase the delay to the max path delay for all packets on that stream due to re-ordering, and therefore it could be beneficial to keep data frames of one stream on the same path if the path capacity is large enough.

mirjak:
fixes #535 

This proposed note is an optional addition. It's also okay to not merge if we don't think it very useful

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1250,7 +1250,7 @@ exceptions), can be sent and received on any open path.
 As such, a packet scheduler is needed to decide which path to use
 for sending the next packet, among those paths with an open congestion window.
 If multiple paths are used to send data frames belonging to the same stream,
-data delivery will expirience the maximum delay of all used paths due to in-order delivery.
+data delivery will experience the maximum delay of all used paths due to in-order delivery.
 The scheduling is a local decision, based on the preferences of the application and the
 implementation.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1252,7 +1252,7 @@ for sending the next packet, among those paths with an open congestion window.
 If multiple paths are used to send data frames belonging to the same stream,
 data delivery will expirience the maximum delay of all used paths due to in-order delivery.
 The scheduling is a local decision, based on the preferences of the application and the
-implementation. 
+implementation.
 
 This implies that an endpoint may send and receive PATH_ACK
 frames on a path different from the one that carried the acknowledged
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1249,8 +1249,10 @@ Most frames, including control frames (PATH_CHALLENGE and PATH_RESPONSE being th
 exceptions), can be sent and received on any open path.
 As such, a packet scheduler is needed to decide which path to use
 for sending the next packet, among those paths with an open congestion window.
+If multiple paths are used to send data frames belonging to the same stream,
+data delivery will expirience the maximum delay of all used paths due to in-order delivery.
 The scheduling is a local decision, based on the preferences of the application and the
-implementation.
+implementation.
```
mirjak, line 1255:
```suggestion
implementation.
```

draft-ietf-quic-multipath.md
```diff
@@ -1249,6 +1249,8 @@ Most frames, including control frames (PATH_CHALLENGE and PATH_RESPONSE being th
 exceptions), can be sent and received on any open path.
 As such, a packet scheduler is needed to decide which path to use
 for sending the next packet, among those paths with an open congestion window.
+If multiple paths are used to send data frames belonging to the same stream,
+data delivery will expirience the maximum delay of all used paths due to in-order delivery.
 The scheduling is a local decision, based on the preferences of the application and the
```
huitema, line 1254: typo experience

huitema (approved):
No objection. I don't think it is really necessary, but it also doesn't hurt.

draft-ietf-quic-multipath.md
```diff
@@ -1249,6 +1249,8 @@ Most frames, including control frames (PATH_CHALLENGE and PATH_RESPONSE being th
 exceptions), can be sent and received on any open path.
 As such, a packet scheduler is needed to decide which path to use
 for sending the next packet, among those paths with an open congestion window.
+If multiple paths are used to send data frames belonging to the same stream,
+data delivery will expirience the maximum delay of all used paths due to in-order delivery.
```
mirjak, line 1253:
```suggestion
data delivery will experience the maximum delay of all used paths due to in-order delivery.
```

qdeconinck (approved):
True and good enough to explicitly state it.

PR #538, opened 2025-04-16
note applicability of future scheduling algorithms

mirjak:
This is an alternative PR to #521 extending the existing text in the intro rather than adding entirely new text later on.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -132,14 +132,16 @@ and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to cope with the remaining scenarios.
 
-Further, this document does not specify detailed algorithms that define
+Further, this document does not specify detailed scheduling algorithms that define
 how multiple, simultaneously open paths are used to send packets.
-As these differ depending on the application requirements,
+As these differ depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
+However, scheduling algorithms, that may be specified by the IETF in future,
+can be used with the multipath extension specified in this document.
 
 Specifically, while failover between Wi-Fi
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporarily in order
+it only uses two paths at the same time temporarily
 to avoid transmission pauses.
 Simultaneous path usage generally, however, needs more consideration
 than specified in this document to avoid negative performance


draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,20 @@ and the actual decision to setup or tear down paths are assumed
...
 defining mechanisms to cope with the remaining scenarios.
 
-Further, this document does not specify detailed algorithms that define
+Further, this document does not specify detailed scheduling algorithms that define
 how multiple, simultaneously open paths are used to send packets.
-As these differ depending on the application requirements,
+As these differ depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
+However, scheduling algorithms, that may be specified by the IETF in future,
+can be used with the multipath extension specified in this document.
```
huitema, line 141: Too many commas? Copy editors insist on no comm before "that", and the one after future is not useful.

huitema (approved):
No big issue.

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,20 @@ and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to cope with the remaining scenarios.
 
-Further, this document does not specify detailed algorithms that define
+Further, this document does not specify detailed scheduling algorithms that define
 how multiple, simultaneously open paths are used to send packets.
-As these differ depending on the application requirements,
+As these differ depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
+However, scheduling algorithms, that may be specified by the IETF in future,
```
mirjak, line 139:
```suggestion
However, scheduling algorithms that may be specified by the IETF in future
```

draft-ietf-quic-multipath.md
mirjak: I thought they help but removed them now.

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,20 @@ and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to cope with the remaining scenarios.
 
-Further, this document does not specify detailed algorithms that define
+Further, this document does not specify detailed scheduling algorithms that define
```
mirjak, line 135: Proposed by Gorry:
```suggestion
Further, this document does not specify scheduling algorithms that define
```

Yanmei-Liu (approved):

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,20 @@ and the actual decision to setup or tear down paths are assumed
...
-As these differ depending on the application requirements,
+As these differ depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
+However, scheduling algorithms that may be specified by the IETF in future
+can be used with the multipath extension specified in this document.
 
 Specifically, while failover between Wi-Fi
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporarily in order
+it only uses two paths at the same time temporarily
```
```suggestion
it only temporarily uses two paths at the same time
```

gorryfair (approved):

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,24 @@ and the actual decision to setup or tear down paths are assumed
...
-Further, this document does not specify detailed algorithms that define
+Further, this document does not specify scheduling algorithms that define
 how multiple, simultaneously open paths are used to send packets.
-As these differ depending on the application requirements,
+As these differ depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
+This extension can be used with different scheduling algorithms that,
+e.g., can range from support for failover to simulatenous
+use of the aggreate capacity across all active paths.
+There are IETF specifications for path fail-over, but
```
mirjak, line 142: Do we really need to mention that there is a failover spec? As I said earlier I don’t think RFC7829 is necessarily relevant here as it doesn’t really assume that both paths are used simultaneously. If we want to keep this however I think we need to give a reference…?

```suggestion
```

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,24 @@ and the actual decision to setup or tear down paths are assumed
...
+there are currently no IETF standards-track specifications for simultaneously
```
mirjak, line 143: Can we maybe just remove „standards-track“? Because this is also true for none standards-track specs.

```suggestion
there are currently no IETF specifications for simultaneously
```

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,24 @@ and the actual decision to setup or tear down paths are assumed
...
+(concurrently) using the aggregated capacity of a set of active paths.
```
mirjak, line 144: Do we need to say „aggregated capacity“ here again?

This would be less wordy:
```suggestion
(concurrently) using multiple paths.
```

mirjak (commented):
I merged a new suggestion from Gorry. If needed I‘d be okay to keep the text as it but I have still three small comments to clarify the text and also make it shorter.

draft-ietf-quic-multipath.md
huitema: I agree that we should strike out that sentence. I was almost ready to commit the suggestion, but will let @mirjak do that once consensus is assessed.

I did read RFC7829, and I don't think citing it would be useful. The multipath extension does not create a new path fail-over capability, because that capability is already defined in RFC 9000. For example, the concern about simultaneous failover creating a thundering herd issue applies equally with and without the multipath extensions to RFC 9000.

draft-ietf-quic-multipath.md
huitema: +1

huitema (approved):
Good suggestions for streamlining the text. Let's apply them.

draft-ietf-quic-multipath.md
```diff
@@ -132,18 +132,24 @@ and the actual decision to setup or tear down paths are assumed
...
 
-Further, this document does not specify detailed algorithms that define
+Further, this document does not specify scheduling algorithms that define
 how multiple, simultaneously open paths are used to send packets.
-As these differ depending on the application requirements,
+As these differ depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
+This extension can be used with different scheduling algorithms that,
+e.g., can range from support for failover to simulatenous
+use of the aggreate capacity across all active paths.
```
mirjak, line 141: This edit fixes a nit and aligns terminology (as active paths include paths with issued CIDs that are not open yet):
```suggestion
use of the aggregated capacity across all open paths.
```

draft-ietf-quic-multipath.md
gorryfair: I hoped mentioning path-failover strengthened the position for making the QUIC MP Extension a PS. It did at least figure when MP items were chartered in the past. It is OK to remove this.

draft-ietf-quic-multipath.md
gorryfair: RFC 6356, RFC8382 and RFC 8699 are all experimental specifications for scheduling, albeit not targeting QUIC, but they are all EXP.

draft-ietf-quic-multipath.md
gorryfair: Sure, I could live without "aggregated capacity" here if the word concurrently is understood to say that.

draft-ietf-quic-multipath.md
mirjak: These RFCs are all about congestion control but do not specify scheduling.

draft-ietf-quic-multipath.md
gorryfair: I agree RFC7829 is not about concurrent usage of paths for data - I see this as opening two paths and can keep them alive until a path change is needed.

qdeconinck (approved):
Agree, provided that current Mirja's suggestions are merged.

PR #541, opened 2025-04-17
Text proposed to address scope/expectation

gorryfair:
Here are a bunch of changes that I think cover all my corner cases of what the IETF has STD'ed/what would be regarded as EXP, and the caveats, while hopefully not constraining the use of MP-QUIC. I am super happier about the idea that this appears in the intro, rather than the spec.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -132,17 +132,17 @@ and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to cope with the remaining scenarios.
 
-The framework specified in this document can be used with 
-a range of scheduling algorithms that define
+This extension can be used with 
+different scheduling algorithms that define
 how multiple simultaneously open paths are used to send packets. 
-Examples of schedulers can range from support for failover to simulatenous concurrent
+Examples of schedulers can range from support for failover to simulatenous
 use of the aggreate capacity across all active paths.
 
-The specification of scheduling for multipath in
-the general Internet is outside the scope of this document.  
-There are existing IETF specifications for path fail-over, but
-concurrently no IETF standards-track specifications for simultaneously
-using the capacity of a set of active paths. 
+This document does not specify scheduling for multipath in
+the general Internet.
+There are IETF specifications for path fail-over, but
+there are currently no IETF standards-track specifications for simultaneously
+(concurrently) using the aggreate capacity of a set of active paths. 
 
 Because the scheduling policy differs depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -132,19 +132,16 @@ and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to cope with the remaining scenarios.
 
-This extension can be used with 
-different scheduling algorithms that define
-how multiple simultaneously open paths are used to send packets. 
-Examples of schedulers can range from support for failover to simulatenous
+Further, this document does not specify scheduling algorithms that define
+how multiple, simultaneously open paths are used to send packets.
+As these differ depending on application requirements,
+only some basic implementation guidance is discussed in {{impl-consideration}}.
+This extension can be used with different scheduling algorithms that,
+e.g., can range from support for failover to simulatenous
 use of the aggreate capacity across all active paths.
-
-This document does not specify a scheduling algorithm.
 There are IETF specifications for path fail-over, but
 there are currently no IETF standards-track specifications for simultaneously
-(concurrently) using the aggreate capacity of a set of active paths. 
-
-Because the scheduling policy differs depending on application requirements,
-only some basic implementation guidance is discussed in {{impl-consideration}}.
+(concurrently) using the aggregated capacity of a set of active paths. 
 Specifically, while failover between Wi-Fi
 and mobile networks is a well-known multipath use case,
 it only temporarily uses two paths at the same time 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -141,7 +141,7 @@ use of the aggreate capacity across all active paths.
 The specification of scheduling for multipath in
 the general Internet is outside the scope of this document.  
 There are existing IETF specifications for path fail-over, but
-concurrently no IETF standards-track specifications for simulatenously
+concurrently no IETF standards-track specifications for simultaneously
 using the capacity of a set of active paths. 
 
 Because the scheduling policy differs depending on application requirements,
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
 to be handled by the application. But this document does not prevent future extensions from
 defining mechanisms to cope with the remaining scenarios.
 
-Further, this document does not specify scheduling algorithms that define
-how multiple, simultaneously open paths are used to send packets.
-As these differ depending on application requirements,
-only some basic implementation guidance is discussed in {{impl-consideration}}.
-However, scheduling algorithms that may be specified by the IETF in future
-can be used with the multipath extension specified in this document.
+The framework specified in this document can be used with 
+a range of scheduling algorithms that define
+how multiple simultaneously open paths are used to send packets. 
+Examples of schedulers can range from support for failover to simulatenous concurrent
+use of the aggreate capacity across all active paths.
 
+The specification of scheduling for multipath in
+the general Internet is outside the scope of this document.  
+There are existing IETF specifications for path fail-over, but
+concurrently no IETF standards-track specifications for simulatenously
+using the capacity of a set of active paths. 
+
+Because the scheduling policy differs depending on application requirements,
+only some basic implementation guidance is discussed in {{impl-consideration}}.
 Specifically, while failover between Wi-Fi
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporarily
+it only temporarily uses two paths at the same time 
 to avoid transmission pauses.
-Simultaneous path usage generally, however, needs more consideration
+More general use of simultaneous paths, however, needs more consideration
 than specified in this document to avoid negative performance
 impacts, e.g., when stream data is distributed over multiple paths with
-different delays.
+different delays or to avoid excessive conestion of shared bottlenecks.
 
 ## Conventions and Definitions {#definition}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -153,7 +153,7 @@ to avoid transmission pauses.
 More general use of simultaneous paths, however, needs more consideration
 than specified in this document to avoid negative performance
 impacts, e.g., when stream data is distributed over multiple paths with
-different delays or to avoid excessive conestion of shared bottlenecks.
+different delays or to avoid excessive congestion of shared bottlenecks.
 
 ## Conventions and Definitions {#definition}
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -138,8 +138,7 @@ how multiple simultaneously open paths are used to send packets.
 Examples of schedulers can range from support for failover to simulatenous
 use of the aggreate capacity across all active paths.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
...
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporarily
+it only temporarily uses two paths at the same time 
 to avoid transmission pauses.
-Simultaneous path usage generally, however, needs more consideration
+More general use of simultaneous paths, however, needs more consideration
 than specified in this document to avoid negative performance
 impacts, e.g., when stream data is distributed over multiple paths with
-different delays.
+different delays or to avoid excessive conestion of shared bottlenecks.
```
mirjak, line 156:
```suggestion
different delays or to avoid excessive congestion of shared bottlenecks.
```

draft-ietf-quic-multipath.md
```diff
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
...
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporarily
+it only temporarily uses two paths at the same time 
 to avoid transmission pauses.
-Simultaneous path usage generally, however, needs more consideration
+More general use of simultaneous paths, however, needs more consideration
 than specified in this document to avoid negative performance
 impacts, e.g., when stream data is distributed over multiple paths with
-different delays.
+different delays or to avoid excessive congestion of shared bottlenecks.
```
mirjak, line 156: This is a separate issue and not necessarily a "negative performance impact". E.g. early version of BBR could lead to high loss but exactly because of its aggressiveness would provide better application performance.
Also congestion control is there to avoid excessive congestion and this has nothing to do with use of multipath. So what is the actually problem for shared bottlenecks that you want to point out here?

draft-ietf-quic-multipath.md
```diff
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
...
+The framework specified in this document can be used with 
+a range of scheduling algorithms that define
+how multiple simultaneously open paths are used to send packets. 
+Examples of schedulers can range from support for failover to simulatenous concurrent
+use of the aggreate capacity across all active paths.
+
+The specification of scheduling for multipath in
+the general Internet is outside the scope of this document.  
+There are existing IETF specifications for path fail-over, but
+concurrently no IETF standards-track specifications for simulatenously
```
```suggestion
concurrently no IETF standards-track specifications for simultaneously
```

draft-ietf-quic-multipath.md
```diff
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
...
+Examples of schedulers can range from support for failover to simulatenous concurrent
+use of the aggreate capacity across all active paths.
+
+The specification of scheduling for multipath in
+the general Internet is outside the scope of this document.  
+There are existing IETF specifications for path fail-over, but
+concurrently no IETF standards-track specifications for simultaneously
+using the capacity of a set of active paths. 
+
+Because the scheduling policy differs depending on application requirements,
```
mirjak, line 147: A couple fo comments:

- This text again used the term "framework" which we don't use in this document and I don't find appropriate.
- I personally prefer the phrasing "this document does not specify" instead of talking about "out of scope". I think that is more clear and otherwise the question comes up why is it out of scope. I think it's reflecting more how we decided to focus the document and leave other things for future work. Why do you think that the "out-of-scope" re-wording is needed/different/better?
- Are there existing IETF specification of path fail-over? Which? if so we should add a reference somewhere (not necessarily here but maybe in the scheduling section) but not sure what you mean.
- I don't think we need to say "simultaneously using the capacity of a set of" that is just more words and still say the same thing as "simultaneously using multiple paths".
- Actually I don't really understand why it is important to mention "standards-track" here?
- Saying "simultaneous concurrent" is saying the same thing twice. I don't see a different here. We keep using "simultaneous" in this document, so I think we should stick with it. We could change that but I don't see a difference.

draft-ietf-quic-multipath.md
gorryfair: It's an example, and if you think there are sensitivities you can skip the second example.
The example is about avoiding *excessive congestion*, and that can arise, or can not arise when  you share a common bottleneck - I'd say congestion is not a negative impact but excessive congestion is. 
Anyway, as I noted I REALLY don't mind not having the example.

draft-ietf-quic-multipath.md
```diff
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
...
 only some basic implementation guidance is discussed in {{impl-consideration}}.
```
mirjak, line 148: A couple fo comments:

- This text again used the term "framework" which we don't use in this document and I don't find appropriate.
- I personally prefer the phrasing "this document does not specify" instead of talking about "out of scope". I think that is more clear and otherwise the question comes up why is it out of scope. I think it's reflecting more how we decided to focus the document and leave other things for future work. Why do you think that the "out-of-scope" re-wording is needed/different/better?
- Are there existing IETF specification of path fail-over? Which? if so we should add a reference somewhere (not necessarily here but maybe in the scheduling section) but not sure what you mean.
- I don't think we need to say "simultaneously using the capacity of a set of" that is just more words and still say the same thing as "simultaneously using multiple paths".
- Actually I don't really understand why it is important to mention "standards-track" here?
- Saying "simultaneous concurrent" is saying the same thing twice. I don't see a different here. We keep using "simultaneous" in this document, so I think we should stick with it. We could change that but I don't see a difference.

mirjak: How about keep the old text and adding the following sentence, which I believes covers what you want to say above with less wordings:

```suggestion
Further, this document does not specify scheduling algorithms that define
how multiple, simultaneously open paths are used to send packets.
As these differ depending on application requirements,
only some basic implementation guidance is discussed in {{impl-consideration}}.
However, this extension can be used with different scheduling algorithms,
e.g., for failover or capacity aggregation, even though
there is no IETF document that specifies scheduling algorithms
for simultaneous path usage at this point of time.
```

draft-ietf-quic-multipath.md
mirjak: sorry I typed my second reply on an out-dated view where Gorry's reply didn't show up and as such our postings crossed each other. I'm okay with just remove it. Thanks!

draft-ietf-quic-multipath.md
```diff
@@ -132,21 +132,28 @@ and the actual decision to setup or tear down paths are assumed
...
-However, scheduling algorithms that may be specified by the IETF in future
-can be used with the multipath extension specified in this document.
-
 Specifically, while failover between Wi-Fi
 and mobile networks is a well-known multipath use case,
-it only uses two paths at the same time temporarily
+it only temporarily uses two paths at the same time 
 to avoid transmission pauses.
-Simultaneous path usage generally, however, needs more consideration
+More general use of simultaneous paths, however, needs more consideration
```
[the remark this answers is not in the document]
gorryfair: I think we differ on what "generally" implies. This isn't important to me.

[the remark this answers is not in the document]
gorryfair: ... You can write: Simultaneous path usage generally, however, needs more consideration.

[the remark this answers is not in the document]
mirjak: As agreed reverting to original text:
```suggestion
Simultaneous path usage generally, however, needs more consideration
```

draft-ietf-quic-multipath.md
mirjak: As agreed removing this additional text:
```suggestion
different delays.
```

draft-ietf-quic-multipath.md
gorryfair: I made a new suggestion, based on the above, see PR text.

mirjak: Thanks for the reference to RFC7829. I wasn't aware of this RFC. However, I’m not sure RFC7829 is really relevant here because I don’t think it necessarily assumes that packet are transmitted on more than one path at any point of time. It’s more you want to detect a breakage quickly and stop sending on one path and then start sending on another path. So in other words no concurrent use. However, I only had a quick look.

On the term concurrent multipath: MP-TCP uses “simultaneous". There is not a single occurrence of “concurrent” in that RFC8684. So why is this a requirement for this document now? Also we now refer to RFC9743 which explains and uses that term but given scheduling is not part of this document, I don’t see why we need to use it. I think “simultaneous” is equally clear. 

However, here is a proposal to add this to my proposed sentence above:
```suggestion
Further, this document does not specify scheduling algorithms that define
how multiple, simultaneously open paths are used to send packets.
As these differ depending on application requirements,
only some basic implementation guidance is discussed in {{impl-consideration}}.
This extension can be used with different scheduling algorithms,
e.g., for failover or capacity aggregation, even though
there is no IETF document that specifies scheduling algorithms
for simultaneous path usage, sometime also referred to as concurrent multipath, at this point of time.
```

@Gorry please let me know if that proposed text/sentence would be okay for you instead of your text?

However, I would also like to get more input from others before merging! @LPardue @huitema @qdeconinck @Yanmei-Liu

draft-ietf-quic-multipath.md
```diff
@@ -132,16 +132,23 @@ and the actual decision to setup or tear down paths are assumed
...
+use of the aggreate capacity across all active paths.
+
+This document does not specify a scheduling algorithm for multipath in
+the general Internet.
+There are IETF specifications for path fail-over, but
+there are currently no IETF standards-track specifications for simultaneously
+(concurrently) using the aggreate capacity of a set of active paths. 
+
+Because the scheduling policy differs depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
```
mirjak, line 148: This is a proposal to rearrange the text to make the different to the existing text smaller:
```suggestion
Further, this document does not specify scheduling algorithms 
for multipath usage in the general Internet.
As these differ depending on application requirements,
only some basic implementation guidance is discussed in {{impl-consideration}}.
This extension can be used with  different scheduling algorithms that
e.g. can range from support for failover to simultaneous
use of the aggregated capacity across all active paths.
There are IETF specifications for path fail-over, but
there are currently no IETF standards-track specifications for simultaneously
(concurrently) using the aggreate capacity of a set of active paths. 
```
However, I still have some comments on this text:
a) I don't think we should say "scheduling for the general Internet" because this document does actually not specify any scheduling algorithm. If you add "for the general Internet" this sounds like it does specify something but it can only not be used in the general Internet. Why is that needed?
b) I would like to still remove "standards track" because I don't think it matter if that doc would be PS or informational and also this sound like there is only no PR doc but there is no doc at all.
c) If we say there are specs for failover we should add a reference, however, as I said earlier I don't think RFC7829 is really relevant here.
d) I would still prefer a shorter version as proposed by my previously

gorryfair: (a) Changed. 
(b, c) The principle behind RFC7829 might or might not be relevant - that would be for the WG to decide. If the WG thinks it is not relevant background for QUIC, then the I-D could say there are no current IETF standards-track documents defining schedulers for MP_QUIC for use in the general Internet.
- I'd much prefer text that also includes something like "if and when the IETF standardises a scheduling method this can be used with this extension".
(d) Of course, you might prefer your own shorter text;-). 

I will anyway be sure to look to what has been finally written, and will ask questions when the WG decides to request publication.

draft-ietf-quic-multipath.md
```diff
@@ -132,16 +132,22 @@ and the actual decision to setup or tear down paths are assumed
...
+Examples of schedulers can range from support for failover to simulatenous
+use of the aggreate capacity across all active paths.
+
+This document does not specify a scheduling algorithm.
+There are IETF specifications for path fail-over, but
+there are currently no IETF standards-track specifications for simultaneously
+(concurrently) using the aggreate capacity of a set of active paths. 
+
+Because the scheduling policy differs depending on application requirements,
 only some basic implementation guidance is discussed in {{impl-consideration}}.
```
mirjak, line 147: Thanks for further updates.

Proposing again to re-order your sentences to keep changes to the existing text minimal:
```suggestion
Further, this document does not specify scheduling algorithms that define
how multiple, simultaneously open paths are used to send packets.
As these differ depending on application requirements,
only some basic implementation guidance is discussed in {{impl-consideration}}.
This extension can be used with different scheduling algorithms that,
e.g., can range from support for failover to simulatenous
use of the aggreate capacity across all active paths.
There are IETF specifications for path fail-over, but
there are currently no IETF standards-track specifications for simultaneously
(concurrently) using the aggregated capacity of a set of active paths. 
```

Issue #562, opened 2025-05-12
Is the text about backup paths and timeouts encouraging waste?


martinthomson:
> To avoid such path breakage, endpoints can send ack-eliciting packets such as packets containing PING frames ([Section 19.2](https://rfc-editor.org/rfc/rfc9000#section-19.2) of [[QUIC-TRANSPORT](https://www.rfc-editor.org/rfc/rfc9000)]) on that path to keep it alive.

There's obviously a balance to be struck, but this text doesn't really acknowledge the flipside of that balance.  It probably should.

mirjak:
At least I think it might be good to more explicitly state that there is no requirement to send keep-alives on an unused but still open paths. I guess that got somehow lost in the last edit or maybe was never clearly there. I'll propose a sentence!

PR #565, opened 2025-05-12
Edits

martinthomson:
Just the minor stuff.  Issues are open for questions and more substantive feedback.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -261,9 +261,9 @@ with the packet number. When multiple packet number spaces are used,
 the packet number alone would not guarantee the uniqueness of the nonce.
 Therefore, the nonce N is calculated for 1-RTT if the multipath extension is used
 by combining the packet protection
-IV with the packet number and with the least significant 32 bits of the
+IV with the packet number and with the 32 bits of the
 path ID. In order to guarantee the uniqueness of the nonce, the path ID
-is limited to a max value of 2^32-1, as specified in Section {{nego}}
+is limited to a max value of 2^32-1, as specified in {{nego}}
 
 To calculate the nonce, a 96-bit path-and-packet-number is composed of the least
 significant 32 bits of the path ID in network byte order,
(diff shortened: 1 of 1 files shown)


Yanmei-Liu (approved):
Looks good! Thanks for all the nits solved.

huitema (approved):
Thanks, Martin. I have verified that all these edits are editorial. And the use of \<sup\>..\</sup\> is pretty cool, I was not aware that we could do that in Markdown. What's next, full LateX compatibility?

martinthomson:
For `<sup>`, the markdown processor passes any XML straight through, so you can just write XML if you preferred that.  `<sup>` is understood by xml2rfc, which passes that straight through to HTML.  It renders it with `^` when rendering text as well.

qdeconinck (approved):

draft-ietf-quic-multipath.md
```diff
@@ -1312,10 +1312,8 @@ paths allows.
 An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
-One simple option, if the PMTUs are relatively similar, is to apply the minimum PMTU of all paths to
-each path. The benefit of such an approach is to simplify retransmission
-processing as the content of lost packets initially sent on one path can be sent
-on another path without further frame scheduling adaptations.
+One simple option, if the PMTUs are similar, is to apply the minimum PMTU of all paths to
+each path.
```
mirjak, line 1316: We have a separate issue for this #562 and I guess are still discussion it. Either we keep this PR open until the issues is resolved or we move this chance to a separate PR.

mirjak: Okay, reverting removing this sentence as this is addressed by PR #578 now to be able to merge this editorial PR now.
```suggestion
One simple option, if the PMTUs are similar, is to apply the minimum PMTU of all paths to
each path. The benefit of such an approach is to simplify retransmission
processing as the content of lost packets initially sent on one path can be sent
on another path without further frame scheduling adaptations.
```

PR #566, opened 2025-05-21
Warn about denial of service using multiple paths.

### LINKED ISSUE #563: Request forgery section is not about request forgery
A lot of the discussion in [Section 7.2](https://datatracker.ietf.org/doc/html/draft-ietf-quic-multipath-14#section-7.2) is not about the risk of request forgery - the construction of UDP datagrams that might appear to be valid datagrams for a different protocol - but the consent to communicate risk inherent in establishing new flows.  That is, amplification.

The request forgery risk with this protocol is almost exactly the same as QUIC[^1].  This is a difference that is not worth mentioning.

[^1]: The addition of the path ID to the nonce has a miniscule impact. But only if the strategy involves what amounts to finding collisions, as you can trial several paths at a time, so you have a bigger search space.  However, that's unlikely to be the main factor anyway.

What the text concerns itself with is amplification risk, which is not that interesting either.  There's a minor cost difference to an attacker when it comes to establishing multiple paths from which to mount an attack, as opposed to multiple connections, that's true.  More on that in the next...

### LINKED ISSUE #564: Validating multiple paths is a DoS risk
Though Section 7.2 is mistitled (#563), the DoS risk it discusses is worth addressing. The main thing I'm concerned about is spikes in packet arrival rate due to path establishment.  It seems like there is no real guidance here.  There is a suggestion that an endpoint might limit its validation.  But that isn't really enough.

If paths share a bottleneck and path validation for 50 new paths would generate a spike in packets that hit that bottleneck in a way that it can't handle, we have a real problem.  I would have said that the only right thing to do is to have a stronger recommendation on the maximum number of paths that can be concurrently validated.  Setting that limit at one would be the obvious place to start.

This is not so much about amplification in the same sense we use in QUIC, because the number of inbound data would still need to reach one third of the outbound, but we also assumed in QUIC that path validation was below the baseline rate that a single link would be able to tolerate.  With multiple paths, that same assumption no longer holds.

huitema:
This PR is based on the discussion of issue #563 and #564

Close #563
Close #564

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1098,7 +1098,8 @@ Handshake and Application packets).
 
 For any given path, connection ID rotation, NAT rebinding, or client initiated migration
 as specified in {{QUIC-TRANSPORT}} might occur, like on a single path.
-These events do not change the path ID,  and do not affect the packet number
+These events do not change the path ID,
+ and do not affect the packet number
 space associated with the path.
 
 It is generally preferable to use multipath mechanisms such as
@@ -1404,7 +1405,7 @@ The multipath extension retains all security properties of {{QUIC-TRANSPORT}} an
 but requires some additional consideration regarding:
 
 - potential additional resource usage for per-path connection IDs and multiple concurrent path contexts;
-- a potentially increased amplification risk for request forgery attacks if multiple paths are used simultaneously;
+- a potentially increased amplification risk for denial of service attacks if multiple paths are used simultaneously;
 - changes to the nonce calculation due to the use of multiple packet number spaces.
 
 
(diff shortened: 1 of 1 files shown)


kazuho (commented):
Thank you for the PR.

I think the direction is good although it seemed to me that the new text uses "path" for two meanings: a flow identified by 5-tuple and a path as identified by a Path ID.

This might be a broader issue but I would hope the text could be clearer.

martinthomson (approved):

martinthomson:
@kazuho's points are good.  Maybe you need to think about establishing a convention that allows you to make a distinction between "path" as used generally (the thing that has a path ID) and the network path itself.

huitema:
In the context of this document, path always refer to "the packets exchanged with the same path-id". This is indeed the potential attack: use multiple path ID, but tweak the addresses so the server ends up sending packets on the same network path.

draft-ietf-quic-multipath.md
```diff
@@ -1098,7 +1098,8 @@ Handshake and Application packets).
 
 For any given path, connection ID rotation, NAT rebinding, or client initiated migration
 as specified in {{QUIC-TRANSPORT}} might occur, like on a single path.
-These events do not change the path ID,  and do not affect the packet number
+These events do not change the path ID,
+ and do not affect the packet number
```
qdeconinck, line 1102: I am not sure if this was an unintentional change (this does not change anything except adding a new line?).
```suggestion
These events do not change the path ID, and do not affect the packet number
```

draft-ietf-quic-multipath.md
```diff
@@ -1423,28 +1424,34 @@ To avoid unnecessary resource usage that could be exploited
...
 for migration is used
-unchanged for path initiation in this extension. Therefore, the security considerations
-on source address spoofing as outlined in {{Section 21.5.4 of QUIC-TRANSPORT}} equally apply.
-Similarly, the anti-amplification limits as specified in {{Section 8 of QUIC-TRANSPORT}} need to be
-followed to limit the amplification risk.
+unchanged for path initiation in this extension.
+But the multipath extensions allow for the creation of multiple paths, which means
+that in addition to the security considerations
+on source address spoofing outlined in {{Section 21.5.4 of QUIC-TRANSPORT}},
+there is a risk of amplified DOS attacks through simultaneous opening
```
qdeconinck, line 1435:
```suggestion
there is a risk of amplified DoS attacks through simultaneous opening
```

qdeconinck (approved):
Overall good, small nits.

Issue #544, opened 2025-05-10
Editorial pass for WGLC on -14 (Tommy)


tfpauly:
During my read-through of the document, I had the following comments. These are generally editorial. This is for draft-ietf-quic-multipath-14.

### Section 1

"4-tuple" is used without being directly defined, and I don't think that term is used in RFC 9000. Having a simple explanation of what tuple you mean would be good for people who aren't familiar. This is also used in section 5.2.

> There are currently no IETF specifications for simultaneously (concurrently) using multiple paths.

I think this would be more clear to explain that "There are currently no IETF specifications _that define scheduling algorithms_ for...". Without that clarification, it could sound like the IETF specifications prohibit simultaneous use (which isn't true) instead of just not specifying the algorithms for how to schedule them.

The intro defines the notion of "path ID" as part of the body of text, but it might be nice to have a separate sub-section that defines the normative aspects of the path ID. Right now, very crucial details about the path ID being an integer that monotonically increases, isn't reused, has special values for 0, etc, is mixed into the overview/conceptual text.

### Section 2.1

> When advertising the initial_max_path_id transport parameter, endpoints MUST use non-zero length Source and Destination Connection IDs.

The hyphens here feel a bit odd, like it should really be "non-zero-length". I suggest either one of these options:
- "endpoints MUST use Source and Destination Connection IDs with non-zero lengths."
- "endpoints MUST NOT use zero-length Source and Destination Connection IDs."

### Section 2.3

> The choice of three times the largest PTO is a trade-off.

It feels like this sentence should say "is a trade-off between ___ and ___.", for clarity.

### Section 3

This is the first time that some of the new frame types are mentioned, such as `PATH_ABANDON`, `PATH_NEW_CONNECTION_ID`, `PATH_RETIRE_CONNECTION_ID`. These don't consistently provide references to the sections where they are defined below. Please either add the references on these first uses, or maybe have an overview at the top of section 3 that mentions all of the new frame types and provides references at that point; or just move the frame definitions up.

### Section 4.3

I have a slight preference to change `PATH_BACKUP` -> `PATH_IS_BACKUP` or something similar. "BACKUP" alone sounds like a verb like "ABANDON", so initially I read it as "back this path up", which doesn't make much sense. A path _being_ a backup, like it can be available, makes more sense.

### Section 4.2.1

The error codes defined here seem specific to paths, but they go in the main error registry. Should their names have more clear path-related names?

[...]

PR #571, opened 2025-06-12
Rename PATH_BACKUP/AVAILABLE to PATH_STATUS_BACKUP/AVAILABLE

mirjak:
this was proposed by Tommy in issue #544

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -542,7 +542,7 @@ to use a different connection ID for the same path ID when doing so.
 
 ## Path Status Management {#path-state}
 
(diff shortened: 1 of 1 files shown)


flub:
The asymmetry between PATH_IS_BACKUP and PATH_AVAILABLE is now worse than the concern in the issue IMHO.  If the naming really needs to change then PATH_STATUS_BACKUP, PATH_STATUS_AVAILABLE is perhaps an other choice to consider?

tfpauly:
Sure, I think @flub's suggestion would work well too

tfpauly (approved):

qdeconinck (approved):
Thank you.

mirjak:
As everybody seems happy with this renaming, I'm merging it now!

PR #575, opened 2025-06-13
replace lower-case may with might or could

### LINKED ISSUE #548: "may" means "could" or "might"?
There are a few cases where an all-lowercase "may" is used.  This is fine, because RFC 8174 exists, but some of these are potentially confusing, because that word has dual meanings that is sometimes permissive and other times just a statement of possibility.

mirjak:
fixes #548

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -695,7 +695,7 @@ sent by the peer will not cause the closure of the QUIC connection.
 
 ### Handling PATH_ACK for Abandoned Paths {#ack-after-abandon}
 
-When an endpoint sends a PATH_ABANDON frame, there mighht
+When an endpoint sends a PATH_ABANDON frame, there might
 still be some packets in transit from the peer.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -695,9 +695,9 @@ sent by the peer will not cause the closure of the QUIC connection.
 
 ### Handling PATH_ACK for Abandoned Paths {#ack-after-abandon}
 
-When an endpoint sends a PATH_ABANDON frame, there may
+When an endpoint sends a PATH_ABANDON frame, there mighht
```
Yanmei-Liu, line 698:
```suggestion
When an endpoint sends a PATH_ABANDON frame, there might
```

Yanmei-Liu: Looks good. Just one nit.

Yanmei-Liu (approved):

huitema (approved):

qdeconinck (approved):
Ok.

PR #576, opened 2025-06-13
Don't use the term ack delay

### LINKED ISSUE #559: Acknowledgement delay definition confusion
> Instead of using RTT, we define the acknowledgment delays as the sum of two one-way delays: the delay on the packet sending path and the delay on the return path chosen for the acknowledgments. 

Acknowledgement delay is defined in RFC 9000 as something else entirely: the delay between when a packet is received and when an acknowledgement is sent.  That suggests both that you have three things, not two, and that the terminology is probably in need of adjustment.

mirjak:
fixes #559

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1194,15 +1194,15 @@ per path ID. However, depending on which path acknowledgements are
 sent, the actual RTT of a path cannot be calculated or may not be
 the right value to be used.
 
-Instead of using RTT, we define the acknowledgment delays as
+Instead of using the real RTT of a path, it is recommended to use
 the sum of two one-way delays: the delay
 on the packet sending path and the delay on the return path chosen
 for the acknowledgments.  When different paths have different
-characteristics, acknowledgment delays can vary
+characteristics, the delays can vary
 widely. Consider for example a multipath transmission using both a
 terrestrial path, with a latency of 50ms in each direction, and a
 geostationary satellite path, with a latency of 300ms in each
-direction.  The acknowledgment delay will depend on the combination
+direction.  The sum of the two one-way delays will depend on the combination
 of paths used for the packet transmission and the acknowledgement transmission,
 as shown in {{fig-example-ack-delay}}.
 


huitema (changes requested):
NO!!!!

This changes will bring further confusion. We do have a direct measurement of one-way-delays, and in the absence of timestamps that's very hard. 

We have an issue of vocabulary, because in RFC 9000 the "ack delay" is defined as the delay between packet arrival and acknowledgement departure. We need  to replace it, but we should not require that endpoints handle one-way-delay.

Let's think about this a bit more.

draft-ietf-quic-multipath.md
```diff
@@ -1194,15 +1194,15 @@ per path ID. However, depending on which path acknowledgements are
 sent, the actual RTT of a path cannot be calculated or may not be
 the right value to be used.
 
-Instead of using RTT, we define the acknowledgment delays as
+Instead of using the real RTT of a path, it is recommended to use
 the sum of two one-way delays: the delay
```
huitema, line 1198: instead of "use", I would say "consider". We don't want implementations to develop estimators of one-way delays!

draft-ietf-quic-multipath.md
```diff
@@ -1222,7 +1222,7 @@ the satellite channel, but it is still the right value for computing
 for example the PTO timeout: if a PATH_ACK is not received after more
 than 350ms, either the data packet or its PATH_ACK were probably lost.
 
-The simplest implementation is to use the acknowledgment delay
+The simplest implementation is to use the sum of the two one-way delays
 to compute smoothed_rtt and rttvar per
```
huitema, line 1226: Maybe replace "the sum of one-way delays" by "the delays measured when receiving new packet acknowledgements"

huitema (commented):
Proposed minimal changes to make sure implementations understand that they do not have to measure one way delays.

draft-ietf-quic-multipath.md
```diff
@@ -1222,7 +1222,7 @@ the satellite channel, but it is still the right value for computing
 for example the PTO timeout: if a PATH_ACK is not received after more
 than 350ms, either the data packet or its PATH_ACK were probably lost.
 
-The simplest implementation is to use the acknowledgment delay
+The simplest implementation is to use the sum of the two one-way delays
```
mirjak, line 1225:
```suggestion
The simplest implementation is to use the the delays measured when receiving new packet acknowledgements
```

huitema (approved):

qdeconinck (approved):
Ok.

PR #577, opened 2025-06-13
calculate RTT if ACKs are not sent consistently

### LINKED ISSUE #560: The simple RTT calculation could be clearer
> The simplest implementation is to use the acknowledgment delay to compute smoothed_rtt and rttvar per [Section 5.3](https://rfc-editor.org/rfc/rfc9002#section-5.3) of [[QUIC-RECOVERY](https://www.rfc-editor.org/rfc/rfc9002)] regardless of the path through which PATH_ACK frames are received. This approach will provide good results as long as acknowledgements are sent consistently over the same paths \[sic].

It's not very clear to me what "same paths" means in this situation.  Does it mean that it assumes that the PATH_ACK is sent on the same path as the packets that it acknowledges?  Or that it always flows over the same path, even if that is a different return path?

It probably doesn't matter: if the PATH_ACK path remains the same for a given path, then things will work out.  But the text could be clearer in what it intends.

Either way, the remainder of this description is unclear.

> If at any time the sender revisits its sending preferences, this can also change the paths that are used to send acknowledgements. 

Why?  This doesn't make sense.  A sender changing its behavior might result in a receiver changing how it acknowledges, but that's pretty unlikely and there would be a few logical steps you'd have to describe.

The text on minimum RTT makes more sense, but it isn't resilient to constant changes.

What if the receiver is not consistent about their choice of path?  That seems incomplete and quite risky if your only recommendation is the simplistic one.

mirjak:
fixes #560

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1241,7 +1241,7 @@ the minimum RTT was received, and restart the minimum RTT computation
 if that acknowledgement path changes or is abandoned.
 If acknowledgements are not send consistently over one path, the
 acknowledgement receiver can monitor over which path acknowledgement
-are received any ony use samples for acknowledgements received on the same
+are received and only use samples for acknowledgements received on the same
 path than the data was sent, if any, or simply use only samples from the
 acknowledgement path with the largest delay.
 
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1226,7 +1226,7 @@ The simplest implementation is to use the sum of the two one-way delays
 to compute smoothed_rtt and rttvar per
 {{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which PATH_ACK frames are
 received. This approach will provide good results
-as long as acknowledgements are sent consistently over one paths.
+as long as acknowledgements are sent consistently over one path.
 If at any time the acknowledgement sender revisits its sending preferences,
 this can also change the paths that are used to send acknowledgements.
 However, this is not very
@@ -1240,7 +1240,7 @@ can, however, remember the path over which the PATH_ACK that produced
 the minimum RTT was received, and restart the minimum RTT computation
 if that acknowledgement path changes or is abandoned.
 If acknowledgements are not send consistently over one path, the
-acknowledgement receiver can monitor over with path acknowledgement
+acknowledgement receiver can monitor over which path acknowledgement
 are received any ony use samples for acknowledgements received on the same
 path than the data was sent, if any, or simply use only samples from the
 acknowledgement path with the largest delay.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1226,8 +1226,8 @@ The simplest implementation is to use the sum of the two one-way delays
 to compute smoothed_rtt and rttvar per
 {{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which PATH_ACK frames are
 received. This approach will provide good results
-as long as acknowledgements are sent consistently over the same paths.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1239,6 +1239,12 @@ An acknowledgement receiver
 can, however, remember the path over which the PATH_ACK that produced
 the minimum RTT was received, and restart the minimum RTT computation
 if that acknowledgement path changes or is abandoned.
+If acknowledgements are not send consistently over one path, the
+acknowledgement receiver can monitor over with path acknowledgement
+are received any ony use samples for acknowledgements received on the same
+path than the data was sent, if any, or simply use only samples from the
+acknowledgement path with the largest delay.
```
huitema, line 1246: Please delete the text after "if any", up to "largest delay." It is hard to find a theoretical justification for using "largest delay".

huitema (commented):
We should acknowledge that we do not have enough experience to define a reasonable strategy. Accepting samples as they go works OK for RFC9002-style algorithm that maintain a running evaluation of smoothed RTT and variations. If the sender changes tactic and ack paths, there will be a period of adaptation, and then the algorithm will converge.

The main issue that I found in practice is the maintenance of min RTT. RFC 9002 algorithms do not rely on it, but other algorithms like BBR or LEDBAT do. If the sender switches from a low latency ack path to a higher latency path, these algorithms will keep using the old min RTT for a long time, and performance will suffer. The fix here is to reevaluate min RTT if the ACK path changes, but that requires finding out that the path did change. A plausible solution is to monitor the last min RTT ack path ID, and reevaluate min RTT if no new ACK is received on that path for some time -- maybe 3RTT.

draft-ietf-quic-multipath.md
```diff
@@ -1226,8 +1226,8 @@ The simplest implementation is to use the sum of the two one-way delays
 to compute smoothed_rtt and rttvar per
 {{Section 5.3 of QUIC-RECOVERY}} regardless of the path through which PATH_ACK frames are
 received. This approach will provide good results
-as long as acknowledgements are sent consistently over the same paths.
-If at any time the sender revisits its sending preferences,
+as long as acknowledgements are sent consistently over one paths.
```
mirjak, line 1229:
```suggestion
as long as acknowledgements are sent consistently over one path.
```

draft-ietf-quic-multipath.md
```diff
@@ -1239,6 +1239,12 @@ An acknowledgement receiver
 can, however, remember the path over which the PATH_ACK that produced
 the minimum RTT was received, and restart the minimum RTT computation
 if that acknowledgement path changes or is abandoned.
+If acknowledgements are not send consistently over one path, the
+acknowledgement receiver can monitor over which path acknowledgement
+are received any ony use samples for acknowledgements received on the same
```
mirjak, line 1244:
```suggestion
are received and only use samples for acknowledgements received on the same
```

huitema:
To repeat my previous comment: per RFC 9002, the measurement of delays is only used to decide to send a PTO. Packet are only considered lost when there is a hole in the ACK sequence. The PTO packet will trigger an ACK and thus provide an opportunity to detect this hole. Sending a PTO is not supposed to have an impact besides triggering the ACK -- in particular, it should not impact congestion control. Getting the delays wrong means sending a PTO too early, which causes a fairly minimal overhead, or sending a PTO too late, which delays loss recovery. Of these two risk, I would rather err on the side of "too early". That's a good reason to not make suggestions about "using the largest delay" -- that would be akin to jumping in the pond to avoid the rain.

draft-ietf-quic-multipath.md
mirjak: Okay, will remove.

PR #578, opened 2025-06-13
not retransmitting frames

### LINKED ISSUE #561: Bad assumptions about retransmissions
> One simple option, if the PMTUs are relatively similar, is to apply the minimum PMTU of all paths to each path. The benefit of such an approach is to simplify retransmission processing as the content of lost packets initially sent on one path can be sent on another path without further frame scheduling adaptations.

This implies a very bad assumption about how implementations handle the loss of packets.  It assumes that the contents are retransmitted, rather than constructing frames anew as needed.  This encourages sloppy thinking about how QUIC operates and this last sentence should be removed.

mirjak:
fixes #561

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1313,9 +1313,7 @@ An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
 One simple option, if the PMTUs are relatively similar, is to apply the minimum PMTU of all paths to
-each path. The benefit of such an approach is to simplify retransmission
-processing as the content of lost packets initially sent on one path can be sent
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1313,9 +1313,7 @@ An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
 One simple option, if the PMTUs are relatively similar, is to apply the minimum PMTU of all paths to
-each path. The benefit of such an approach is to simplify retransmission
-processing as the content of lost packets initially sent on one path can be sent
-on another path without further frame scheduling adaptations.
+each path which could also help to simplify retransmission processing.
```
huitema, line 1316: maybe comma after path.

huitema (approved):

Yanmei-Liu (approved):

Issue #581, opened 2025-06-16
Implementation Consideration - PMTU management


tzaeschke:
From the Section `Paths Having Different PMTU Sizes`

> As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
> DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.

Is it worth clarifying (here or elsewhere) that QUIC-MP implementation should probably maintain MPS per path ID instead of the 4-tuple?
We can now have multiple paths per 4-tuple, and if we use an underlay that is capable of path aware networking, these paths may have different MPSs.

See also discussion here: https://github.com/quicwg/multipath/issues/561

mirjak:
Thanks for flagging. However, this sentence is just citing what's written down in RFC9000:

"QUIC endpoints implementing DPLPMTUD maintain a DPLPMTUD Maximum Packet Size (MPS) ([Section 4.4](https://www.rfc-editor.org/rfc/rfc8899#section-4.4) of [[DPLPMTUD](https://datatracker.ietf.org/doc/html/rfc8899)]) for each combination of local and remote IP addresses."

We actually don't state any requirements. I guess we could add a sentence that explains that you could even have different MPS on the same 4-tuple given you can have multiple path on the same 4-tuple... would that help?

tzaeschke:
Yes, that help. I proposed a PR to this effect: #581

huitema:
I think the less said the better. There is no guarantee that two paths sharing the same 4 tuple will in fact experience the same MTU. That might be true more than 99% of the time, but certainly not 100%. For example, paths using the same 4-tuple but different DS code points may very well be routed differently.

mirjak:
PR merged

PR #587, opened 2025-07-03
Clarify PMTU management - #581

tzaeschke:
Clarify that a single 4-tuple can have multiple PMTU sizes. See #581

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1333,7 +1333,7 @@ paths allows.
 An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
-Note that with the multipath extension multiple paths could even use the same 4-tuple
+Note that with the multipath extension multiple paths could use the same 4-tuple
 which do not necessarily need to have the same MPS.
 One simple option, if the PMTUs are similar, is to apply the minimum PMTU of all paths to
 each path. The benefit of such an approach is to simplify retransmission
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1334,7 +1334,7 @@ An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
 Note that with the multipath extension multiple paths could use the same 4-tuple
-which do not necessarily need to have the same MPS.
+but might have different MPS.
 One simple option, if the PMTUs are similar, is to apply the minimum PMTU of all paths to
 each path. The benefit of such an approach is to simplify retransmission
 processing as the content of lost packets initially sent on one path can be sent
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1331,7 +1331,9 @@ paths allows.
 ## Paths Having Different PMTU Sizes
 
 An implementation should take care to handle different PMTU sizes across
-multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
+multiple paths. PMTU size could even differ on the same 4-tuple given that
+multiple path can use the same 4-tuple.
```
mirjak, line 1335:
```suggestion
multiple paths can use the same 4-tuple.
```

draft-ietf-quic-multipath.md
```diff
@@ -1333,6 +1333,8 @@ paths allows.
 An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
+Note that with the multipath extension multiple paths could use the same 4-tuple
+which do not necessarily need to have the same MPS.
 One simple option, if the PMTUs are similar, is to apply the minimum PMTU of all paths to
```
huitema, line 1338: This says the right thing, but does not sound right.

As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
_However, the Internet architecture does not guarantee that the paths chosen for different path IDs will actually
experience the same MPS, even if the paths use the same 4-tuple. It might be more robust to independently
perform DPLPMTUD independently for each path._

huitema (approved):
I think this is OK, but could be more precise. Your choice.

draft-ietf-quic-multipath.md
mirjak: That's okay for me but my main point was actually that other than with RFC9000, with the multipath extension you can have more than one path per 4-tuple/endpoint pair and therefore the statement to maintain one MSP per IP address pair is less useful.

huitema: OK. Maybe a simpler suggestion:

_Note that with the multipath extension multiple paths could use the same 4-tuple
but that these paths will not always have the same MPS._

tzaeschke: My original point was the same as @mirjak 's, but I also like the suggestion of independent DPLPMTUD.
Maybe, to make it a bit shorter, how about:
*However,  due to differences in routing, the paths chosen for different path IDs may not experience the same MPS, even if the paths use the same 4-tuple. It may be more robust to perform DPLPMTUD independently for each path.* ?

tzaeschke: Or even:
*Note that with the multipath extension multiple paths may use the same 4-tuple
but may have different MPS.* ?

draft-ietf-quic-multipath.md
```diff
@@ -1333,6 +1333,8 @@ paths allows.
 An implementation should take care to handle different PMTU sizes across
 multiple paths. As specified in {{Section 14.3 of QUIC-TRANSPORT}} the
 DPLPMTUD Maximum Packet Size (MPS) is maintained for each combination of local and remote IP addresses.
+Note that with the multipath extension multiple paths could use the same 4-tuple
+which do not necessarily need to have the same MPS.
```
mirjak, line 1337:
```suggestion
Note that with the multipath extension multiple paths could use the same 4-tuple
but might have different MPS.
```

huitema (approved):

PR #588, opened 2025-07-05
remove "data packet"

### LINKED ISSUE #586: Migration/Probing (was: Use of "data packet")
There are four occasions of "data packet" in the specification. This term is not defined, but I assume it means a QUIC packet that contains one or more STREAM or DATAGRAM frames.

Either define it or rewrite those occurances. The term is only used a few time in RFC 9000 and also not defined there. So I think avoiding it is probably for the better. 

This issue was noticed by Floris Bruynooghe.

mirjak:
fixes #586

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1272,7 +1272,7 @@ with different delays explicitly.
 
 ## Packet Scheduling {#packet-scheduling}
 
-The transmission of packets containing data frames is limited
+The transmission of packets containing data is limited
 by the arrival of data from the application and by congestion control.
 Generally, QUIC packets that increase the number of bytes in flight can only be sent
 when the congestion window for the selected path allows it.
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -352,13 +352,13 @@ MUST use a connection ID associated to the same path ID as used in the packet
 received by the endpoint when it intends to send packets on the same path.
 
 A client that wants to use an
-new path MUST validate the peer's address before sending any data packets
+new path MUST validate the peer's address before sending any data
 as described in {{Section 8.2 of QUIC-TRANSPORT}},
 unless it has previously validated the 4-tuple used for that path.
 
 After receiving packets from the
 client on a new path, if the server decides to use the new path,
-the server MUST validate the peer's address before sending any data packets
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -1246,7 +1246,7 @@ computed RTT value for the satellite path will be about 350ms. This is
 lower than the 600ms that would be measured if the PATH_ACK came over
 the satellite channel, but it is still the right value for computing
 for example the PTO timeout: if a PATH_ACK is not received after more
-than 350ms, either the data packet or its PATH_ACK were probably lost.
+than 350ms, either the sent packet or its PATH_ACK were probably lost.
```
huitema, line 1250: You could just say "either the *packet* ..."

draft-ietf-quic-multipath.md
```diff
@@ -1272,7 +1272,7 @@ with different delays explicitly.
 
 ## Packet Scheduling {#packet-scheduling}
 
-The transmission of data packets is limited
+The transmission of packets containing data frames is limited
 by the arrival of data from the application and by congestion control.
```
huitema, line 1276: Actually, that's true for not just data frames but also datagram frames, and in fact all frames not marked with the flag "C" in the table 3 of section 12.4 of RFC 9000. Only the ACK frames are not subject to congestion control. Not sure how to say that succintly. Maybe 

The transmission of ack-eliciting packets is limited

huitema (commented):
Couple of suggestions. Please fix the "data packet" part -- just referencing "data frames" is incorrect. The problematic text is in a paragraph describing scheduling issues, which confirm my bias that we should just say nothing about scheduling...

draft-ietf-quic-multipath.md
```diff
@@ -1246,7 +1246,7 @@ computed RTT value for the satellite path will be about 350ms. This is
 lower than the 600ms that would be measured if the PATH_ACK came over
 the satellite channel, but it is still the right value for computing
 for example the PTO timeout: if a PATH_ACK is not received after more
-than 350ms, either the data packet or its PATH_ACK were probably lost.
+than 350ms, either the sent packet or its PATH_ACK were probably lost.
```
mirjak, line 1249:
```suggestion
than 350ms, either the packet or its PATH_ACK were probably lost.
```

PR #591, opened 2025-07-11
Clarify probing packets do not exist

### LINKED ISSUE #590: Should frames be defined as probing or non-probing?
[RFC 9000 § 9.1](https://www.rfc-editor.org/rfc/rfc9000.html#name-probing-a-new-path) defines probing frames, probing packets and also non-probing frames and packets.  The multipath draft does not do so about the new frames it defines.  Is this an oversight and should it?

There is a vague hint in [a paragraph of QUIC-MULTIPATH § 4.3](https://www.ietf.org/archive/id/draft-ietf-quic-multipath-15.html#section-4.3-13) saying that PATH_STATUS_* frames MAY be bundled with a PATH_NEW_CONNECTION_ID of PATH_RESPONSE frame.  PATH_RESPONSE is a probing frame, NEW_CONNECTION_ID is also a probing frame.  So it is not unreasonable to expect PATH_NEW_CONNECTION_ID and PATH_STATUS_* frames to also be probing frames.

In particular it would seem reasonable to bundle PATH_CHALLENGE, PATH_NEW_CONNECTION_ID and PATH_STATUS_* frames all in one packet.  But as far as I can tell this is not currently allowed.  Declaring PATH_NEW_CONNECTION_ID and PATH_STATUS_* frames as probing frames would solve this.

flub:
This tweaks the language wrt to probing to make it more clear that probing and non-probing packets no longer exist.

Closes #590

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -451,10 +451,9 @@ associated with the same path ID.
 {{Section 9.1 of QUIC-TRANSPORT}} introduces the concept of
 "probing" and "non-probing" frames. A packet that contains at least
 one "non-probing" frame is a "non-probing" packet. When the multipath extension
-is negotiated, the reception of a "non-probing"
-packet on a new path with a new so far unused path ID
-does not impact the path status of any existing
-path. Therefore, any frame can be sent on a new path at any time
+is negotiated, no such distinction exist and opening new paths does
+not affect the status of any existing path.
+Therefore, any frame can be sent on a new path at any time
 as long as the anti-amplification limits
 (see {{Section 21.1.1.1 of QUIC-TRANSPORT}}) and the congestion control
 limits for this path are respected.


divagant-martian:
There's also this mention, if any packet "probes" a path it's perhaps clearer to no introduce the "distinction" of being a probing packet here as well:

> Similarly, the server MAY wait for a
   short, limited time such as one PTO if a path probing packet is
   received on a new path before sending the CONNECTION_CLOSE frame.

mirjak:
> There's also this mention, if any packet "probes" a path it's perhaps clearer to no introduce the "distinction" of being a probing packet here as well:
> 
> > Similarly, the server MAY wait for a
> > short, limited time such as one PTO if a path probing packet is
> > received on a new path before sending the CONNECTION_CLOSE frame.

Yes, I think that is an oversight that we didn't change after removing the probing concept and we can simply remove "path probing" here. I'll create a PR.

flub:
> This change is not correct. The probing concept still exists in the context of RFC9000 migration; this means if you receive a packet with a new 4-tuple using an active path ID.

I don't understand how the probing concept is still needed for this.  An RFC9000 migration is only allowed for involuntary migrations I think, because "To open a new path, an endpoint MUST use a new connection ID associated with an unused path ID".  So voluntary RFC9000-style migrations are disallowed.  And as this section (3.1.2) clearly states that probing or non-probing packets do not affect the path status when an RFC9000 migration happens.  So whether the RFC9000-style migration did happen with a probing packet or not does not make any difference.

PR #593, opened 2025-07-15
Attempt to clarify initial path probing behaviour

flub:
This is an attempt to improve understanding one of the two issues discussed in #586.  I'm not particularly happy with this wording, but do think it does add important context to understand what is allowed.  Even if it is not strictly needed from a normative perspective.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -351,9 +351,12 @@ with an unused path ID. An endpoint
 MUST use a connection ID associated to the same path ID as used in the packet
 received by the endpoint when it intends to send packets on the same path.
 
-A client that wants to use a
-new path MUST validate the peer's address before sending any data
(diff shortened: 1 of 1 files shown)


mirjak:
I feel this change is still not correct. Path validation and probing are two different things. The path validation requirements are the same for RFC9000 migration and opening a new path as described in this extension. Only probing doesn't exists, which simply means that the path can be used immediately without impact the state of any existing path. While with RFC9000 migration, you might sending only probing packets first and validate the path and then only when the client decided to send the first non-probing packet on that path, the server also need to switch over and start using the new and can't not use the old path anymore which then is implicitly closed.

flub:
The wording suggested here isn't much of an improvement.  The existing wording already does not mention the anti-amplification limit for the client side, while explicitly mentioning it for the server.  This is probably sufficient.

PR #594, opened 2025-07-15
Make it clear the CIDs are not actually retired

flub:
The way this was worded suggested that the CIDs are really retired, but that would involve sending PATH_RETIRE_CONNECTION_ID frames for all CIDs of an abandoned path.  This is not required by § 3.4 which only says you need to "treat the CIDs as immediately retired".

Aligning this wording here makes the intention clearer.


huitema (approved):

huitema (changes requested):
Uh no, this is wrong, the text means exactly what is said. Sending Abandon implicitly retires all CID defined for that path-ID by the server. There is no conditional, and the endpoint will not send explicit "retire CID" messages for these CID.

Issue #549, opened 2025-05-11
Is path migration being replaced?


martinthomson:
Section 3.1.2 talks about the relationship between this document and migration.  It isn't very clear.

It is unambiguous about involuntary migration to a new address tuple[^1].  It is not clear about whether a deliberate migration might happen, other than the explicit allowance for preferred address migration on Path ID 0. It does override the notion of probing vs. non-probing (a concept I'm not particularly wed to in QUIC, but an essential one if you are going to allow migration).

[^1]: This draft uses 4-tuple, which is a separate problem. I believe that 5-tuple is more correct, as it captures IP version, but we avoided that entire debate in QUIC by minting a different term entirely.

If this is replacing the concept of migration with the concept of path additions and removals[^2], which is fine, then please say that directly.

[^2]: With the exception of the preferred address thing, or "under normal operation", I guess, if you want to get pedantic.

huitema:
We are trying to say two things:

1. That the mechanisms of path migration defined in RFC9000 are still supported, because this is an extension, and also because we absolutely need them to manage the "preferred address" scenario.
2. That the multipath extensions provide a much better way to support scenarios like "make before break": create a new path, set its status to "backup" until the endpoints decide that it is needed, and after deciding to transition explicitly abandon the old path.

Do you have suggestions on how to improve the wording?

martinthomson:
That link is broken.  I think that you mean https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-5.1

Section 3.1.2 is where I would go to look for normative advice on the distinction.  But it's pretty vague, even in Section 5.1.

> However, migration of a single path cannot be avoided 

This could be more direct: Though multipath offers clients with better options for use of different network paths, a server implementation MUST handle migration, as defined in {{Section Foo of !QUIC}}.

mirjak:
I created #585 trying to add your sentence in section 3.2.1. As it seemed redundant to just add your proposed sentence, I adjusted the existing one with your wording. However, I'm not sure that's better/right this way. Would be good to get more people to review and comment!

Issue #347, opened 2024-05-25
Are MP_PATH_(AVAILABLE|STANDBY) probing frames?


MikeBishop:
The draft does not modify RFC 9000's definition that "PATH_CHALLENGE, PATH_RESPONSE, NEW_CONNECTION_ID, and PADDING frames are 'probing frames', and all other frames are 'non-probing frames'." However, it does say that "A PATH_(AVAILABLE|STANDBY) frame MAY be bundled with ... a PATH_RESPONSE frame in order to indicate the preferred path usage before or during path initiation."

This suggests that these frames might be treated as probing frames. Either way, the draft should explicitly state that they are or are not.

(Alternatively, the concept of probing packets/frames might be outmoded in a multipath world. This concept seems largely unused in this document except in Section 4.3.5.)

mirjak:
We discussed this already and decided that we don't need the concept of probing frames. This is why the draft has the following text:

```
[Section 9.1](https://rfc-editor.org/rfc/rfc9000#section-9.1) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)] introduces the concept of "probing" and "non-probing" frames. When the multipath extension is negotiated, the reception of "non-probing" packet on a new path needs to be considered as a path initiation attempt that does not impact the path status of any existing path. Therefore, any frame can be sent on a new path at any time as long as the anti-amplification limits ([Section 21.1.1.1](https://rfc-editor.org/rfc/rfc9000#section-21.1.1.1) of [[QUIC-TRANSPORT](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#QUIC-TRANSPORT)]) and the congestion control limits for this path are respected.[¶](https://quicwg.org/multipath/draft-ietf-quic-multipath.html#section-4.1-5)
```

Can we close this issue or is there anything else needed?

MikeBishop:
If so, I think I would make that an explicit statement that there is no distinction between probing/non-probing frames or packets once this extension is negotiated.  However, 4.3.5 does still make distinctions between probing and non-probing packets with regard to idle timeouts, I think?

mirjak:
Okay, we can double-check that!

mirjak:
[...]

PR #595, opened 2025-07-18
Simplify the relation to RFC9000-style migration text

flub:
This is an attempt to address #586 and aims to reduce the number of confusing references.

- This removes referring to probing and non-probing packets.  This is purely a mechanism concernging RFC9000-style migration and does not affect anything in multipath.

- Do not re-iterate how anit-amplification and congestion control interacts with anything.  Those are already described in this draft and RFC9000 respectively.  There is no special interaction with those so no need to repeat them.

- Do still describe that the path ID remains the same and thus CIDs must all be of the same path ID.

- Do still point out that such a migration does not affect any path status.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -448,23 +448,14 @@ associated with the same path ID.
 
 ### Relation to Probing and Migration
 
-{{Section 9.1 of QUIC-TRANSPORT}} introduces the concept of
-"probing" and "non-probing" frames. A packet that contains at least
-one "non-probing" frame is a "non-probing" packet. When the multipath extension
-is negotiated, the reception of a "non-probing"
-packet on a new path with a new so far unused path ID
-does not impact the path status of any existing
-path. Therefore, any frame can be sent on a new path at any time
-as long as the anti-amplification limits
-(see {{Section 21.1.1.1 of QUIC-TRANSPORT}}) and the congestion control
-limits for this path are respected.
+The connection migration system described in {{Section 9 of
+QUIC-TRANSPORT}} remains fully supported when the multipath extension
+is negotiated and MUST be accepted. Such a migration does not change
+the path ID, and as such the connection ID used on the new new path
+MUST be associated with the path ID of the path being migrated.
 
-An endpoint could receive a packet with a connection ID
-associated to an active path ID where the packet's 4-tuple does not match the 4-tuple
-currently used with that path ID. This MUST be treated as path migration,
-as specified in {{Section 9.3 of QUIC-TRANSPORT}}, with the constraint that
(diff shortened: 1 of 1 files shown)


mirjak:
> * [Is path migration being replaced? #549](https://github.com/quicwg/multipath/issues/549) seeks to clarify whether path migration is being replaced.  The text here makes this very clear.

I think it's a matter of taste which text is perceived as more clear. However, as we just recently extended the text, I wouldn't want to change it again easily without broad support that there is an issue that needs change.

> * [Clarification on Path Migration in relation to opening a new path #458](https://github.com/quicwg/multipath/issues/458) is concerned with whether RFC9000-style migrations are supported at all again, same so far.  But also clearly wants to recommend **not** using this mechanism if possible.  This is covered very clearly in § 5.1.

The point of this change was to also add something to 3.1 which contains all the normative text and not only in 5.1 which is "just" implementation guidance.

> * [Anti-amplification limits #422](https://github.com/quicwg/multipath/issues/422) is concerned about the anti-amplification limit.  This is clearly covered in the first few paragraphs of § 3.1 already.

Most of this text moved somewhere else already. This issues was just another examples of the large amount of discussion we already had.

> * [Are MP_PATH_(AVAILABLE|STANDBY) probing frames? #347](https://github.com/quicwg/multipath/issues/347) is wondering about what frames are probing frames.  This is basically a very similar question route that I went on because the current language is confusing and makes you think this matters.  But as you say, all new frames are non-probing by the RFC9000 definition, no need to change this.

I disagree here. As this comes up over and over again, it's better to be explicit than say nothing. We can rather add another sentence pointing out that by default all frames are non-probing but I so far felt that is not necessary. 

> So I don't think any of these issues are blockers to this change.

It's not about blocking but it's showing that there was already a huge amount of discussion which lead to the current text. Therefore I don't want to change this text easily again if there is no technical issue. Editorial changes can be a matter of taste and with all the discussion we had, it's not clear if there is a general problem about clarity here. It's not an easy matter but I think the current text is carefully focused on the things that were identified by previous discussions.

flub:
After chasing various wrong interpretations of the current text and being corrected on how to interpret it in #586, I was asked to suggest improvements to make it clearer.  That is all I tried to do here while taking the previous discussions into account as best I could.

mirjak:
@flub thanks for your engagement and effort in order to clarify things! As I think I said already, I know this is a pain point. However, given the past discussion we had on this already, I prefer to keep the current text rather than basically reverting some of the additions we made over time. I hope that is okay for you and I would therefore like to close this PR without merging. If you still have smaller wording improvements, I'm happy to consider them but I don't think this proposed change makes it necessarily more clear for all readers.

PR #597, opened 2025-08-05
Remove paragraph on uncoordinated close "rule"

### LINKED ISSUE #558: Uncoordinated abandonment of paths
> Uncoordinated abandon \[sic] of both endpoints may result in deleting both paths instead of just one. To avoid this pitfall, endpoints could adopt a simple coordination rule, such as only letting the client initiate closure of duplicate paths, or perhaps relying on the application protocol to decide which paths should be closed.

This seems like a great recipe for interoperability failure.

Why does the specification not indicate which endpoint is responsible for abandonment?  Given that a client is the only one that initiate paths, this seems like a pretty unrealistic scenario to begin with (#554 notwithstanding).  If this is possible, is there any reason to allow servers to do anything other than request that a client act to close a path?  At what layer does that decision-making occur?

mirjak:
fixes #558

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1174,11 +1174,6 @@ source and destination ports:
 Such unintentional use of the same 4-tuple on different paths ought to
 be rare. When they happen, the two paths would be redundant, and the
 endpoint could want to close one of them.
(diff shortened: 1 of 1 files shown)


huitema (approved):
Yes, suppressing this paragraph is the right thing to do at this stage.

PR #598, opened 2025-08-12
3 PTOs "rule"

### LINKED ISSUE #556: 3 PTO requirement for PATH_ABANDON
The text observes that packets can arrive on a path after it has been abandoned.  At a minimum, there is a one RTT window, plus variance, in which this is unavoidable.

The text correctly recommends that PATH_ACK be sent on another path.  But it makes no recommendation about how long an endpoint retain state.  I would recommend that 3 PTO is the best recommendation we could give for that.  Again, this is a "SHOULD" not "MUST" as the only consequence is loss of packets, which the protocol is resilient enough to handle.

mirjak:
fixes #556 

Minor rewording to hopefully make things more clear.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -710,8 +710,9 @@ Due to network delays, packets sent on an abandoned path can
 arrive well after the connection IDs have been retired.
 If not recognized as bound to the local
 connection, such packet triggers the peer to send a Stateless Reset
-packet. The rule to "retain knowledge of connection ID for 3 PTO
-after receiving a PATH_ABANDON"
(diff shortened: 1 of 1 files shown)


huitema (approved):
I do not object.

I am a bit concerned that someone will be trigger by "it cannot completely avoid that risk". The risk is actually fairly minimal, because the peer has deleted the "reset secret" state for these CIDs when sending the Abandon. The spurious Stateless Reset packet will be discarded, and will not cause any damage, apart from the extra energy spent handling a useless packet.

Yanmei-Liu (approved):
Looks good.

PR #599, opened 2025-08-12
Further clarifications on "probing"

### LINKED ISSUE #586: Migration/Probing (was: Use of "data packet")

mirjak:
fixes #586 

This is another attempt to further clarify the text on probing. It'a adding more words rather than having less, however, given we have an own section about this because the question came up repeatedly, it might be better to be as explicit as possible...?

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -450,7 +450,7 @@ associated with the same path ID.
 
 {{Section 9.1 of QUIC-TRANSPORT}} introduces the concept of
 "probing" and "non-probing" frames. A packet that contains at least
-one "non-probing" frame is a "non-probing" packet. 
+one "non-probing" frame is a "non-probing" packet.
 Migration as specified in {{Section 9.2 of QUIC-TRANSPORT}} is initiated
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -450,11 +450,14 @@ associated with the same path ID.
 
 {{Section 9.1 of QUIC-TRANSPORT}} introduces the concept of
 "probing" and "non-probing" frames. A packet that contains at least
-one "non-probing" frame is a "non-probing" packet. When the multipath extension
-is negotiated, the reception of a "non-probing"
-packet on a new path with a new so far unused path ID
+one "non-probing" frame is a "non-probing" packet.
```
mirjak, line 453:
```suggestion
one "non-probing" frame is a "non-probing" packet.
```

huitema (approved):
OK.

PR #600, opened 2025-08-13
Rename PATH_INTERFACE_UNSTABLE to PATH_UNSTABLE_OR_POOR

### LINKED ISSUE #557: Missing PATH_ABANDON codes
Interface instability is a local choice, which maybe some implementations can detect.  However, I would consider loss of the ability to communicate on a path to be a pretty strong reason for PATH_ABANDON, so maybe COMMUNICATION_FAILED.

Then there is the case where a path, such as a backup back, simply times out.  TIMED_OUT.

And then there are a bunch of cases where the choice is not application-driven, but driven by other considerations, like POOR_PERFORMANCE.

What other things have implementation and deployment experience with the protocol revealed?

mirjak:
fixes #557


huitema (commented):
Sure, why not. But why keep the textual description unchanged?

mirjak:
I thought updating the text would make this also more clear. Any real concerns about this @huitema ?

huitema:
No concern. That's what we want to say.

Issue #612, opened 2025-09-28
Application or Application data space?


LPardue:
Section 5.1 says

> With the multipath extension, each path uses a separate packet number space. This is a major difference from [[QUIC-TRANSPORT](https://www.rfc-editor.org/rfc/rfc9000)], which only defines three number spaces (Initial, Handshake and Application packets).

RFC 9000 seems to use the term "application data packet number space" so I think you'll want to add the "data" qualifier

mirjak:
Not sure I understand. RFC9000 also uses "packet number space(s)" with "application data".

mirjak:
Not sure I understand. RFC9000 also uses "packet number space(s)" without "application data".

LPardue:
Pr was merged so closing this

Issue #611, opened 2025-09-28
grammar: delay sending the PATH_RESPONSE until sufficient resource are available


LPardue:
old
delay sending the PATH_RESPONSE until sufficient resource are available

new
delay sending the PATH_RESPONSE until sufficient resources are available

Issue #610, opened 2025-09-28
Consider citing RFC 9000 when referring to frames


LPardue:
Section 3 deals with path management and the frames that relate to that. On the first occurence of frames already defined in RFC 9000 it is not obvious (to me at least) that these are not new frames. Consider

old

> If an endpoint sends a PATH_RESPONSE, it MUST be sent on the same path as used by the packet that contained the PATH_CHALLENGE frame, using a connection ID associated with the same path ID.

new

> If an endpoint sends a PATH_RESPONSE frame (Section 19.18 of [QUIC-TRANSPORT]), it MUST be sent on the same path as used by the packet that contained the PATH_CHALLENGE frame (Section 19.17 of [QUIC-TRANSPORT]), using a connection ID associated with the same path ID.

Issue #609, opened 2025-09-28
PATH_BACKUP (see Section 4.3)


LPardue:
I think this should be 

PATH_STATUS_BACKUP (see [Section 4.3](https://datatracker.ietf.org/doc/html/draft-ietf-quic-multipath#path-backup-available-frame))

PR #615, opened 2025-10-10
Mostly editorial comments by Lucas

### LINKED ISSUE #604: Importing of RFC 9000 terms
In Section 1.1, the draft says:

> We assume that the reader is familiar with the terminology used in [[QUIC-TRANSPORT](https://www.rfc-editor.org/rfc/rfc9000)].

1. It might help to drop the `we assume` part here, the reader needs to understand the QUIC terminology or else they are unlikely to be albe to use the extension. 

2. It might help to state the use of QUIC notational conventions too

mirjak:
fixes #604, #605, #606, #608, #609, #610, #611, #612

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1,6 +1,6 @@
 ---
-title: Multipath Extension for QUIC
-abbrev: Multipath QUIC
+title: Managing multiple paths for a QUIC connection
+abbrev: Multiple Paths for QUIC
 docname: draft-ietf-quic-multipath-latest
 date: {DATE}
 category: std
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-multipath.md
```diff
@@ -160,18 +160,22 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
...
-{{QUIC-TRANSPORT}}. When this document uses the term "path", it refers
+This document uses the terminology defined in {{QUIC-TRANSPORT}}.
+When this document uses the term "path", it refers
 to the notion of "network path" used in {{QUIC-TRANSPORT}}.
 
-# Transport Handshake and Cryptographic Packet Protection
+Further, this document uses the variable-length integer encoding from {{QUIC-TRANSPORT}}.
+And packet diagrams in this document use the format defined in
+{{Section 1.3 of QUIC-TRANSPORT}} to illustrate the order and size of fields.
+
```
huitema: I was always warned against starting a sentence with And or But... How about:

The packet diagrams in this document uses the conventions defined  in {{Section 1.3 of QUIC-TRANSPORT}},
including the notation `(i)` to denote variable-length integers, encoded as specified in {{Section 16 of QUIC-TRANSPORT}}.

qdeconinck (approved):
LGTM

PR #619, opened 2025-11-12
Editorial rewording

flub:
If I understood this correctly this change is editorial.

"The peers that ..." reads rather strange combined with "... from the
peer ..." later in that sentence.  I think using "Endpoints" makes
this sentence easier to read.  I appreciate this is subjective though.


huitema (approved):
Yes, this reads better.

Issue #626, opened 2025-12-16
Operational Directorate (opsdir) review by Adrian Farrel


huitema:
Document: draft-ietf-quic-multipath
Title: Managing multiple paths for a QUIC connection
Reviewer: Adrian Farrel
Review result: Has Issues

Hi,

Thank you for your work on this document.

I have been selected as the Operational Directorate (opsdir) reviewer
for this Internet-Draft.

The Operational Directorate reviews all operational and management-
related Internet-Drafts to ensure alignment with operational best
practices and that adequate operational considerations are covered.

A complete set of "Guidelines for Considering Operations and Management
in IETF Specifications" can be found at
https://datatracker.ietf.org/doc/draft-opsarea-rfc5706bis/.

While these comments are primarily for the Operations and Management
Area Directors (Ops ADs), the authors should consider them alongside
other feedback received. Please feel free to contact me for
clarification and/or discussion

Apologies for the nits: I can't read a document without noting them.

I'd be happy to discuss any of these points with you.

Regards,
Adrian

======

Document: draft-ietf-quic-multipath-18

Title: Managing multiple paths for a QUIC connection

Reviewer: Adrian Farrel ([adrian@olddog.co.uk](mailto:adrian@olddog.co.uk))

Review Date: 2025-12-12

Intended Status: Standards Track

IETF last call ends: 2025-12-22

Review result: Has Issues: I have some minor concerns about this document that
I think should be resolved before publication.

== Summary ==

This document describes a way to use multiple QUIC paths to support a
single connection by using path identifiers. It deliberately puts the
assignment of addresses to these paths/connections and the steering of
traffic over the different paths out of scope.

== Operational Considerations ==

This document does not call out operational or manageability
considerations. It does discuss "path management" but this is in the
context of the QUIC protocol itself, not the management of QUIC or of
the implementations.

It would be best if the document included an Operational Considerations
section following the guidance in draft-ietf-opsawg-rfc5706bis. But
there is currently no immediate requirement to do that.

In the absence of a full section, the following points do need to be
discussed:

- How is this feature incrementally deployed? Does it matter? I think it
  is as simple as noting that implementations only use the feature if the
  remote endpoint also support it, otherwise they continue as before.

- What things need to be configured? I see:
  - Max Path ID
  - Whether to attempt multi-path system-wide or per connection
  - How to behave if the remote endpoint does not support this function
    (revert to single path, give up, log the situation)
  Anything else?

[...]

PR #632, opened 2026-01-08
Address management concerns

huitema:
This PR addresses the concerns expressed in the Operation Directorate review (see issue #626), adding a short paragraph in the Introduction with a pointer to RFC 9312, as well as a minor edit. The paragraph also addresses the issue of application mixing managed and unmanaged (not monitored) paths noted in the review.

As suggested in the review and discussion, this PR also add a not on transport parameter negotiation enabling incremental deployment.

The review suggested specifying that the initial number of path and the subsequent increments of the number are specified by the application "if not already obvious". This PR does snot add specific text here, because it should indeed be obvious that the numbers are set by the "endpoint", i.e., the combination of QUIC stack and application code. Going further than that would require discussion of how code is implemented, which would probably go beyond the scope of the document.

The PR fixes an editorial issue, replacing the statement that the nonce includes "the least significant 32 bits" of the path ID by "the 32 bits" of the path ID, which is cleared.

The other editorial suggestion are already addressed in PR #628 and #630.

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -161,7 +161,7 @@ The operational considerations for QUIC are addressed in {{RFC9312}}.
 They also apply to QUIC connections using the extensions defined in this
 document. An additional complexity is that applications might use a combination
 of monitored and non-monitored paths, but that complexity already
-exist when using path migration as defined in {{QUIC-TRANSPORT}}.
+exists when using path migration as defined in {{QUIC-TRANSPORT}}.
 
 ## Conventions and Definitions {#definition}
 


draft-ietf-quic-multipath.md
```diff
@@ -156,6 +157,12 @@ than specified in this document to avoid negative performance
 impacts, e.g., when stream data is distributed over multiple paths with
 different delays.
 
+The operation considerations for QUIC are addressed in {{RFC9312}}.
```
AdrianFarrel: s/operation/operational/

draft-ietf-quic-multipath.md
```diff
@@ -156,6 +157,12 @@ than specified in this document to avoid negative performance
...
+They apply to QUIC connections using the extensions defined in this
```
AdrianFarrel: s/They apply/They also apply/

huitema: OK.

draft-ietf-quic-multipath.md
huitema: Yes indeed. Sorry.

draft-ietf-quic-multipath.md
```diff
@@ -156,6 +157,12 @@ than specified in this document to avoid negative performance
...
+document. An additional complexity is that applications might use a combination
+of monitored and non-monitored paths, but that complexity already
+exit when using path migration as defined in {{QUIC-TRANSPORT}}.
```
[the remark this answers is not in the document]
huitema: Good catch!

Yanmei-Liu (approved):

PR #636, opened 2026-02-04
Address the review by Mike Bishop

### LINKED ISSUE #638: Discuss how we can have two path with the same tuple routed differently
From [Mike Bishop's IESG review](https://mailarchive.ietf.org/arch/msg/quic/CDIlYE3_YsexZFZSQ84IhV2aChY/):

### Section 5.8, paragraph 1
```
     [QUIC-TRANSPORT] the DPLPMTUD Maximum Packet Size (MPS) is maintained
     for each combination of local and remote IP addresses.  Note that
     with the multipath extension multiple paths could use the same
     4-tuple but might have different MPS.  One simple option, if the
```
How would two paths with "the same 4-tuple" ever have a different "combination
of local and remote IP address"? Isn't that a subset of the 4-tuple by
definition?

huitema:
This PR addresses the comments by Mike Bishop [in his IESG review](https://mailarchive.ietf.org/arch/msg/quic/CDIlYE3_YsexZFZSQ84IhV2aChY/).

Close #638
Close #640

### DIFF
--- a/draft-ietf-quic-multipath.md
+++ b/draft-ietf-quic-multipath.md
@@ -1504,7 +1504,8 @@ Further, multiple paths could be initialized simultaneously, limited by the maxi
 number of allowed paths.
 
 
-Therefore, endpoints need to limit the maximum number of paths and might consider
+Therefore, endpoints are advised to keep the maximum number of paths small
+and might consider
 additional measures to limit the number of concurrent path validation processes
 e.g., by pacing them out or limiting the number of path initiation attempts
 over a certain time period.


draft-ietf-quic-multipath.md
```diff
@@ -695,7 +695,7 @@ connection, the receiving peer SHOULD send a CONNECTION_CLOSE frame
 and enter the closing state. Alternatively, a client MAY instead try to open a new path, if
 available, and only initiate connection closure if path validation fails
 or a CONNECTION_CLOSE frame is received from the server. Similarly,
-the server MAY wait for a short, limited time such as one PTO if a
+the server MAY wait for a short, limited time such as one PTO too see if a
```
Yanmei-Liu:
```suggestion
the server MAY wait for a short, limited time such as one PTO, to see if a
```

qdeconinck:
```suggestion
the server MAY wait for a short, limited time such as one PTO to see if a
```

Yanmei-Liu (approved):
Looks good. Just one nit.

mirjak:
I agree that the current text is actually not correct because it does not increase the amplification risk. I proposed some more small edits to explicit point out the existing limits. @huitema please have another look!

PR #652, opened 2026-04-24
ack to IETF last call reviews

mirjak:


mirjak:
This addition was communicated to the RPC and will be integrated for the final RFC.

